PR #50, opened 2018-10-18
FPL-39 created form to enter other in the case with child information place holder

kurt-uk:
https://tools.hmcts.net/jira/browse/FPL-39

### Change description ###
FPL-39 created form to enter other in the case, used the place holder for the child information until the CCD functionality is added to support complex type within complex type, added tests and fixture to represent other as a person

```
[ ] Yes
[x] No
```


kurt-uk:
fireshot capture 6 - core case data i hmcts_ - http___localhost_3451_case_publicla
fireshot capture 7 - core case data i hmcts_ - http___localhost_3451_case_publicla

e2e/fixtures/others.js
```diff
@@ -0,0 +1,12 @@
...
+  name: 'John Smith',
+  DOB: {
+    day: '01',
+    month: '01',
+    year: '1985',
+  },
+  gender: 'Male',
+  birthPlace: 'Scotland',
+  address: '123 London lane, London',
+  telephoneNumber: '07888288288'
```
geek-soft: Missing semicolon

```suggestion
  telephoneNumber: '07888288288',
```

e2e/pages/enterOthers/enterOthers.js
```diff
@@ -0,0 +1,60 @@
...
+    I.selectOption(this.fields.gender, other.gender);
+    I.fillField(this.fields.birthPlace, other.birthPlace);
+    I.fillField(this.fields.address, other.address);
+  },
+
+  enterChildren(childInformation) {
+    I.fillField(this.fields.relationshipToChild, childInformation)
+  },
+
+  enterContactDetailsHidden(option) {
```
geek-soft: I would split this into two methods:

```
  revealContactDetails() {
```

and

```
  hideContactDetailsHidden(reason) {
```

e2e/pages/enterOthers/enterOthers.js
```diff
@@ -0,0 +1,60 @@
...
+    I.click(this.fields.detailsHidden(option).option);
+    if (option === 'Yes') {
+      I.fillField(this.fields.detailsHidden(option).reason, 'mock reason');
+    }
+  },
+
+  enterLitigationIssues(option) {
```
geek-soft: A `enterLitigationIssues` should take just reason as an argument e.g. `enterLitigationIssues(reason)`. Using that method would impose clicking `Yes` radio button.

e2e/pages/enterOthers/enterOthers.js
```diff
@@ -0,0 +1,60 @@
...
+  addOther: 'Add new',
+
+  enterOtherDetails(other) {
+    I.fillField(this.fields.name, other.name);
+    I.fillField(this.fields.DOB.day, other.DOB.day);
+    I.fillField(this.fields.DOB.month, other.DOB.month);
+    I.fillField(this.fields.DOB.year, other.DOB.year);
+    I.selectOption(this.fields.gender, other.gender);
+    I.fillField(this.fields.birthPlace, other.birthPlace);
+    I.fillField(this.fields.address, other.address);
```
geek-soft: No telephone number? Please add.

e2e/pages/enterOthers/enterOthers.js
kurt-uk: im not sure this would be ideal here. the call to action is to enter whether or not they want the contact details hidden and then can enter the yes or no for the radio button

e2e/pages/enterOthers/enterOthers.js
kurt-uk: good spot

e2e/pages/enterOthers/enterOthers.js
swalker125: if we're taking the view point of 'this is how the user interacts with our system' then I do think a function that takes in an option of yes or no. There should probably be an if around the reason section as I'm not sure the reason field will be in the HTML if No is selected

swalker125: detailsHidden probably isn't the best function name either. you're hiding contactDetails so, I'm currently working with enterContactDetailsAnswer. Thoughts?

kurt-uk: there is an if surrounding the reason field

swalker125:
@geek-soft @kurt-uk I think the spreadsheet could use some work naming wise. Currently the data structure that comes out should be:

```
data: {
    others_firstOther: {}
    others_additionalOthers{
        0: {}
        1: {}
    }
}
```

kurt-uk:
> @geek-soft @kurt-uk I think the spreadsheet could use some work naming wise. Currently the data structure that comes out should be:
> 
> ```
> data: {
>     others_firstOther: {}
>     others_additionalOthers{
>         0: {}
>         1: {}
>     }
> }
> ```

and thats wrong?

geek-soft:
I agree with @swalker125 that config needs some work naming wise. Happy to keep it as is as with intention to revisit it later.

swalker125:
So it'd be better as something like this

```
data: { 
    others: {
        first_other: {}
        additional: { 
            0: {}, 
            1: {}
        }
    }
}
```

PR #56, opened 2018-10-23
FPL-99: Add endpoint for a submitted case that take case data and create PDF

kurt-uk:
https://tools.hmcts.net/jira/browse/FPL-99

### Change description ###
added end point for a submitted case that would take the form data and generate a pdf returning a response of OK

```
[ ] Yes
[x] No
```


service/build.gradle
```diff
@@ -164,3 +171,10 @@ bootJar {
     attributes('Implementation-Version': project.version.toString())
   }
 }
+sourceSets {
```
geek-soft: Please remove, it works without it. that source set is configured by default in Gradle.

service/out/production/resources/application.yaml
```diff
@@ -0,0 +1,12 @@
+server:
```
geek-soft: Remove everything from service/out and add that directory to .gitignore

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+
+@RunWith(SpringRunner.class)
+@WebMvcTest
+public class CaseSubmissionControllerTest {
+
+    @Autowired
+    private MockMvc mockMvc;
+
+
```
geek-soft: Remove excessive blank lane

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+
+
+@RestController
+@RequestMapping("/callback/case-submission")
+public class CaseSubmissionController {
+
+    private final HTMLToPDFConverter converter = new HTMLToPDFConverter();
+    private final DocumentTemplates documentTemplates = new DocumentTemplates();
```
geek-soft: Please inject instance of DocumentTemplates using @Autowire

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @PostMapping
+    public ResponseEntity submittedCase(@RequestBody @NotNull Map<String,
```
geek-soft: Please add Swagger API annotations so that API documentation can be generated

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
geek-soft: Please remove ` = new DocumentTemplates()` part as this is no longer needed if DI (auto wiring) is used.

PR #57, opened 2018-10-23
fpl-95: added submit case workaround

swalker125:

https://tools.hmcts.net/jira/browse/FPL-95

### Change description ###

Added Review and Submit action that allows a Local Authority user to submit a case. Currently this is bare bones and will be fleshed out in a future story when we know what this looks like. When a case is submitted a Local Authority user no longer has edit rights on the case.

```
[ ] Yes
[x] No
```

### DIFF
--- a/ccd-definition/AuthorisationCaseState.json
+++ b/ccd-definition/AuthorisationCaseState.json
@@ -2,15 +2,15 @@
   {
     "LiveFrom": 42736,
     "CaseTypeID": "Shared_Storage_DRAFTType",
-    "CaseStateID": "1_Initiation",
+    "CaseStateID": "Open",
     "UserRole": "caseworker-publiclaw-localAuthority",
     "CRUD": "CRU"
   },
   {
     "LiveFrom": 42736,
     "CaseTypeID": "Shared_Storage_DRAFTType",
-    "CaseStateID": "2_Submitted",
+    "CaseStateID": "Submitted",
     "UserRole": "caseworker-publiclaw-localAuthority",
-    "CRUD": "R"
+    "CRUD": "CRU"
   }
 ]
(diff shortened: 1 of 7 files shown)


ccd-definition/AuthorisationCaseState.json
```diff
@@ -5,5 +5,12 @@
     "CaseStateID": "1_Initiation",
     "UserRole": "caseworker-publiclaw-localAuthority",
     "CRUD": "CRU"
+  },
+  {
+    "LiveFrom": 42736,
+    "CaseTypeID": "Shared_Storage_DRAFTType",
+    "CaseStateID": "2_Submitted",
```
geek-soft: Do we wanna prefix states with numbers? That probably have no purpose. Let's remove.

ccd-definition/State.json
```diff
@@ -6,5 +6,13 @@
     "Name": "Open",
```
geek-soft: Should that be changed to `In progress` or `Draft`? Two states should not have same name.

e2e/paths/caseSubmitted_test.js
```diff
@@ -0,0 +1,19 @@
+const config = require('../config.js');
+
+//Temporary placeholder tests.
```
geek-soft: Please remove.

e2e/paths/caseSubmitted_test.js
```diff
@@ -0,0 +1,19 @@
...
+
+Feature('Submit Case');
+
+Before((I, caseViewPage) => {
+  I.logInAndCreateCase(config.localAuthorityEmail, config.localAuthorityPassword, config.eventSummary, config.eventDescription);
+  I.selectOption(caseViewPage.actionsDropdown, config.applicationActions.submitCase);
```
geek-soft: That should be

ccd-definition/CaseEvent.json
```diff
@@ -195,8 +195,9 @@
     "Description": "Review and submit application",
     "DisplayOrder": 17,
     "PreConditionState(s)": "*",
```
geek-soft: Pre condition should be set to `1_Initiation` so only in such state submit is possible.

PR #67, opened 2018-10-25
FPL-97: added callback url to caseEvent

swalker125:

https://tools.hmcts.net/jira/browse/FPL-97

### Change description ###

Configured fpl-service to run on port 4000 in docker. Added networks to docker-compose to link service with ccd images. Refactored case fixture file to match incoming data structure. 

```
case_data: { 
    id: 
}
```
Refactored HTML template to contain this data. (case_data.id)

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/README.md
+++ b/service/README.md
@@ -39,7 +39,7 @@ This will start the API container exposing the application's port (set to `4000`
 In order to test if the application is up, you can call its health endpoint:
 
 ```bash
-  curl http://fpl-service:4000/health
+  curl http://localhost:4000/health
 ```
 
 You should get a response similar to this:
--- a/service/README.md
+++ b/service/README.md
@@ -39,7 +39,7 @@ This will start the API container exposing the application's port (set to `4000`
 In order to test if the application is up, you can call its health endpoint:
 
 ```bash
-  curl http://localhost:4000/health
+  curl http://fpl-service:4000/health
 ```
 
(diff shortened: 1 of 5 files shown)


service/src/main/resources/HTMLTemplate.html
```diff
@@ -1,9 +1,9 @@
 <html lang="en">
 <head>
-    <meta charset="UTF-8"/>
-    <title>CASE</title>
+  <meta charset="UTF-8"/>
+  <title>CASE</title>
 </head>
 <body>
-CASE ID: {{caseId}}
+Case data: {{case_details.id}}
```
geek-soft: I would love to have whitespace after `{{` and before `}}`. What about:
```suggestion
Case data: {{ case_details.id }}
```

service/README.md
```diff
@@ -39,7 +39,7 @@ This will start the API container exposing the application's port (set to `4000`
 In order to test if the application is up, you can call its health endpoint:
 
 ```bash
-  curl http://localhost:4000/health
+  curl http://fpl-service:4000/health
```
geek-soft: We have to revert it back to 

```suggestion
  curl http://localhost:4000/health
```

This call is meant to be made from host machine so `fpl-service` won't be recognised unless added to `/etc/hosts`

PR #68, opened 2018-10-26
FPL-88: Add HMCTS admin user

swalker125:

https://tools.hmcts.net/jira/browse/FPL-88

### Change description ###

Added hmcts admin user to ccd configuration. This user currently cannot complete any actions within a case and can only see submitted cases in the work basket.

```
USER DETAILS
id: hmcts-admin@example.com
password: Password12
```

```
[ ] Yes
[x] No
```


ccd-definition/AuthorisationCaseField.json
```diff
@@ -229,5 +229,236 @@
...
+  },
+  {
+    "LiveFrom": 42736,
+    "CaseTypeID": "Shared_Storage_DRAFTType",
+    "CaseFieldID": "factorsParenting",
+    "UserRole": "caseworker-publiclaw-courtadmin",
+    "CRUD": "R"
+  },
+  {
+    "LiveFrom": 42737,
```
geek-soft: Let's keep '42736' value everywhere.

geek-soft:
Events tab is empty :( I would assume that HMCTS Admin can see all events.

screenshot 2018-10-26 at 13 14 07

swalker125:
@geek-soft spoken to Moray. Version history or knowing the events leading up to submitting the case is not important for a hmcts admin. I will update the definition files to display the submitted event only for this user.

swalker125:
screenshot_2018-10-26 core case data hmcts 3
screenshot_2018-10-26 core case data hmcts 4
screenshot_2018-10-26 core case data hmcts 5

e2e/pages/login/loginPage.js
```diff
@@ -14,5 +14,6 @@ module.exports = {
     I.fillField(this.fields.username, username);
     I.fillField(this.fields.password, password);
     I.click(this.submitButton);
+    I.wait(3);
```
geek-soft: Similar wait is already added to `logInAndCreateCase`. If we add wait here we should remove it from actor.

Also we should use `waitForNavigation`, this fixed wait is very fragile.

e2e/paths/loginAsHmctsAdmin_test.js
```diff
@@ -0,0 +1,19 @@
+const config = require('../config.js');
+let caseId;
+
+Feature('Login as hmcts admin');
+
+Before(async (I, caseViewPage) => {
+  I.logInAndCreateCase(config.localAuthorityEmail, config.localAuthorityPassword, config.eventSummary, config.eventDescription);
+  caseId = await I.grabTextFrom('h2');
+  caseViewPage.goToNewActions(config.applicationActions.submitCase);
+  I.click('.button[type=submit]');
```
geek-soft: We should use `/e2e/pages/createCase/addEventSummary.js#submitCase` here with `summary` and `description` set to default `''`

swalker125: I don't think I can use this. The test will fail as there is no event description or event summary text fields on the page with the submit button

PR #106, opened 2018-11-19
FPL-208: Assign Local Authority on case initiation

swalker125:

https://tools.hmcts.net/jira/browse/FPL-208

### Change description ###

Added backend service to fetch user based on authorisation and add a local authority to CCD case data based on the domain of that user's email address. Backend Service needs to be up and running using docker-compose up for callback to succeed locally.

```
[ ] Yes
[x] No
```


service/build.gradle
```diff
@@ -177,6 +178,7 @@ dependencies {
     exclude group: 'junit', module: 'junit'
   }
   testCompile group: 'org.junit.jupiter', name :'junit-jupiter-api', version: '5.3.1'
+  testCompile group: 'org.junit.jupiter', name :'junit-jupiter-params', version: '5.2.0'
```
geek-soft: Use same version consistently

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseRepositoryTest.java
```diff
@@ -62,6 +64,10 @@ void shouldSetSubmittedFormPDF() throws IOException {
         verify(coreCaseDataApi).submitEventForCaseWorker(eq(AUTH_TOKEN), eq(SERVICE_AUTH_TOKEN), eq(USER_ID), eq(JURISDICTION), eq(CASE_TYPE), eq(caseId), eq(true), caseDataContentArgumentCaptor.capture());
 
         CaseDataContent caseDataContent = caseDataContentArgumentCaptor.getValue();
+
+        System.out.println("caseDataContent = " + caseDataContent.getData());
+
```
geek-soft: No logs using println, please remove

swalker125: Good spot

service/src/integrationTest/resources/fixtures/initiate-case.json
```diff
@@ -0,0 +1,8 @@
+{
```
geek-soft: I believe we could use `core-case-data-store-api/callback-request.json` and remove `initiate-case.json`. Having one fixture should make maintenance easier and we can always introduce new one when really needed.

service/src/main/java/uk/gov/hmcts/reform/fpl/events/InitiateCaseEvent.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.hmcts.reform.fpl.events;
+
+import uk.gov.hmcts.reform.ccd.client.model.CallbackRequest;
+
+public class InitiateCaseEvent {
```
geek-soft: Because every single post submit event will be the same I would suggest creating base class and then having two subclasses, one for `InitiateCaseEvent` and other one for `SubmittedCaseEvent`.

```
CallbackEvent:
  - InitiateCaseEvent
  - SubmittedCaseEvent
```

We should also consider using the same wording format for all the events, at the moment one uses past tense other does not.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/InitiateCaseEventHandler.java
```diff
@@ -0,0 +1,45 @@
...
+     * @param event initiate case event.
+     */
+    @Async
+    @EventListener
+    public void handleCaseInitiation(InitiateCaseEvent event) {
+        String userId = event.getUserId();
+        String authorization = event.getAuthorization();
+        CaseDetails caseDetails = event.getCallbackRequest().getCaseDetails();
+
+        String caseLocalAuthority = userService.getUserDetails(authorization);
```
geek-soft: Behaviour is not as per Tech Design. We should use Lookup Table and store LA CODE in data as opposite to storing LA name.

See: https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=817169070#NotifyonSubmission-LocalAuthorities

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * Removes all characters from an email address apart from the domain.
+     *
+     * @param email the email address retrieved from Idam.
+     * @return caseLocalAuthority the value to be stored in case data.
+     */
+    private String getEmailDomain(String email) {
+        int removeUser = email.indexOf("@");
```
geek-soft: Please rename to inline. Name looks pretty much like method name.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,48 @@
...
+        String tempState = email.substring(removeUser + 1);
+
+        int removeEnd = tempState.indexOf(".");
```
geek-soft: Same here, variable name looks like method name.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,48 @@
...
+        String caseLocalAuthority = tempState.substring(0, removeEnd);
+
+        return caseLocalAuthority;
```
geek-soft: We should probably lower case domain before we start processing it any further just in case it contains anything other than lower case letters.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/UserServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String AUTH_TOKEN = "Bearer token";
+
+    @Mock
+    protected IdamApi idamApi;
+
+    @InjectMocks
+    private UserService userService;
+
+    @ParameterizedTest
+    @ValueSource(strings = {"mock@example.gov.uk", "mock@example.com", "mock.mock@example.gov.uk"})
```
geek-soft: Nice, taking advantage of jUnit5 already 👍

service/src/test/java/uk/gov/hmcts/reform/fpl/service/UserServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.BDDMockito.given;
+
+@ExtendWith(SpringExtension.class)
+class UserServiceTest {
+
+    private static final String AUTH_TOKEN = "Bearer token";
+
+    @Mock
+    protected IdamApi idamApi;
```
geek-soft: Let's make it private, no reason to have it protected.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/UserServiceTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+        Assertions.assertThat(domain).isEqualTo(expectedDomain);
+    }
+
+    @Test
+    void shouldReturnExceptionWhenIdamApiThrows() {
+        given(idamApi.retrieveUserDetails(AUTH_TOKEN)).willThrow(
+            new RuntimeException("user does not exist"));
+
+
```
geek-soft: Please remove excessive blank line.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,48 @@
...
+     * @return caseLocalAuthority gained from the getEmailDomain function.
+     */
+    public String getUserDetails(String authorization) {
+        UserDetails userDetails = idamApi.retrieveUserDetails(authorization);
+        String email = userDetails.getEmail();
+        return getEmailDomain(email);
+    }
+
+    /**
+     * Removes all characters from an email address apart from the domain.
```
geek-soft: I am not a fun of JavaDoc on private methods. IMO private methods should be so small that method name should describe exactly what it does. For me it is a bit unnecessary effort that does not give much value.

I am OK with JavaDoc on public methods as it sometimes helps using the API. Are we as. a team happy writing JavaDoc for every single public method?

ccd-definition/CaseEvent.json
```diff
@@ -7,7 +7,21 @@
     "Description": "Create a new case – add a title",
     "DisplayOrder": 1,
     "PostConditionState": "Open",
-    "SecurityClassification": "Public"
+    "SecurityClassification": "Public",
+    "CallBackURLSubmittedEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/case-initiation"
```
geek-soft: TBH I though more of `AboutToStart` callback which is to be used to prepare data.

Only question is whether failure in `AboutToStart` would stop case from being able to be created.

swalker125: Submitted events are intended to show a case has transitioned into a new state. i.e non existant to created... I think it's the right type of callback for this sitution.

About to start callback is more for data validation from what I can see. Yes it would stop the event from happening if local authority was not set correctly.

geek-soft: A `AboutToSubmit` event is dedicated for validation.

The way I look at this is if we had no CCD and some sort of fire and forget processing, regardless how robust that is, extracting LA name from user token would happen within main transaction, not as post processing. Similarly here I find `AboutToStart` more appropriate for what we doing.

Happy to discuss when I get back from the meeting.

service/src/integrationTest/resources/fixtures/initiate-case.json
geek-soft: Wired, what is the error?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
swalker125: A good test case. I've added to lower case as follows:

```
 String tempState = email.toLowerCase().substring(start + 1);
```

Is this ok?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
swalker125: Yep, happy to write docs for the public methods

ccd-definition/CaseEvent.json
swalker125: AboutToSubmit could be more appropriate. Reading docs we can put it into a different state if that local authority isn't set. We could maybe configure a manual select local authority field if this happens?

ccd-definition/CaseEventToFields.json
```diff
@@ -3,8 +3,20 @@
...
     "CaseTypeID": "Shared_Storage_DRAFTType",
     "CaseEventID": "openCase",
-    "CaseFieldID": "caseName",
+    "CaseFieldID": "caseLocalAuthority",
     "PageFieldDisplayOrder": 1,
+    "DisplayContext": "READONLY",
+    "PageID": 1,
+    "PageDisplayOrder": 1,
+    "PageColumnNumber": 1,
+    "FieldShowCondition": "caseName=\"null\""
```
geek-soft: :( @swalker125 , have you got any advise from CCD on this?

docker-compose.yml
```diff
@@ -10,10 +10,12 @@ services:
     image: docker.artifactory.reform.hmcts.net/fpl/fpl-service
     container_name: fpl-service
     environment:
+      - IDAM_API_URL=http://idam-api:8080
       - DOCUMENT_MANAGEMENT_URL=http://dm-store:8080
       - CORE_CASE_DATA_API_URL=http://ccd-data-store-api:4452
       - IDAM_S2S_AUTH_URL=http://service-auth-provider-api:8080
       - IDAM_S2S_AUTH_TOTP_SECRET=AABBCCDDEEFFGGHH
+      - FPL_LOCAL_AUTHORITY_MAPPING=example.com=>example
```
geek-soft: Let's rename this to be `FPL_LOCAL_AUTHORITY_NAME_MAPPING` to make it less generic.

kubernetes/deployment.template.yaml
```diff
@@ -37,6 +37,7 @@ data:
   CORE_CASE_DATA_API_URL: "http://ccd-data-store-api"
   IDAM_S2S_AUTH_URL: "http://service-auth-provider-api"
   IDAM_S2S_AUTH_TOTP_SECRET: "AABBCCDDEEFFGGHH"
+  FPL_LOCAL_AUTHORITY_MAPPING: "example.com=>example;swansea.gov.uk=>SA;hampshire.gov.uk=>HE"
```
geek-soft: Let's remove `swansea.gov.uk` and `hampshire.gov.uk` for now.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+        MvcResult response = mockMvc
+            .perform(post("/callback/case-initiation")
+                .header("authorization", AUTH_TOKEN)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(readBytes("core-case-data-store-api/empty-case-details.json")))
+            .andExpect(status().isOk())
+            .andReturn();
+
+        assertThat(response.getResponse().getContentAsString()).contains(caseLocalAuthority);
```
geek-soft: Would be good to assert proper structure so that we assure that the right field is in data block.

service/src/main/java/uk/gov/hmcts/reform/fpl/config/LocalAuthorityLookupConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.context.annotation.Configuration;
+
+import java.util.Arrays;
+import java.util.Map;
+
+@Configuration
+public class LocalAuthorityLookupConfiguration {
+
+    @Value("${fpl.local_authority.mapping}")
+    private String mapping;
```
geek-soft: What about?

```
public LocalAuthorityLookupConfiguration(@Value("${fpl.local_authority.mapping}") String config) {
         ImmutableMap.Builder<String, String> localAuthorities = ImmutableMap.builder();

         Arrays.stream(config.split(";")).forEach(mapping -> {
             String[] localAuthorityData = mapping.split("=>");
             localAuthorities.put(localAuthorityData[0], localAuthorityData[1]);
         });

         this.mapping = localAuthorities.build();
}

public Map<String, String> getLookupTable() {
         return mapper;
}
```

ccd-definition/CaseEventToFields.json
swalker125: @geek-soft I did not. I don't think it is possible in any other way. I did get a reply saying please let me know if you do discover a way though 😂

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
swalker125: @geek-soft I've created a JSONObject and I now assert that the response.toString contains the JSONObject.toString

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationController.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.Map;
+
+@Api
+@RestController
+@RequestMapping("/callback/case-initiation")
+public class CaseInitiationController {
+
+    @Autowired
+    private UserService userService;
```
geek-soft: Let's use constructor injection for consistency and to make testing easier.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Map;
+
+/**
+ * Gets Users from Idam.
+ */
+@Service
+public class UserService {
+
+    private final IdamApi idamApi;
+    private final LocalAuthorityLookupConfiguration localAuthorityLookupConfiguration;
```
geek-soft: Shall we rename it to be `LocalAuthorityNameLookupConfiguration`? Again current name is a bit generic.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,51 @@
...
+    public UserService(IdamApi idamApi, LocalAuthorityLookupConfiguration localAuthorityLookupConfiguration) {
+        this.idamApi = idamApi;
+        this.localAuthorityLookupConfiguration = localAuthorityLookupConfiguration;
+    }
+
+    /**
+     * Makes a request to Idam and returns the user details for the user who is making the request.
+     *
+     * @param authorization the authorization token generated by the callback.
+     * @return caseLocalAuthority gained from the extractEmailDomain function.
```
geek-soft: I do not think we should say what internal method is used. It will change and from user who makes call should not care. What is important is what return value is.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,51 @@
...
+    @Autowired
+    public UserService(IdamApi idamApi, LocalAuthorityLookupConfiguration localAuthorityLookupConfiguration) {
+        this.idamApi = idamApi;
+        this.localAuthorityLookupConfiguration = localAuthorityLookupConfiguration;
+    }
+
+    /**
+     * Makes a request to Idam and returns the user details for the user who is making the request.
+     *
+     * @param authorization the authorization token generated by the callback.
```
geek-soft: Saying just IDAM authorisation token would be sufficient, we do not need to mention where that value comes from as it might come from various places in the future.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,51 @@
...
+    private final LocalAuthorityLookupConfiguration localAuthorityLookupConfiguration;
+
+    @Autowired
+    public UserService(IdamApi idamApi, LocalAuthorityLookupConfiguration localAuthorityLookupConfiguration) {
+        this.idamApi = idamApi;
+        this.localAuthorityLookupConfiguration = localAuthorityLookupConfiguration;
+    }
+
+    /**
+     * Makes a request to Idam and returns the user details for the user who is making the request.
```
geek-soft: Let's also update comment specifying what is being returned, it is not 'user details' as doc claims.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/UserServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+    private IdamApi idamApi;
+
+    @Mock
+    private LocalAuthorityLookupConfiguration localAuthorityLookupConfiguration;
+
+    @InjectMocks
+    private UserService userService;
+
+    @ParameterizedTest
+    @SuppressWarnings({"LineLength"})
```
geek-soft: Do we need that? Lines seems reasonable length.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/UserServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @ValueSource(strings = {"mock@example.gov.uk", "mock.mock@example.gov.uk", "mock@ExAmPlE.gov.uk"})
+    void shouldReturnDomainForSuccessfulIdamCall(String email) {
+        final String expectedLaCode = "EX";
+        Map<String, String> localAuthorities = new HashMap<>();
```
geek-soft: Please replace with Immutable.Map builder. That fluent API is much better.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserService.java
```diff
@@ -0,0 +1,51 @@
...
+        int start = email.indexOf("@");
+        String tempState = email.toLowerCase().substring(start + 1);
+
+        return lookUpCode(tempState);
+    }
+
+    private String lookUpCode(String emailDomain) {
+        Map<String, String> lookupTable = localAuthorityLookupConfiguration.getLookupTable();
+
+        return lookupTable.get(emailDomain);
```
geek-soft: Let's throw custom exception in case value does not exist in lookup table.

geek-soft: + let's have right test.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final String AUTH_TOKEN = "Bearer token";
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @MockBean
+    private IdamApi idamApi;
+
+    @Test
+    void shouldAddCaseLocalAuthorityToCaseData() throws Exception {
```
geek-soft: Please add additional test case to assert that when user is not recognised endpoint returns no data block but error array with one error instead.

service/src/main/java/uk/gov/hmcts/reform/fpl/config/LocalAuthorityNameLookupConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Arrays;
+import java.util.Map;
+
+@Configuration
+public class LocalAuthorityNameLookupConfiguration {
+
+    private static final String MAPPINGS_SEPARATOR = ";";
+    private static final String ENTRY_SEPARATOR = "=>";
+
+    private Map<String, String> mapping;
```
geek-soft: Let's make it final. I like to have class level fields final making sure no-one will ever override reference.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityNameService.java
```diff
@@ -0,0 +1,56 @@
...
+     * @return caseLocalAuthority for user.
+     */
+    public String getLocalAuthorityCode(String authorization) {
+        UserDetails userDetails = idamApi.retrieveUserDetails(authorization);
+        String email = userDetails.getEmail();
+
+        return extractEmailDomain(email);
+    }
+
+    private String extractEmailDomain(String email) {
```
geek-soft: That method should just extract domain out of an email address but it triggers lookup as well. Please move `lookUpCode(tempState)` to `getLocalAuthorityCode` method.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityNameService.java
```diff
@@ -0,0 +1,56 @@
...
+        String tempState = email.toLowerCase().substring(start + 1);
+
+        return lookUpCode(tempState);
+    }
+
+    private String lookUpCode(String emailDomain) {
+        Map<String, String> lookupTable = localAuthorityLookupConfiguration.getLookupTable();
+
+        if (lookupTable.get(emailDomain) == null) {
+            throw new IllegalArgumentException(emailDomain + " not found");
```
geek-soft: Let's throw custom exception e.g. `UnknownLocalAuthorityDomain` which then we can nicely handle in exception handler.

See: https://github.com/hmcts/cmc-claim-store/blob/master/src/main/java/uk/gov/hmcts/cmc/claimstore/controllers/advices/ResourceExceptionHandler.java#L103

geek-soft:
Mutation test coverage looks good as well @swalker125 👍🏻

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
swalker125: Will this ever happen? A user needs to be on Idam to login, and their authorisation token will always link them to that entry in idam

sivas23:
Tested this story in Preview environment and local environment and passed QA.

PR #115, opened 2018-11-26
FPL-146: Grants access to local authority users

swalker125:

https://tools.hmcts.net/jira/browse/FPL-218

### Change description ###

Added mapping for local authority to user ids. 

Grants access to users based on the local authority stored in case data

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
```diff
@@ -86,4 +100,31 @@ void shouldPopulateErrorsInResponseWhenDomainNameIsNotFound() throws Exception {
...
+            .perform(post("/callback/case-initiation/submitted")
+                .header("authorization", AUTH_TOKEN)
+                .header("user-id", USER_ID)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(MAPPER.writeValueAsString(request)))
+            .andExpect(status().isOk());
+
+        Thread.sleep(3000);
+
+        verify(caseAccessApi, times(3)).grantAccessToCase(
```
geek-soft: Please make sure correct values are passed to `grantAccessToCase` method. We should be as strict as we only can.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationController.java
```diff
@@ -43,4 +50,15 @@ public ResponseEntity createdCase(
...
+
+    @PostMapping("/submitted")
+    public ResponseEntity handleCaseInitiation(
+        @RequestHeader(value = "authorization") String authorization,
+        @RequestHeader(value = "user-id") String userId,
+        @RequestBody CallbackRequest callbackRequest) {
+
+        applicationEventPublisher.publishEvent(new InitiatedCaseEvent(callbackRequest, authorization, userId));
+
+        return new ResponseEntity(HttpStatus.OK);
```
geek-soft: Please remove and change handleCaseInitiation to return void. It will work exactly same way.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationController.java
```diff
@@ -43,4 +50,15 @@ public ResponseEntity createdCase(
 
         return ResponseEntity.ok(body);
     }
+
+    @PostMapping("/submitted")
+    public ResponseEntity handleCaseInitiation(
```
geek-soft: Please rename method to be something like `handleSubmittedEvent`.

Please also rename `createdCase` to something like `handleAboutToSubmitEvent`

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/InitiatedCaseEventHandler.java
```diff
@@ -0,0 +1,31 @@
...
+    private final LocalAuthorityUserService localAuthorityUserService;
+
+    @Autowired
+    public InitiatedCaseEventHandler(LocalAuthorityUserService localAuthorityUserService) {
+        this.localAuthorityUserService = localAuthorityUserService;
+    }
+
+    @Async
+    @EventListener
+    public void handleCaseInitiation(InitiatedCaseEvent event) {
```
geek-soft: I think Handlers should communicate what is happening through method signatures. Please rename to `grantCaseAccessToAllLocalAuthorityUsers`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityUserService.java
```diff
@@ -0,0 +1,50 @@
...
+            authorization,
+            authTokenGenerator.generate(),
+            userId,
+            JURISDICTION,
+            CASE_TYPE,
+            caseId,
+            new UserId(id)));
+    }
+
+    private List<String> lookUpCode(String localAuthorityCode) {
```
geek-soft: Please rename to highlight more real behaviour. We should communicate what is being returned as by looking at current method signature `List<String>` could be anything.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/InitiatedCaseEventHandlerTest.java
```diff
@@ -0,0 +1,4 @@
+package uk.gov.hmcts.reform.fpl.handlers;
+
+public class InitiatedCaseEventHandlerTest {
```
geek-soft: I guess we could remove that test as there is very small responsibility in handler itself.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityUserServiceTest.java
```diff
@@ -0,0 +1,61 @@
...
+    private CaseAccessApi caseAccessApi;
+    @Mock
+    private LocalAuthorityUserLookupConfiguration localAuthorityUserLookupConfiguration;
+
+    @InjectMocks
+    private LocalAuthorityUserService localAuthorityUserService;
+
+    @Test
+    @SuppressWarnings({"LineLength"})
+    void shouldCallGrantAccessToCaseWithExpectedParams() {
```
geek-soft: We should also unit test:
- make sure failure in granting access to first user does not stop other calls
- make sure exception is thrown when LA is unknown
- make sure exception is thrown when users list assigned to LA is blank (potentially)
- we could also make sure user crating case does not trigger call for itself as there is no need for that

geek-soft:
Mapping needs to be configured for AAT as well.

service/src/test/java/uk/gov/hmcts/reform/fpl/Constants.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.hmcts.reform.fpl;
+
+@SuppressWarnings("LineLength")
+public interface Constants {
+
+    /**
+     * Service auth token issued for 'fpl_case_service' valid until end of the day on 31st December 2049.
```
kurt-uk: are you going to be monitoring this?
#DontForget

service/src/test/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityUserServiceTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Mock
+    private CaseAccessApi caseAccessApi;
+    @Mock
+    private LocalAuthorityUserLookupConfiguration localAuthorityUserLookupConfiguration;
+
+    @InjectMocks
+    private LocalAuthorityUserService localAuthorityUserService;
+
+    @Test
+    @SuppressWarnings({"LineLength"})
```
kurt-uk: can you sort the line length?

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/InitiatedCaseEventHandlerTest.java
kurt-uk: this seems a small file

service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityUserService.java
```diff
@@ -0,0 +1,50 @@
...
+                                     AuthTokenGenerator authTokenGenerator) {
+        this.caseAccessApi = caseAccessApi;
+        this.localAuthorityUserLookupConfiguration = localAuthorityUserLookupConfiguration;
+        this.authTokenGenerator = authTokenGenerator;
+    }
+
+    public void grantUserAccess(String authorization, String userId, String caseId, String caseLocalAuthority) {
+        List<String> userIds = lookUpCode(caseLocalAuthority);
+
+        userIds.forEach((id) -> caseAccessApi.grantAccessToCase(
```
geek-soft: Please also log any failure in granting access using at least WARNING severity. Logged entry should give us enough details to grant access manually if needed.

geek-soft: Finally when everything is sorter please look into ways of having simple retry with spring so when API call failed due to some intermittent issue we can try again.

e2e/paths/accessSegregation_test.js
```diff
@@ -0,0 +1,31 @@
...
+  caseId = caseId.toString().replace('#', '');
+  caseViewPage.goToNewActions(config.applicationActions.submitCase);
+  I.click('Submit');
+  I.click('Sign Out');
+});
+
+Scenario('Different user in the same local authority can see case created', (I, loginPage, caseListPage) => {
+  loginPage.signIn(config.swanseaLocalAuthorityEmailDamian, config.localAuthorityPassword);
+  caseListPage.changeStateFilter('Submitted');
+  I.see(caseId);
```
geek-soft: This test is fragile and will break when there is more than 25 submitted cases per environment.

e2e/config.js
```diff
@@ -2,7 +2,11 @@
 
 module.exports = {
   // users
-  localAuthorityEmail: 'kurt@swansea.gov.uk',
+  swanseaLocalAuthorityEmailKurt: 'kurt@swansea.gov.uk',
+  swanseaLocalAuthorityEmailDamian: 'damian@swansea.gov.uk',
+  swanseaLocalAuthorityEmailJames: 'james@swansea.gov.uk',
```
geek-soft: Never used, please remove.

e2e/paths/accessSegregation_test.js
swalker125: @geek-soft we happy to skip them?

e2e/config.js
```diff
@@ -2,7 +2,11 @@
...
+  hillingdonLocalAuthorityEmailSam: 'sam@hillingdon.gov.uk',
+  hillingdonLocalAuthorityEmailSiva: 'siva@hillingdon.gov.uk',
```
geek-soft: Never used, please remove.

e2e/config.js
```diff
@@ -2,7 +2,11 @@
 
 module.exports = {
   // users
-  localAuthorityEmail: 'kurt@swansea.gov.uk',
+  swanseaLocalAuthorityEmailKurt: 'kurt@swansea.gov.uk',
```
geek-soft: I would prefer to make these names more generic so that we can always change user names without affecting E2E too much. Just have no good suggestion for the var name. Will came up with something tomorrow.

swalker125: changed whilst removing. gone for swanseaLocalAuthorityUserOne etc

e2e/paths/accessSegregation_test.js
geek-soft: Nope, we should make them work. I remember the idea we had during 3 amigos session was to just jump to the case details url directly.

e2e/paths/accessSegregation_test.js
```diff
@@ -0,0 +1,33 @@
...
+  caseId = await I.grabTextFrom('.heading-medium');
+  caseUrl = caseId.toString().replace('#', '').replace(/-/g, '');
+  caseViewPage.goToNewActions(config.applicationActions.submitCase);
+  I.click('Submit');
+  I.click('Sign Out');
+});
+
+Scenario('Different user in the same local authority can see case created', async (I, loginPage) => {
+  loginPage.signIn(config.swanseaLocalAuthorityEmailUserTwo, config.localAuthorityPassword);
+  I.navigateToUrl('http://localhost:3451/case/PUBLICLAW/Shared_Storage_DRAFTType/' + caseUrl);
```
geek-soft: Please use `URL` env otherwise it will never work on non localhost env.

e2e/pages/caseList/caseList.js
```diff
@@ -9,4 +9,9 @@ module.exports = {
     const caseLink = locate('a').withAttr({href: href});
     I.click(caseLink);
   },
+
+  changeStateFilter(desiredState) {
```
geek-soft: This is not really needed now. Shall we delete it or keep it in case we need it in the future?

e2e/paths/accessSegregation_test.js
```diff
@@ -0,0 +1,36 @@
...
+  I.see(caseId);
+});
+
+Scenario('Different user in a different local authority cannot see case created', (I, loginPage) => {
+  loginPage.signIn(config.hillingdonLocalAuthorityEmailUserOne, config.localAuthorityPassword);
+  I.navigateToUrl(baseUrl + '/case/PUBLICLAW/Shared_Storage_DRAFTType/' + caseUrl);
+  I.seeInCurrentUrl('error');
+});
+
+Scenario('HMCTS admin user can see the case', (I, loginPage) => {
```
geek-soft: Shall we add similar scenario to make sure CAFCASS user can see the case as well?

e2e/pages/caseList/caseList.js
swalker125: I was thinking it would be used in the future when play some stories around filtering

geek-soft: Alright, happy to keep it.

e2e/paths/accessSegregation_test.js
```diff
@@ -0,0 +1,36 @@
...
+
+Feature('Cases visible only to respective local authority and admin');
+
+Before(async (I, caseViewPage) => {
+  I.logInAndCreateCase(config.swanseaLocalAuthorityEmailUserOne, config.localAuthorityPassword);
+  caseId = await I.grabTextFrom('.heading-medium');
+  caseUrl = caseId.toString().replace('#', '').replace(/-/g, '');
+  caseViewPage.goToNewActions(config.applicationActions.submitCase);
+  I.click('Submit');
+  I.click('Sign Out');
```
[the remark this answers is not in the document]
geek-soft: We will create small PR to do it after this is merged.

SamanthaSJ:
Verified PR-115 user roles. Passed.

PR #121, opened 2018-11-29
addedfunctionaltestinnightlypipeline

sivas23:

### Change description ###
Added functional tests in nightly pipeline

```
[ ] Yes
[ x] No
```

### DIFF
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -13,7 +13,7 @@ def component = "fpl-service"
 
 withNightlyPipeline(type, product, component) {
 
-  env.TEST_URL = 'https://ccd-case-management-web-aat.service.core-compute-aat.internal'
+  //env.TEST_URL = 'https://ccd-case-management-web-aat.service.core-compute-aat.internal'
   env.URL = 'http://ccd-case-management-web-aat.service.core-compute-aat.internal';
   env.PROXY_SERVER = "proxyout.reform.hmcts.net:8080";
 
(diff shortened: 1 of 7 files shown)


Jenkinsfile_nightly
```diff
@@ -6,17 +6,25 @@ properties([
 
 @Library("Infrastructure")
 
-def type = "java"
+def type = "nodejs"
```
geek-soft: Revert to Java and run functional tests from Gradle.

sivas23: Reverted to Java and added command to run functional tests from Gradle

Jenkinsfile_nightly
```diff
@@ -12,11 +12,20 @@ def product = "fpl"
...
+//  enableMutationTest()
+//
+//  after('mutationTest') {
+//    withSonarQubeEnv("SonarQube") {
+//      sh "./gradlew --info sonarqube"
+//    }
+//  }
+
+  //to Run functional test
+  sh "./gradlew runFunctionalTest"
```
geek-soft: I do not think it would even run TBH. It is definite not what the pipeline should look like.

geek-soft: Also calling `runFunctionalTest` itself would not work as there would not be yarn dependencies installed. A `functional` task is an entry point to functional tests.

sivas23: Hi Damian, changed to functional task as an entry point.

geek-soft:
Closing to start from scratch.

PR #132, opened 2018-12-05
Use HTML reporter for E2E (smoke and functional) tests

sivas23:
There is no ticket raised for this.

### Change description ###
Displaying Mocha reports  in the Jenkins

```
[ ] Yes
[ x] No
```


codecept.conf.js
```diff
@@ -77,6 +77,14 @@ exports.config = {
           mochaFile: 'test-results/result.xml',
         },
       },
+      'mochawesome': {
+        stdout: './output/console.log',
```
geek-soft: This is absolutely useless because it reports output from one chunk only. Replace with `stdout: '-',`.

sivas23: Replaced 'stdout: './output/console.log' with stdout: '-', as per the comments Damian.

geek-soft:
@sivas23 can you be more careful in terms of not committing files that should not be in repo?

Please remove `yarn-error.log`. Nobody needs that.

PR #149, opened 2018-12-14
Fix E2E tests issue caused by stateful page objects

geek-soft:

None

### Change description ###

This PR changes the way state in page objects is handled. From now on state will is kept inside of the page object and when page object is injected into test scenario a deep clone is performed to get initial state.

```
[ ] Yes
[x] No
```

### DIFF
--- a/codecept.conf.js
+++ b/codecept.conf.js
@@ -29,7 +29,7 @@ exports.config = {
       show: process.env.SHOW_BROWSER_WINDOW || false,
       restart: true,
       waitForNavigation: 'networkidle0',
-      waitForTimeout: 30000,
+      waitForTimeout: 15000,
       chrome: {
         ignoreHTTPSErrors: true,
         args: process.env.PROXY_SERVER ? [
@@ -90,7 +90,7 @@ exports.config = {
     },
   },
   tests: './e2e/paths/*_test.js',
-  timeout: 30000,
+  timeout: 15000,
   mocha: {
     reporterOptions: {
       'codeceptjs-cli-reporter': {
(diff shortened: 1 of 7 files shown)


codecept.conf.js
```diff
@@ -12,7 +29,7 @@ exports.config = {
       show: process.env.SHOW_BROWSER_WINDOW || false,
       restart: true,
       waitForNavigation: 'networkidle0',
-      waitForTimeout: 30000,
+      waitForTimeout: 15000,
```
geek-soft: This is to reduce wait time in E2E tests that are about to fail. 15 seconds should be good as default for local and AKS. Could be extended in the future for slower envs.

e2e/fragments/addressPostcodeLookup.js
```diff
@@ -19,10 +19,11 @@ module.exports = {
   },
   findAddressButton: 'Find address',
   cantEnterPostcodeLink: locate('span').withText('I can\'t enter a UK postcode'),
-  
+
   lookupPostcode(address) {
     I.fillField(this.fields.postcodeLookup, address.postcode);
     I.click(this.findAddressButton);
+    I.waitForElement(locate(this.fields.addressList).find('option').withText(address.lookupOption));
```
geek-soft: This gives up to 15sec wait to give AJAX query time to come back with addresses.

e2e/pages/enterChildren/enterChildren.js
```diff
@@ -1,10 +1,14 @@
 const I = actor();
 const postcodeLookup = require('../../fragments/addressPostcodeLookup');
-let activeChild = 'firstChild';
 
 module.exports = {
-
-  fields: (childNo) => {
+  state: {
+    context: 'firstChild',
```
geek-soft: State is supposed to be handled within page objects now to allow clone.

codecept.conf.js
```diff
@@ -49,21 +66,31 @@ exports.config = {
...
     autoDelay: {
       enabled: true,
+      methods: [
+        'click',
+        'doubleClick',
+        'rightClick',
+        'fillField',
+        'pressKey',
+        'checkOption',
+        'selectOption',
```
geek-soft: This adds `selectOption` operation to list of auto delayed operations. It will add 100ms delay before the call and 200ms delay after.

PR #152, opened 2018-12-14
FPL-364: Change jurisdiction and case type names

mh-hmcts:

### Change description ###

```
[ ] Yes
[ x] No
```

### DIFF
--- a/ccd-definition/CaseType.json
+++ b/ccd-definition/CaseType.json
@@ -2,8 +2,8 @@
   {
     "LiveFrom": "01/01/2017",
     "ID": "Shared_Storage_DRAFTType",
-    "Name": "Shared_Storage_DRAFT_v0.3",
-    "Description": "Shared_Storage_DRAFt_v01",
+    "Name": "Care, supervision and EPOs",
+    "Description": "Care, supervision and emergency protection orders",
     "JurisdictionID": "PUBLICLAW",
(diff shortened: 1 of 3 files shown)


ccd-definition/CaseType.json
```diff
@@ -2,8 +2,8 @@
   {
     "LiveFrom": "01/01/2017",
     "ID": "Shared_Storage_DRAFTType",
```
geek-soft: Please change case ID

geek-soft: If not deployment to production will override current shared storage POC.

mh-hmcts: agreed planned and raised raised as separate task

ccd-definition/Jurisdiction.json
```diff
@@ -2,7 +2,7 @@
   {
     "LiveFrom": "01/01/2017",
     "ID": "PUBLICLAW",
```
geek-soft: I guess that ID can remain unchanged. New case type will just appear under jurisdiction that is in use. I guess having rename only is sensible.

ccd-definition/CaseType.json
geek-soft: Not sure what is best name for the ID, I guess we could `CARE_SUPERVISION_EPO`.

PR #182, opened 2019-01-09
FPL-508: Move PDF generation to about-to-submit callback

geek-soft:

https://tools.hmcts.net/jira/browse/FPL-508

### Change description ###

Change moves PDF generation to about-to-submit callback. This way PDF linking can be done within same event that brings two main benefits:
- event previously used by system for linking is no longer needed and won't appear in dropdown
- document link is ready before transaction commits which means that no reload is needed to show link in tabs

There is one side effect though. If for some reason transaction commit failed an orphan document will contaminate DM storage. 

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -50,6 +63,31 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
...
+        );
+
+        Document document = uploadDocumentService.uploadPDF(userId, authorization, pdf, buildFileName(caseDetails));
+
+        Map<String, Object> data = caseDetails.getData();
+        data.put("submittedForm", ImmutableMap.<String, String>builder()
+            .put("document_url", document.links.self.href)
+            .put("document_binary_url", document.links.binary.href)
+            .put("document_filename", document.originalDocumentName)
+            .build());
```
swalker125: This is much better than needing the extra event.

SamanthaSJ:
Verified in PR-182 on 14/01/2019 Pass
We no longer need to hit the refresh button or exit the case and go back into it to view the PDF. PDF appears instantly in the Documents tab after the case has been submitted.

PR #189, opened 2019-01-17
FPL-227: Grounds for EPO reasoning

swalker125:

https://tools.hmcts.net/jira/browse/FPL-227

### Change description ###

Added AboutToStart callback to enter grounds event.
Callback populates EPO_REASONING_SHOW with SHOW_FIELD value.

Added new complex type groundsForEPO, which is only displayed when EPO_REASONING_SHOW contains the SHOW_FIELD value.

Updated pdf top contain new groundsForEPO complex type.

screenshot_2019-01-17 core case data hmcts
screenshot_2019-01-17 core case data hmcts 1
screen shot 2019-01-17 at 12 07 03

```
[ ] Yes
[x] No
```


ccd-definition/CaseEventToFields.json
```diff
@@ -37,11 +37,35 @@
...
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseEventID": "enterGrounds",
-    "CaseFieldID": "grounds",
+    "CaseFieldID": "EPO_REASONING_SHOW",
     "PageFieldDisplayOrder": 1,
     "DisplayContext": "OPTIONAL",
     "PageID": 1,
     "PageDisplayOrder": 1,
+    "PageColumnNumber": 1,
+    "FieldShowCondition": "groundsForEPO CONTAINS \"SHOW_FIELD\""
```
geek-soft: I wonder what we could do to explain that this is nasty workaround. If somebody sees that piece of code 3 months later and tries to understand that code, it might end with conclusion it does not make sense and even remove it.

Maybe we should have something like `"groundsForEPO CONTAINS \"This is workaround that hides field from the UI\""` or any other explanation that makes sense.

swalker125: It can definitely be changed to something more descriptive.

ccd-definition/CaseField.json
```diff
@@ -25,11 +25,28 @@
     "FieldType": "Text",
     "SecurityClassification": "Public"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "EPO_REASONING_SHOW",
+    "Label": "EPO Reason show or hide",
+    "FieldType": "MultiSelectList",
```
geek-soft: Does it have to be multi select list? What if we made it `Text`?

swalker125: Doesn't work with Text

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededAboutToSubmitCallbackControllerTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.test.web.servlet.MvcResult;
+import uk.gov.hmcts.reform.ccd.client.model.AboutToStartOrSubmitCallbackResponse;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+import static uk.gov.hmcts.reform.fpl.utils.ResourceReader.readBytes;
+
+@ActiveProfiles("integration-test")
+@WebMvcTest(CaseSubmissionController.class)
```
geek-soft: Please change to point to right controller. Wired that it works with wrong controller.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededAboutToSubmitCallbackController.java
```diff
@@ -0,0 +1,51 @@
...
+    public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
+        @RequestBody CallbackRequest callbackrequest) {
+        CaseDetails newCaseDetails = callbackrequest.getCaseDetails();
+        CaseDetails oldCaseDetails = callbackrequest.getCaseDetailsBefore();
+
+        Map<String, Object> newData = newCaseDetails.getData();
+        Optional<List<String>> newOrderType = Optional.ofNullable((Map<String, Object>) newData.get("orders"))
+            .map(orders -> (List<String>) orders.get("orderType"));
+
+        Map<String, Object> oldData = oldCaseDetails.getData();
```
geek-soft: Do we rally need to bother with what was previously selected? Could we just operate on latest version and do:
```
1. check if EMERGENCY_PROTECTION_ORDER is selected
2.1. if EMERGENCY_PROTECTION_ORDER exists then set EPO_REASONING_SHOW
2.2. if EMERGENCY_PROTECTION_ORDER does not exist then check whether groundsForEPO exists
2.2.1 if `groundsForEPO` exists then remove it
```

Without bothering with previous state implementation should be simpler.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1840,6 +1848,25 @@ <h3 class="heading heading--two">Other {{ counter }}</h3>
...
   <h3 class="heading heading--two">Grounds for the application</h3>
   <dl class="data-list">
+    {% if case_data.orders is not empty and
+    case_data.orders.orderType is not empty and
+    case_data.orders.orderType contains 'EMERGENCY_PROTECTION_ORDER' %}
+      <div class="data-list__row">
+        <dt class="data-list__title">Grounds for emergency protection order</dt>
+        <dd class="data-list__definition">
+          {% if case_data.groundsForEPO is not empty and
+          case_data.groundsForEPO.EPO_reason is not empty %}
```
geek-soft: We should rename `EPO_reason` to something simpler. A `EPO` prefix is not needed because we operate within context of `GroundsForEPO` object. Maybe simple `reason` or `grounds` be enough.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededAboutToSubmitCallbackController.java
swalker125: I actually think your suggested implementation is more complex. Two ifs is simpler than 3 or 4 nested ifs surely? Happy to chat about it, message me on slack maybe?

service/src/test/resources/submitted-form-pdf-content.txt
```diff
@@ -11,8 +11,8 @@ Contested interim care orderHearing type
 Reason: Baby will be discharged from hospital on 21 June
```
geek-soft: I wished value of that file be greater. At the moment it only detects accidental changes :(

ccd-definition/CaseField.json
geek-soft: I hoped that CONTAINS works with strings.

Disgusting workaround :(

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededAboutToSubmitCallbackControllerTest.java
```diff
@@ -0,0 +1,58 @@
...
+            .perform(post("/callback/orders-needed/about-to-submit")
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(readBytes("fixtures/case.json")))
+            .andExpect(status().isOk())
+            .andReturn();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = MAPPER.readValue(response.getResponse()
+            .getContentAsByteArray(), AboutToStartOrSubmitCallbackResponse.class);
+
+        assertThat(callbackResponse.getData().get("EPO_REASONING_SHOW").toString()).isEqualTo("[SHOW_FIELD]");
```
geek-soft: Please remove `toString`. Cast it to `List<String>` instead.

service/src/integrationTest/resources/fixtures/case.json
```diff
@@ -1,5 +1,16 @@
...
+    "case_data": {
+      "caseLocalAuthority": "example",
+      "orders": {
+        "orderType": [
+          "EMERGENCY_PROTECTION_ORDER"
+        ]
+      }
+    }
+  },
+  "case_details_before": {
```
geek-soft: Please remove `case_details_before`, we do not need it anymore. It just makes fixture unnecessarily more complex.

service/src/integrationTest/resources/fixtures/caseDataWithRemovedEPO.json
```diff
@@ -0,0 +1,29 @@
...
+    "case_data": {
+      "caseLocalAuthority": "example",
+      "groundsForEPO": {
+        "reason": [
+          "PLACED_IN_PROVIDED_ACCOMMODATION"
+        ]
+      }
+    }
+  },
+  "case_details_before": {
```
geek-soft: Same here.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededAboutToSubmitCallbackController.java
```diff
@@ -0,0 +1,44 @@
...
+    public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
+        @RequestBody CallbackRequest callbackrequest) {
+        String epo = "EMERGENCY_PROTECTION_ORDER";
+        CaseDetails caseDetails = callbackrequest.getCaseDetails();
+        Map<String, Object> data = caseDetails.getData();
+        Optional<List<String>> orderType = Optional.ofNullable((Map<String, Object>) data.get("orders"))
+            .map(orders -> (List<String>) orders.get("orderType"));
+
+        if (orderType.toString().contains(epo)) {
+            data.put("EPO_REASONING_SHOW", new String[]{"SHOW_FIELD"});
```
geek-soft: We could use ImmutableList.of("SHOW_FIELD") for consistency.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededAboutToSubmitCallbackController.java
```diff
@@ -0,0 +1,44 @@
...
+public class OrdersNeededAboutToSubmitCallbackController {
+
+    @PostMapping("/about-to-submit")
+    @SuppressWarnings("unchecked")
+    public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
+        @RequestBody CallbackRequest callbackrequest) {
+        String epo = "EMERGENCY_PROTECTION_ORDER";
+        CaseDetails caseDetails = callbackrequest.getCaseDetails();
+        Map<String, Object> data = caseDetails.getData();
+        Optional<List<String>> orderType = Optional.ofNullable((Map<String, Object>) data.get("orders"))
```
geek-soft: Why not to do:

```
Optional.ofNullable((Map<String, Object>) data.get("orders"))
  .map(orders -> (List<String>) orders.get("orderType"))
  .ifPresent(orderTypes -> {
    ... // orderTypes here should be of type List<String>
  })
```

Doing `.toString()` is really nasty. It relies on an implementation of `toString` in `Optional` and `List` classes that can change. Not likely but can change.

ccd-definition/FixedLists.json
```diff
@@ -346,5 +346,29 @@
     "ID": "LitigationCapacityIssues",
     "ListElementCode": "DONT_KNOW",
     "ListElement": "Don't know"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "EPOReasons",
+    "ListElementCode": "PLACED_IN_PROVIDED_ACCOMMODATION",
```
geek-soft: I am not sure we named these values best possible way. Looking at snippet below

```
    "groundsForEPO": {
       "reason": [
         "PLACED_IN_PROVIDED_ACCOMMODATION"
       ]
     }
```

I would understand that EPO is raised because kid is placed in provided accommodation while it really means that it might suffer if NOT kept in provided accommodation. 

IMO we should rename this to be for example:
`POSSIBLE_HARM_IF_NOT_PLACED_IN_PROVIDED_ACCOMMODATION`, `POSSIBLE_HARM_IF_KEPT_IN_CURRENT_ACCOMMODATION`,
`URGENT_ACCESS_TO_CHILD_IS_OBSTRUCTED`

PR #191, opened 2019-01-17
FPL-369: Rename "Open case" event to "Start application"

rebeccabrennan:

https://tools.hmcts.net/jira/browse/FPL-369

### Change description ###

open case changed to Start application
Case Name changed to Case name

```
[ ] Yes
[x ] No
```


ccd-definition/CaseEvent.json
```diff
@@ -3,7 +3,7 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "ID": "openCase",
```
geek-soft: I think it might be good idea to change event ID as well to `startApplication`. Before doing so please check how it impacts existing cases.

1. create new case using `openCase` event
2. rename event and upload new definition
3. check whether you can still see case crated in #1 and that you can see full history.

geek-soft: My prediction was right. You did not notice that "Open case" event disappeared from event history of case that was created before rename :( Please be more vigilant next time, it is in developer hands to ensure backward compatibility.

geek-soft: Please revert event ID rename due to its price :(

geek-soft: @rebeccabrennan since you have issues with machine I carried out that change for you.

PR #212, opened 2019-01-29
Enable and configure GitHub stale bot

geek-soft:

None

### Change description ###

PR enables stale bot that will mark and close stale PRs to reduce resource usage for piece of work that does not progress.

Copied from: https://github.com/hmcts/cnp-rhubarb-recipes-service/blob/master/.github/stale.yml

Documentation: https://github.com/probot/stale

```
[ ] Yes
[x] No
```

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 3
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 4
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: >
+  This issue is being closed automatically as it was stale


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 3
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 4
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
```
swalker125: can we add WIP here to keep our work in progress PRs open?

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
```
swalker125: this is number of days of inactivity before a PR becomes stale?

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 4
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
```
swalker125: this issue or this pull request?

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: >
+  This issue is being closed automatically as it was stale
```
swalker125: same as above

.github/stale.yml
geek-soft: Yes. Activity is defined as:

> Any change to an issues and pull request is considered an update, including comments, changing labels, applying or removing milestones, or pushing commits.

.github/stale.yml
geek-soft: The only way of specifying exceptions is by adding labels. There are no rules available around PR title.

.github/stale.yml
geek-soft: Both really. Configuration will apply to both issues and pull requests. That comment is really copied from example documentation published by bot team.

I am not sure why they say 'issue', maybe because usually PR is associated with an issue. On the other hand it does not make much difference IMO.

PR #273, opened 2019-05-03
Update E2E tests to reflect UI changes in CCD

sivas23:
### Change description ###
fixed failed e2e tests

```
[ ] Yes
[ x] No
```


e2e/fragments/addressPostcodeLookup.js
```diff
@@ -8,14 +8,14 @@ module.exports = {
     postcodeLookup: '#postcodeInput',
     addressList: '#addressList',
     buildingAndStreet: {
-      lineOne: '#AddressLine1',
-      lineTwo: '#AddressLine2',
-      lineThree: '#AddressLine3',
+      lineOne: '#applicant_address__AddressLine1',
```
geek-soft: This is wrong solution. A`addressPostcodeLookup.js` fragment is generic so that it can be used anywhere where address is needed, to provide a way to either look it up or enter manually.

Change here will make it not reusable in the future. Find place where `enterAddressManually` is called and fix it there.

PR #324, opened 2019-06-18
FPLA-121: Extend search cases functionality

rebeccabrennan:
https://tools.hmcts.net/jira/browse/FPLA-121

### Change description ###
Search functionality extended to include Case name, CaseID (now working), State and Date submitted whilst Case creation date search removed.


ccd-definition/SearchResultFields.json
```diff
@@ -9,22 +9,29 @@
...
+    "CaseFieldID": "[STATE]",
+    "Label": "State",
     "DisplayOrder": 3
   },
   {
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
-    "CaseFieldID": "[CREATED_DATE]",
-    "Label": "Case creation date",
+    "CaseFieldID": "caseLocalAuthority",
```
geek-soft: Is that requirement to have different fields ordering? If not I suggest having one order of search fields so user does not have to learn location every time.

ccd-definition/AuthorisationCaseField.json
```diff
@@ -258,13 +258,27 @@
     "UserRole": "caseworker-publiclaw-solicitor",
     "CRUD": "R"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "dateSubmittedOnly",
```
swalker125: can you make it clear that this ID is for searching / filtering. Only means nothing to me

rebeccabrennan: searchAndFilterDate? 
dateForSearchAndFilter?
searchFilterDate?

rebeccabrennan: dateSubmitted -> dateAndTimeSubmitted
dateSubmittedOnly -> dateSubmitted

PR #333, opened 2019-06-20
FPLA-383: Update for migrated respondent structure

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-383

### Change description ###

Added models for MigratedRespondent Object
Updated Respondent MidEvent callback to be compatible with both structures

```
[ ] Yes
[x] No
```


swalker125:
new respondents form
Screenshot_2019-06-25 Core Case Data HMCTS

swalker125:
Tab view for new respondents: 
Screenshot_2019-06-25 Core Case Data HMCTS(1)

(Note Respondents at the top of the tabs. This is a side effect of only being able to set the old respondents to {}) I would assume this will be solved when we carry out the next step of the migration to put respondents1 -> respondents

ccd-definition/AuthorisationCaseField.json
```diff
@@ -482,6 +496,13 @@
     "UserRole": "caseworker-publiclaw-courtadmin",
     "CRUD": "R"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "respondents1",
+    "UserRole": "caseworker-publiclaw-courtadmin",
```
geek-soft: I guess you did not give same to gatekeeper role because no-one is using it in production yet and when we rename 'respondents1' to 'respondents' it will work again as it was.

ccd-definition/CaseField.json
```diff
@@ -251,6 +251,23 @@
     "FieldType": "Respondent",
     "SecurityClassification": "Public"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "respondents1",
+    "Label": "Respondent",
+    "FieldType": "Collection",
+    "FieldTypeParameter": "partyRespondent",
```
geek-soft: Let's remember to rename complex type back to `Respondent` when we removed currently used complex type that holds first* / additional*.

ccd-definition/CaseTypeTab.json
```diff
@@ -26,9 +26,19 @@
     "TabID": "CasePeopleTab",
     "TabLabel": "People in the case",
     "TabDisplayOrder": 2,
-    "CaseFieldID": "children",
+    "CaseFieldID": "respondentsMigrated",
```
geek-soft: Do we need to have it here? I think end user is not interested in knowing whether migration is complete or not.

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,206 @@
     "FieldType": "Document",
     "ElementLabel": "Upload a file",
     "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "telephone",
```
geek-soft: I'd like to have case type names in camel case format like we do for classes in Java. Please rename it here and in other places.

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,206 @@
...
+    "FieldType": "Text",
+    "ElementLabel": "Party ID",
+    "FieldShowCondition": "partyType=\"DO_NOT_SHOW\"",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "party",
+    "ListElementCode": "partyType",
+    "FieldType": "Text",
```
geek-soft: Wouldn't it be better to choose party type from radio group? Allowing users to type it in free text field sounds like a bad idea to me.

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,206 @@
...
+    "FieldType": "telephone",
+    "ElementLabel": "Telephone",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "party",
+    "ListElementCode": "respondent_border_1",
+    "FieldType": "Label",
+    "ElementLabel": "-------------------------------------------------------------------------------------------------------------",
```
geek-soft: I am curious, does the length of it matter? In markdown usually 3 dashes are enough to draw horizontal line. If so I'd rather have 3 instead of the number we have right now. I fear that we end up having lots of horizontal rule type of labels with inconsistent number of dashes.

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,206 @@
...
+    "LiveFrom": "01/01/2017",
+    "ID": "partyRespondent",
+    "ListElementCode": "leadRespondentIndicator",
+    "FieldType": "YesOrNo",
+    "ElementLabel": "Lead Respondent Indicator",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "partyRespondent",
```
geek-soft: I think in the future we should name that complex type to `Respondent` (as I mentioned before)

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,206 @@
...
+    "ListElementCode": "party",
```
geek-soft: I think we should rename that complex type to `RespondentParty` because obviously we will keep it when migration is done and generic name won't work well when we have other instances of party for applicant or others.

ccd-definition/CaseField.json
swalker125: yep - so there's a second step to this migration. Maybe a conversation we need to have, I am not sure how we will remove respondents completely. Might have to go as far as remove ALL data and rebuild. I don't think .remove on the data map will remove it (I mean we need to set it to empty currently)

ccd-definition/CaseTypeTab.json
swalker125: Let me test - I think this may be left over from me trying to get data to appear in the tabs. But unsure

e2e/pages/enterRespondents/enterRespondents.js
```diff
@@ -4,70 +4,74 @@ const postcodeLookup = require('../../fragments/addressPostcodeLookup');
...
     const id = this.state.context;
 
     return {
       respondent: {
-        name: `#respondents_${id}_name`,
-        dob: {
-          day: `#respondents_${id}_dob-day`,
-          month: `#respondents_${id}_dob-month`,
-          year: `#respondents_${id}_dob-year`,
+        party: {
```
geek-soft: I'd remove `party` element as it adds no value. Everything is inside anyway simpler structure such as 👇 is better IMO.

```
respondent: {
    partyType: `#respondents1_${id}_party_partyType`,
    title: `#respondents1_${id}_party_title`,
   ...
}
```

ccd-definition/CaseTypeTab.json
swalker125: can be removed 👍 good spot

ccd-definition/ComplexTypes.json
swalker125: I was populating this in the migration as individual. It can only be individual or organisation correct? Happy to change to another type - shall ask Mia and Eilis now

ccd-definition/ComplexTypes.json
swalker125: shall test now - to be honest I just copied an existing field!

swalker125: you're correct 3 dashes produces the same. I shall change the fields I've added, but let's leave the others that are like this for a separate PR to address. I think this PR is big enough...

ccd-definition/ComplexTypes.json
swalker125: I definitely agree - Respondent is currently taken by the old structure. So we shall do this in step 2 of the migration. Again, let's have a chat about this - hopefully today?

ccd-definition/ComplexTypes.json
swalker125: makes sense to me 👍

e2e/pages/enterRespondents/enterRespondents.js
swalker125: agreed - adds uneccessary nesting

ccd-definition/AuthorisationCaseField.json
swalker125: less unpicking - you happy with this?

ccd-definition/CaseTypeTab.json
swalker125: Keeping in to add temporary show hide to old respondents after migration

ccd-definition/ComplexTypes.json
swalker125: Yes King

service/src/main/java/uk/gov/hmcts/reform/fpl/model/PartyExtended.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import uk.gov.hmcts.reform.fpl.model.common.EmailAddress;
+import uk.gov.hmcts.reform.fpl.model.common.Party;
+import uk.gov.hmcts.reform.fpl.model.common.TelephoneNumber;
+
+import java.util.Date;
+
+@Data
+@EqualsAndHashCode(callSuper = false)
```
geek-soft: Because base class contains some properties we should call methods in base class. Please set it to true.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/PartyExtended.java
```diff
@@ -0,0 +1,54 @@
...
+public final class PartyExtended extends Party {
```
geek-soft: Since that class will remain in the code base after migration to support any operations on respondents please rename it into `RespondentParty`.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/migration/MigratedRespondent.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import uk.gov.hmcts.reform.fpl.model.PartyExtended;
+
+@Data
+@Builder
+@AllArgsConstructor
+public class MigratedRespondent {
+    @JsonProperty("party")
```
geek-soft: I bet it is not needed since field has the same name. Please try to remove it.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentMigrationService.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+@Service
+public class RespondentMigrationService {
+
+    public AboutToStartOrSubmitCallbackResponse setMigratedValue(CaseDetails caseDetails) {
```
geek-soft: Why we need that method and associated callback? Is it to make the show / hide condition work properly while `respondentsMigrated` does not exist prior migration?

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1064,7 +1070,7 @@ <h3 class="heading heading--two">Respondent
       </dd>
     </div>
     <div class="data-list__row">
-      <dt class="data-list__title">Current address</dt>
+      <dt class="data-list__title">Current Address</dt>
```
geek-soft: Is that an expected change? Seems unrelated to migration and not quite right to me?

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1249,7 +1254,7 @@ <h3 class="heading heading--two">Respondent {{counter}}</h3>
           </dd>
         </div>
         <div class="data-list__row">
-          <dt class="data-list__title">Address</dt>
+          <dt class="data-list__title">Current Address</dt>
```
geek-soft: Shouldn't address be lowercase? I am not a content guy so just asking :)

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CoreCaseDataStoreLoader.java
```diff
@@ -25,6 +25,11 @@ public static CaseDetails populatedCaseDetails() throws IOException {
         return mapper.readValue(response, CaseDetails.class);
     }
 
+    public static CaseDetails reformRespondentCaseDetails() throws IOException {
```
geek-soft: I can imagine when migration is done and we clean up old definitions we remove that method and add new structure to `core-case-data-store-api/populated-case-details.json`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentMigrationService.java
```diff
@@ -0,0 +1,46 @@
...
+                    ))
+                );
+                data.put("respondents1", populatedRespondent);
+            }
+
+            return AboutToStartOrSubmitCallbackResponse.builder()
+                .data(data)
+                .build();
+        } else {
+            data.put("respondentsMigrated", "No");
```
geek-soft: Wouldn't be better to default to `Yes` so we capture details in new format as soon as possible and migrate old cases later?

If `No` is a default option then any case created after migration will again have old structure and another migration will be needed.

Am I missing something here?

geek-soft: Ignore me, I misread it. Your test helped me understand behaviour.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/RespondentAboutToStartControllerTest.java
```diff
@@ -0,0 +1,66 @@
...
+@OverrideAutoConfiguration(enabled = true)
+class RespondentAboutToStartControllerTest {
+
+    private static final String AUTH_TOKEN = "Bearer token";
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    @Autowired
+    private MockMvc mockMvc;
+
+    @Autowired
```
geek-soft: All below are not used and should be removed:

```

    @Autowired
    private RespondentMigrationService respondentMigrationService;

    @MockBean
    private IdamApi idamApi;

    @MockBean
    private ServiceAuthorisationApi serviceAuthorisationApi;
```

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/RespondentMidEventControllerTest.java
```diff
@@ -72,6 +75,41 @@ void shouldReturnNoErrorsWhenAllDatesOfBirthAreTodayOrInPast() throws Exception
         assertThat(callbackResponse.getErrors()).doesNotContain(ERROR_MESSAGE);
     }
 
+    @Test
+    void shouldReturnErrorsWhenThereIsNewRespondentAndNoOldRespondent() throws Exception {
```
geek-soft: The test says `NewRespondentAndNoOldRespondent` while only new respondent structure is populated.

Also there is no test for happy scenario. Having this done well now will allow us just to remove old 3 test cases when old structure is removed.

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,206 @@
...
+    "ID": "RespondentParty",
+    "ListElementCode": "respondent_border_3",
+    "FieldType": "Label",
+    "ElementLabel": "---",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "RespondentNew",
+    "ListElementCode": "leadRespondentIndicator",
```
geek-soft: We should make it mandatory if it is mandatory in schema design.

Same for other mandatory fields.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/migration/MigratedRespondent.java
swalker125: I think this is left over from a previous name, good spot 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentMigrationService.java
swalker125: because respondentsMigrated is a hidden field it only exists for the lifecycle of the event. It needs to be set every time you enter the event in order to show the right respondent structure. I extracted to a separate service to unit test that this is working as expected

service/src/main/resources/HTMLTemplate.html
swalker125: Yes. Because of the new format, Current Address is used for each entry in new respondents. I want our fixture file to work with both the old respondent AND the new respondent. This was impossible without the above change. It standardises the format. In the old way, "Current Address" was for first respondent, "Current address" was used for the following.

service/src/main/resources/HTMLTemplate.html
swalker125: maybe so ^ see comment above, I standardised it to C and A. Happy to change to C and a instead. Needs to be one of the other

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CoreCaseDataStoreLoader.java
swalker125: Yes - I did not want to break existing tests with this change.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/RespondentAboutToStartControllerTest.java
swalker125: I was getting test failures without these being set - did have a chat about you with this.

ccd-definition/AuthorisationCaseField.json
geek-soft: Happy. I like your pragmatic approach @swalker125 ;)

ccd-definition/ComplexTypes.json
geek-soft: I wonder what Mia and Eilis think from user perspective however if we need to have it fixed list of choices is better than free text from data mining perspective.

ccd-definition/ComplexTypes.json
geek-soft: Sure, just change fields you adding so others do not spread it even more.

ccd-definition/ComplexTypes.json
swalker125: No - new direction is to make this field hidden and populate in an AboutToSubmit callback. Our users have the following understanding of respondents:

The first respondent (mother) is [name]
The second respondent (father/father of ……………………………) is [name]
The third respondent(s) is/are (the children) by their children’s guardian [name]

Taken from a template provided by Sinead

ccd-definition/CaseEventToFields.json
```diff
@@ -325,6 +325,20 @@
...
+    "CaseEventID": "enterRespondents",
+    "CaseFieldID": "respondentsMigrated",
+    "PageFieldDisplayOrder": 1,
+    "DisplayContext": "OPTIONAL",
+    "PageID": 1,
+    "PageDisplayOrder": 1,
+    "PageColumnNumber": 1,
+    "CallBackURLMidEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/enter-respondents/mid-event",
+    "FieldShowCondition": "respondents1CONTAINS\"DO_NOT_SHOW_THIS_FIELD\""
+
```
Jamesnelson117: Clean up white space

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RespondentController.java
```diff
@@ -0,0 +1,104 @@
...
+                .collect(toList());
+
+            if (migratedRespondents.stream()
+                .map(MigratedRespondent::getParty)
+                .map(Party::getDateOfBirth)
+                .filter(Objects::nonNull)
+                .anyMatch(dob -> dob.after(new Date()))) {
+                errors.add("Date of birth cannot be in the future");
+            }
+        } else {
```
Jamesnelson117: Add else if condition to avoid null exceptions

service/src/main/resources/HTMLTemplate.html
Jamesnelson117: Raise with Mia

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1359,6 +1364,183 @@ <h3 class="heading heading--two">Respondent {{counter}}</h3>
     {% set counter = counter + 1 %}
     {% endfor %}
   {% endif %}
+  {% endif %}
+
```
Jamesnelson117: Fix indentation in this file please, it will make the upcoming Docmosis epic easier if good spacing persists.

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,218 @@
...
+    "ID": "Telephone",
+    "ListElementCode": "telephoneNumber",
+    "FieldType": "Text",
+    "ElementLabel": "Telephone number",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "Telephone",
+    "ListElementCode": "contactDirection",
```
Jamesnelson117: Is it worth including hint text with these new properties? i.e For example, "Please ask for Bob"

Jamesnelson117: contactDirection and usageType are both optional in reform TelephoneNumber. Do we need these?

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1359,6 +1368,187 @@ <h3 class="heading heading--two">Respondent {{counter}}</h3>
...
+        {% else %}
+        -
+        {% endif %}
+      </dd>
+    </div>
+    <div class="data-list__row">
+      <dt class="data-list__title">Telephone number</dt>
+      <dd class="data-list__definition">
+        {% if respondent.value.party.telephoneNumber is not empty %}
+        {% if respondent.value.party.telephoneNumber.telephoneNumber is not empty %}
```
Jamesnelson117: Please remove inner if statement to the outer if i.e 
{% if respondent.value.party.telephoneNumber is not empty and respondent.value.party.telephoneNumber.telephoneNumber is not empty  %}

ccd-definition/ComplexTypes.json
```diff
@@ -1556,5 +1556,218 @@
...
+    "FieldType": "Telephone",
+    "ElementLabel": "Telephone",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "RespondentParty",
+    "ListElementCode": "relationshipLabel",
+    "FieldType": "Label",
+    "ElementLabel": "## Relationship to the child",
```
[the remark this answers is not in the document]
swalker125: UI has already got UX approval 👍

service/src/test/resources/core-case-data-store-api/reform-respondent-case-data.json
```diff
@@ -0,0 +1,471 @@
+{
```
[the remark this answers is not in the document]
swalker125: no - once migration has occured this will be moved into the existing case data json

e2e/pages/enterRespondents/enterRespondents.js
```diff
@@ -4,66 +4,68 @@ const postcodeLookup = require('../../fragments/addressPostcodeLookup');
...
     const id = this.state.context;
 
     return {
       respondent: {
-        name: `#respondents_${id}_name`,
-        dob: {
-          day: `#respondents_${id}_dob-day`,
-          month: `#respondents_${id}_dob-month`,
-          year: `#respondents_${id}_dob-year`,
+        partyType: `#respondents1_${id}_party_partyType`,
```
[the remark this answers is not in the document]
swalker125: partyType has been hidden from view

e2e/pages/enterRespondents/enterRespondents.js
```diff
@@ -4,64 +4,62 @@ const postcodeLookup = require('../../fragments/addressPostcodeLookup');
...
-          option: `#respondents_${id}_contactDetailsHidden-${option}`,
-          reason: `#respondents_${id}_contactDetailsHiddenReason`,
+          option: `#respondents1_${id}_party_contactDetailsHidden-${option}`,
+          reason: `#respondents1_${id}_party_contactDetailsHiddenReason`,
         };
       },
     };
   },
-  addRespondentButton: 'Add new',
+  addRespondentButton: '#respondents1 > div:nth-child(1) > button:nth-child(2)',
```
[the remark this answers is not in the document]
swalker125: why nasty

kubernetes/configurer/add-users-and-roles.sh
```diff
@@ -17,3 +17,4 @@ ${dir}/utils/ccd-add-role.sh caseworker-publiclaw-solicitor
 ${dir}/utils/ccd-add-role.sh caseworker-publiclaw-courtadmin
 ${dir}/utils/ccd-add-role.sh caseworker-publiclaw-cafcass
 ${dir}/utils/ccd-add-role.sh caseworker-publiclaw-gatekeeper
+${dir}/utils/ccd-add-role.sh caseworker-publiclaw-systemupdate
```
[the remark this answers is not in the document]
swalker125: unsure what is meant here

service/src/main/resources/HTMLTemplate.html
swalker125: there is far too much that needs fixing in the HTML file and I do not want large file changes that are unrelated to this story. I have indented the lines you've mentioned, but will not do any more

ccd-definition/ComplexTypes.json
swalker125: removed with new @JsonIgnore annotation

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
```diff
@@ -0,0 +1,24 @@
...
+    public final String firstName;
+    public final String lastName;
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd")
+    public final Date dateOfBirth;
+    public final Address address;
+    public final EmailAddress email;
+    public final TelephoneNumber telephoneNumber;
+}
+
+
```
geek-soft: Please remove extra blank line. One is enough.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Data;
+import uk.gov.hmcts.reform.fpl.model.Address;
+
+import java.util.Date;
+
+@Data
+@AllArgsConstructor
+public class Party {
+    public final String partyID;
+    public final String partyType;
```
geek-soft: Please crate enum for party type and use it here. If not other developers will build their code around string values.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
```diff
@@ -0,0 +1,24 @@
...
+    public final String firstName;
+    public final String lastName;
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd")
+    public final Date dateOfBirth;
+    public final Address address;
+    public final EmailAddress email;
+    public final TelephoneNumber telephoneNumber;
```
geek-soft: I suggest renaming `TelephoneNumber` type to `Telephone` to make it consistent with design. I will be easier to talk in the future.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/RespondentParty.java
```diff
@@ -0,0 +1,55 @@
...
+import lombok.EqualsAndHashCode;
+import uk.gov.hmcts.reform.fpl.model.common.EmailAddress;
+import uk.gov.hmcts.reform.fpl.model.common.Party;
+import uk.gov.hmcts.reform.fpl.model.common.TelephoneNumber;
+
+import java.util.Date;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
geek-soft: I wonder if this is needed here. It might be enough to annotate only common types such as Telephone or EmailAddress.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -0,0 +1,259 @@
...
+            .containsEntry("firstName", "James")
+            .containsEntry("partyType", "INDIVIDUAL");
+
+        assertThat(party.get("partyID")).isNotNull();
+    }
+
+    @Disabled
+    @SuppressWarnings("unchecked")
+    @Test
+    void shouldMapParameterNamesToConstructorArguments() {
```
geek-soft: I do not get what this test is for and it is disabled. Please remove if not needed anymore.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -0,0 +1,259 @@
...
+            .containsEntry("firstName", "James")
+            .containsEntry("partyType", "INDIVIDUAL");
+
+        assertThat(firstParty.get("partyID")).isNotNull();
+
+        assertThat(secondParty)
+            .containsEntry("firstName", "Lucy")
+            .containsEntry("partyType", "INDIVIDUAL");
+
+        assertThat(firstParty.get("partyID")).isNotNull();
```
geek-soft: First respondent is rested twice. It should be `secondParty` here.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -0,0 +1,259 @@
...
+        Map<String, Object> firstValue = (Map<String, Object>) respondents.get(0).get("value");
+        Map<String, Object> secondValue = (Map<String, Object>) respondents.get(1).get("value");
+        Map<String, Object> firstParty = (Map<String, Object>) firstValue.get("party");
+        Map<String, Object> secondParty = (Map<String, Object>) secondValue.get("party");
+
+        assertThat(firstParty).containsEntry("firstName", "James");
+        assertThat(firstParty.get("partyID")).isNotNull();
+
+        assertThat(secondParty).containsEntry("firstName", "Lucy");
+        assertThat(firstParty.get("partyID")).isNotNull();
```
geek-soft: Same here, first party ID is tested twice.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+        Map<String, Object> data = response.getData();
+        List<Map<String, Object>> respondents = (List<Map<String, Object>>) data.get("respondents1");
+        Map<String, Object> firstValue = (Map<String, Object>) respondents.get(0).get("value");
+        Map<String, Object> secondValue = (Map<String, Object>) respondents.get(1).get("value");
+        Map<String, Object> firstParty = (Map<String, Object>) firstValue.get("party");
+        Map<String, Object> secondParty = (Map<String, Object>) secondValue.get("party");
+
+        assertThat(firstParty).containsEntry("firstName", "James");
+        assertThat(firstParty.get("partyID")).isNotNull();
```
geek-soft: If it is important to assure that new ID is added while old one is preserved and old one is known there should be an assertion to make sure existing ID is not overridden.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/RespondentAboutToStartControllerTest.java
geek-soft: I removed these and I am not getting any errors. Can you try again?

service/src/main/resources/HTMLTemplate.html
geek-soft: I understand need to standardise it 👍

service/src/main/resources/HTMLTemplate.html
geek-soft: Checked with Mia, it should be Ca. Please change.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
```diff
@@ -0,0 +1,91 @@
...
+        if (caseDetails.getData().containsKey("respondents1") || !caseDetails.getData().containsKey("respondents")) {
+            data.put("respondentsMigrated", "Yes");
+
+            if (!caseDetails.getData().containsKey("respondents1")) {
+                List<Map<String, Object>> populatedRespondent = new ArrayList<>();
+                populatedRespondent.add(ImmutableMap.of(
+                    "id", UUID.randomUUID().toString(),
+                    "value", ImmutableMap.of(
+                        "party", ImmutableMap.of(
+                            "partyId", UUID.randomUUID().toString()
```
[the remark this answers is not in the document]
swalker125: I think the better fix here is to extract into another method as you pointed out above. Thoughts?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
swalker125: this was the test that was failing - I am concious that my tests do not accurately reflect how the data will come into the system, as I am using builders to construct my data. Thoughts. Happy to remove. It is failing due to mapping exceptions

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
swalker125: I have this unit test to verify this   

void shouldKeepExistingPartyIDWhenAlreadyExists() {

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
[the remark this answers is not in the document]
geek-soft: The comment you added should be near place where partyId is set.

The fact that we need to add party to the list is clear to me but reason we need to generate partyId wasn't.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
geek-soft: Sorry @swalker125 I meant renaming class / type not field. To match design we need:

```
public final Telephone telephoneNumber;
```

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
geek-soft: I understand that test I question says `should keep existing party id and continue adding new party id`. That suggests that keeping existing ID for first entry is important and due to that there should be assertion to guarantee it.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
geek-soft: image

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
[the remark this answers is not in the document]
swalker125: would it be cleared if partyId was just an empty string also?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
swalker125: I see what you're saying now, yes the first assertion should be asserting the id I have set. Made the change 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
```diff
@@ -0,0 +1,93 @@
...
+            if (!caseDetails.getData().containsKey("respondents1")) {
+                // Populates first respondent so UI contains expanded Respondent Object.
+                List<Map<String, Object>> populatedRespondent = new ArrayList<>();
+                populatedRespondent.add(ImmutableMap.of(
+                    "id", UUID.randomUUID().toString(),
+                    "value", ImmutableMap.of(
+                        "party", ImmutableMap.of(
+                            // Variable within CCD party structure must be set to expand Collection.
+                            // PartyId is a hidden field so setting a value will not persist to the db
+                            "partyId", UUID.randomUUID().toString()
```
Jamesnelson117: Should this not be partyID rather than partyId?

kubernetes/configurer/add-users-and-roles.sh
Jamesnelson117: Sorry Sam, I thought there was not a space at the end of this file.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
geek-soft: Good spot but does not really matter as value is never used / persisted.

PR #351, opened 2019-06-27
Update user mapping after IDAM upgrade

geek-soft:

None

### Change description ###

After IDAM upgrade user ID's shifted. PR corrects user mappings.

```
[ ] Yes
[x] No
```


geek-soft:
User list in preview env 👇 

Screenshot 2019-06-27 at 23 33 44

User list in local env 👇 

Screenshot 2019-06-27 at 23 34 48

rebeccabrennan:
Does order of users matter in the different environments?

geek-soft:
Only local and preview are affected. AAT and DEMO does not have this problem because DB is never cleared.

PR #359, opened 2019-07-04
adding address lookup token to tf files

swalker125:
test build to run in pipeline


geek-soft:
What is that change for @swalker125 ?

geek-soft:
Closing as there is no need to set postcode lookup API token for case service.

PR #365, opened 2019-07-12
FPLA-374: Confidential contact details in PDF

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-374

### Change description ###

Added conditional statements for children address, respondent address and telephone and others address and telephone.

Expected behaviour:
- If contact details hidden quesiton is unanswered -> show address
- If contact details hidden question is answered no -> show address
- If contact details hidden question is answered yes -> do not show address

```
[ ] Yes
[x] No
```


service/src/main/resources/HTMLTemplate.html
```diff
@@ -799,24 +803,28 @@ <h3 class="heading heading--two">Child {{ counter }}</h3>
             <dd class="data-list__definition">
               {% if child.value.livingSituation is not empty %}
                 <span class="u-block">{{child.value.livingSituation}}</span>
```
Jamesnelson117: Would you mind adding space between the braces.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -525,24 +525,28 @@ <h3 class="heading heading--two">Child
...
-              <span class="u-comma-before">{{ case_data.children.firstChild.address.PostTown }}</span>
-            {% endif %}
-            {% if case_data.children.firstChild.address.PostCode is not empty %}
-              <span class="u-comma-before">{{ case_data.children.firstChild.address.PostCode }}</span>
-            {% endif %}
-            {% if case_data.children.firstChild.address.Country is not empty %}
-              <span class="u-comma-before">{{ case_data.children.firstChild.address.Country }}</span>
+          {% if case_data.children.firstChild.detailsHidden is not empty and case_data.children.firstChild.detailsHidden == "Yes" %}
+            Confidential
+          {% else %}
```
Jamesnelson117: Is it worth changes these to elseif blocks? Perhaps the condition might be too lengthy to merit this change. Thoughts?

service/src/main/resources/HTMLTemplate.html
```diff
@@ -525,24 +525,28 @@ <h3 class="heading heading--two">Child
         case_data.children.firstChild is not empty and
```
Jamesnelson117: In children we have social worker telephone number, does this need to be confidential also?

service/src/main/resources/HTMLTemplate.html
swalker125: I thought if and then an else would be less code. Essentially I want to show address when detailsHidden is empty or when detailsHidden == No. If I did ifelse then I would have address code twice, no?

service/src/main/resources/HTMLTemplate.html
swalker125: good Q, Abi says no. This just applies to details directly to do with the child

service/src/main/resources/HTMLTemplate.html
swalker125: who called the code police 🙄

service/src/main/resources/HTMLTemplate.html
Jamesnelson117: You would need to do else if detailsHidden is not empty || detailsHidden == "No" and address is not empty...

Bit too lengthy, inner if statement is cleaner.

service/src/main/resources/HTMLTemplate.html
Jamesnelson117: Thanks for checking

service/src/main/resources/HTMLTemplate.html
Jamesnelson117: ;)

service/src/main/resources/HTMLTemplate.html
```diff
@@ -525,24 +525,28 @@ <h3 class="heading heading--two">Child
...
-            {% endif %}
-            {% if case_data.children.firstChild.address.PostTown is not empty %}
-              <span class="u-comma-before">{{ case_data.children.firstChild.address.PostTown }}</span>
-            {% endif %}
-            {% if case_data.children.firstChild.address.PostCode is not empty %}
-              <span class="u-comma-before">{{ case_data.children.firstChild.address.PostCode }}</span>
-            {% endif %}
-            {% if case_data.children.firstChild.address.Country is not empty %}
-              <span class="u-comma-before">{{ case_data.children.firstChild.address.Country }}</span>
+          {% if case_data.children.firstChild.detailsHidden is not empty and case_data.children.firstChild.detailsHidden == "Yes" %}
```
geek-soft: Story says:

> And the current living situation is answered with 'Living with Respondents'

I cannot see condition around `case_data.children.firstChild.livingSituation`

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1066,44 +1074,52 @@ <h3 class="heading heading--two">Respondent
...
-          {% endif %}
-          {% if case_data.respondents.firstRespondent.address.PostCode is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.PostCode }}</span>
-          {% endif %}
-          {% if case_data.respondents.firstRespondent.address.County is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.County }}</span>
-          {% endif %}
-          {% if case_data.respondents.firstRespondent.address.Country is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.Country }}</span>
+        {% if case_data.respondents is not empty and case_data.respondents.firstRespondent is not empty %}
```
geek-soft: The way it is implemented now hides the fact that address was not provided when details were marked confidential. Shouldn't we check whether address exists first and then if exists check whether it is confidential / should be hidden?

AC says

```
Given the address of a respondent is added in the application
And the respondents address field is marked as confidential by saying yes to the question Do you need contact details hidden from other parties? (Optional)
When the application document is generated
Then the address of the respondent will read 'Confidential' on the generated pdf document
And the telephone number of the respondent will read 'Confidential' on the generated pdf document
```

which to me sounds like an action should be taken only when address exists. If not regardless whether confidential or not we should display `-`

service/src/main/resources/HTMLTemplate.html
swalker125: good q, let me double check this is correct

service/src/main/resources/HTMLTemplate.html
swalker125: good spot, that's a better way of implementing

service/src/main/resources/HTMLTemplate.html
swalker125: this should be hide for any address entered. AC to be updated.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -521,11 +521,12 @@ <h3 class="heading heading--two">Child
     <div class="data-list__row">
       <dt class="data-list__title">Child's living situation</dt>
       <dd class="data-list__definition">
-        {% if case_data.children is not empty and
-        case_data.children.firstChild is not empty and
-        case_data.children.firstChild.livingSituation is not empty %}
+        {% if case_data.children is not empty and case_data.children.firstChild is not empty and case_data.children.firstChild.livingSituation is not empty %}
```
geek-soft: I like that one line format better, easier to read IMO despite longer line.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1066,44 +1072,52 @@ <h3 class="heading heading--two">Respondent
...
-          {% if case_data.respondents.firstRespondent.address.PostCode is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.PostCode }}</span>
-          {% endif %}
-          {% if case_data.respondents.firstRespondent.address.County is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.County }}</span>
-          {% endif %}
-          {% if case_data.respondents.firstRespondent.address.Country is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.Country }}</span>
+        {% if case_data.respondents is not empty and case_data.respondents.firstRespondent is not empty %}
+          {% if case_data.respondents.firstRespondent.address is not empty %}
```
geek-soft: IMO that if statement should be joined with previous one otherwise if address is empty no dash will be printed because there is no 'else' statement.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1066,44 +1072,52 @@ <h3 class="heading heading--two">Respondent
...
     <div class="data-list__row">
       <dt class="data-list__title">Telephone number</dt>
       <dd class="data-list__definition">
-        {% if case_data.respondents is not empty and
-        case_data.respondents.firstRespondent is not empty and
-        case_data.respondents.firstRespondent.telephone is not empty %}
-          {{ case_data.respondents.firstRespondent.telephone }}
-        {% else %}
-          -
+        {% if case_data.respondents is not empty and case_data.respondents.firstRespondent is not empty %}
```
geek-soft: This should be changed as well to mark confidential only when telephone number is provided. Currently it prints Confidential label when address is not provided.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1281,10 +1299,14 @@ <h3 class="heading heading--two">Respondent {{counter}}</h3>
         <div class="data-list__row">
           <dt class="data-list__title">Telephone number</dt>
           <dd class="data-list__definition">
-            {% if respondent.value.telephone is not empty %}
-              {{ respondent.value.telephone }}
+            {% if respondent.value.contactDetailsHidden is not empty and respondent.value.contactDetailsHidden == "Yes" %}
```
geek-soft: Same here, it should check whether telephone is empty first.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1602,44 +1624,52 @@ <h3 class="heading heading--two">
...
-          {% if case_data.others.firstOther.address.PostCode is not empty %}
-            <span class="u-comma-before">{{ case_data.others.firstOther.address.PostCode }}</span>
-          {% endif %}
-          {% if case_data.others.firstOther.address.County is not empty %}
-            <span class="u-comma-before">{{ case_data.others.firstOther.address.County }}</span>
-          {% endif %}
-          {% if case_data.others.firstOther.address.Country is not empty %}
-            <span class="u-comma-before">{{ case_data.others.firstOther.address.Country }}</span>
+        {% if case_data.others is not empty and case_data.others.firstOther is not empty %}
+          {% if case_data.others.firstOther.address is not empty %}
```
geek-soft: Same here, it would be safer to have both 'if' statements combined.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1602,44 +1624,52 @@ <h3 class="heading heading--two">
...
     <div class="data-list__row">
       <dt class="data-list__title">Telephone number</dt>
       <dd class="data-list__definition">
-        {% if case_data.others is not empty and
-        case_data.others.firstOther is not empty and
-        case_data.others.firstOther.telephone is not empty %}
-          {{ case_data.others.firstOther.telephone }}
-        {% else %}
-          -
+        {% if case_data.others is not empty and case_data.others.firstOther is not empty %}
```
geek-soft: Same here, not empty check should go before confidential check.

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1807,10 +1841,16 @@ <h3 class="heading heading--two">Other {{ counter }}</h3>
           <div class="data-list__row">
             <dt class="data-list__title">Telephone number</dt>
             <dd class="data-list__definition">
-              {% if other.value is not empty and other.value.telephone is not empty %}
-                {{ other.value.telephone }}
-              {% else %}
-                -
+              {% if other.value is not empty %}
+                {% if other.value.detailsHidden is not empty and other.value.detailsHidden == "Yes" %}
```
geek-soft: Same here, first check whether not empty then check whether confidential.

service/src/main/resources/HTMLTemplate.html
swalker125: same, I find it very confusing to have these multiline if statements

service/src/main/resources/HTMLTemplate.html
swalker125: The else for the address if is on line 1102 with the dash

swalker125: Am I understanding this correctly?

service/src/main/resources/HTMLTemplate.html
swalker125: good spot, forgot to propagate the change I did to address onto telephone

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1066,29 +1072,31 @@ <h3 class="heading heading--two">Respondent
...
-          {% if case_data.respondents.firstRespondent.address.PostCode is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.PostCode }}</span>
-          {% endif %}
-          {% if case_data.respondents.firstRespondent.address.County is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.County }}</span>
-          {% endif %}
-          {% if case_data.respondents.firstRespondent.address.Country is not empty %}
-            <span class="u-comma-before">{{ case_data.respondents.firstRespondent.address.Country }}</span>
+        {% if case_data.respondents is not empty and case_data.respondents.firstRespondent is not empty and case_data.respondents.firstRespondent.address is not empty %}
+          {% if case_data.respondents.contactDetailsHidden is not empty and case_data.respondents.contactDetailsHidden == "Yes" %}
```
geek-soft: Shouldn't that be `case_data.respondents.firstRespondent.contactDetailsHidden`?

service/src/main/resources/HTMLTemplate.html
```diff
@@ -1098,10 +1106,12 @@ <h3 class="heading heading--two">Respondent
     <div class="data-list__row">
       <dt class="data-list__title">Telephone number</dt>
       <dd class="data-list__definition">
-        {% if case_data.respondents is not empty and
-        case_data.respondents.firstRespondent is not empty and
-        case_data.respondents.firstRespondent.telephone is not empty %}
-          {{ case_data.respondents.firstRespondent.telephone }}
+        {% if case_data.respondents is not empty and case_data.respondents.firstRespondent is not empty and case_data.respondents.firstRespondent.telephone is not empty %}
+          {% if case_data.respondents.contactDetailsHidden is not empty and case_data.respondents.contactDetailsHidden == "Yes" %}
```
geek-soft: Same here? Didn't you forget to have `firstRespondent` segment in the middle?

service/src/main/resources/HTMLTemplate.html
```diff
@@ -798,27 +800,31 @@ <h3 class="heading heading--two">Child {{ counter }}</h3>
...
-                    <span class="u-comma-before">{{ child.value.address.AddressLine1 }}</span>
-                  {% endif %}
-                  {% if child.value.address.AddressLine2 is not empty %}
-                    <span class="u-comma-before">{{ child.value.address.AddressLine2 }}</span>
-                  {% endif %}
-                  {% if child.value.address.AddressLine3 is not empty %}
-                    <span class="u-comma-before">{{ child.value.address.AddressLine3 }}</span>
+                <span class="u-block">{{ child.value.livingSituation }}</span>
+                  {% if child.value.address is not empty %}
+                    {% if child.value.detailsHidden is not empty and child.value.detailsHidden == "Yes"%}
```
geek-soft: Since you make other changes it would be good to add whitespace before closing `}`

PR #377, opened 2019-07-24
refactor: use custom types in respondent methods

swalker125:


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RespondentController.java
```diff
@@ -30,55 +31,61 @@
...
-        return respondentService.addHiddenValues(caseDetails);
+        return AboutToStartOrSubmitCallbackResponse.builder()
+            .data(caseDetails.getData())
+            .build();
     }
 
-    @SuppressWarnings("unchecked")
     private List<String> validate(CaseDetails caseDetails) {
         ImmutableList.Builder<String> errors = ImmutableList.builder();
+        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
```
Jamesnelson117: Is this needed? Can we not replace the else if below with just` else if (caseData.getRespondents() != null)`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
```diff
@@ -1,93 +1,70 @@
...
+                            .partyId(UUID.randomUUID().toString())
+                            .partyType(PartyType.INDIVIDUAL).build());
+                    } else {
+                        respondentBuilder.party(element.getValue().getParty().toBuilder().build());
                     }
 
-                    return partyBuilder.build();
+                    return Element.<MigratedRespondent>builder()
+                        .id(element.getId())
+                        .value(respondentBuilder.leadRespondentIndicator("No").build())
```
Jamesnelson117: Set the leadRespondentIndicator outside of the return

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -1,222 +1,181 @@
 package uk.gov.hmcts.reform.fpl.service;
```
Jamesnelson117: Add an assertion for leadRespondent in addHiddenFields

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RespondentController.java
swalker125: Yes, converts my data into a more usable format rather than needing to operate on maps

PR #397, opened 2019-08-09
Use helm charts in preview environment

geek-soft:

https://tools.hmcts.net/jira/browse/FPLA-357

### Change description ###

This PR removes native Kubernetes deployment file and replaces it with helm charts.

## Environment changes

Structure of preview environment is changed in the following ways:
- environment points to strategic IDAM in AAT instead of deploying it's own instance in AKS
- environment points to S2S service in AAT instead of deploying it's own instance in AKS
- environment points to document store in AAT instead of deploying it's own instance in AKS

Below is structure of new preview environment:

image

## Scripts changes

Scripts that are used to config environments (including local) has changed) in the following way:
- utility scripts are preconfigured to target localhost environment by default now that allowed me to removed few scripts
- exciting scripts got split slightly different into the following ones and moved to `bin/configurer` directory:
  - `add-oauth-clients.sh` - adds IDAM OAuth2 clients
  - `add-roles.sh` - adds roles to IDAM (only in tactical IDAM) and registers these IDAM roles with CCD
  - `add-users.sh` - adds users to IDAM
  - `import-ccd-definition.sh` - imports CCD definition

Since IDAM instance is reused in preview environment CI pipeline uses only the following scripts that aim to configure private instance of CCD:
 - `bin/configurer/add-roles.sh`
 - `bin/configurer/import-ccd-definition.sh`

```
[ ] Yes
[x] No
```


bin/configurer/utils/idam-lease-user-token.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -eu
+
+if [[ "${IDAM_VERSION:-tactical}" == 'tactical' ]]; then
```
geek-soft: It is bit of technical debt. In one of next steps we will stop using tactical IDAM and that will be the time when we remove this tech debt.

e2e/steps_file.js
```diff
@@ -67,6 +67,7 @@ module.exports = function () {
 
     signOut() {
       this.click('Sign Out');
+      this.wait(2); // in seconds
```
akomisarek: Waiting for? Again, couldn't we do this a bit nicer? Maybe we could open Tech debt to get your changes over the line quicker and ensure we are getting rid of those wait(2)?

charts/fpl-case-service/values.yaml
```diff
@@ -0,0 +1,16 @@
...
+    IDAM_S2S_AUTH_TOTP_SECRET: fake-secret
+    CORE_CASE_DATA_API_URL: http://ccd-data-store-api-aat.service.core-compute-aat.internal
+    DOCUMENT_MANAGEMENT_URL: http://dm-store-aat.service.core-compute-aat.internal
+    CCD_UI_BASE_URL: https://ccd-case-management-web-aat.service.core-compute-aat.internal
+    NOTIFY_API_KEY: fake-key
+    FPL_LOCAL_AUTHORITY_EMAIL_TO_CODE_MAPPING: hmcts.net=>FPLA
+    FPL_LOCAL_AUTHORITY_CODE_TO_NAME_MAPPING: FPLA=>FPLA team
+    FPL_LOCAL_AUTHORITY_USER_MAPPING: FPLA=>fake-user-id
+    FPL_LOCAL_AUTHORITY_CODE_TO_HMCTS_COURT_MAPPING: FPLA=>FPLA Court:FamilyPublicLaw+fpla-court@gmail.com
+    FPL_LOCAL_AUTHORITY_CODE_TO_CAFCASS_MAPPING: FPLA=>FPLA CAFCASS:FamilyPublicLaw+fpla-cafcass@gmail.com
```
akomisarek: Don't we need to configure APPINSIGHTS_INSTRUMENTATIONKEY?

charts/fpl-case-service/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+name: fpl-case-service
+home: https://github.com/hmcts/fpl-ccd-configuration
+version: 0.1.0
+description: FPL Case Service
+maintainers:
+  - name: HMCTS Family Public Law team
+    email: fpl-developers@HMCTS.NET
```
akomisarek: Am I on that group as well? If not probably I should be added!

bin/variables/unload-environment-variables.sh
```diff
@@ -0,0 +1,17 @@
...
+
+echo 'unset IDAM_VERSION'
+
+# urls
+echo 'unset SERVICE_AUTH_PROVIDER_API_BASE_URL'
+echo 'unset IDAM_API_BASE_URL'
+echo 'unset CCD_DEFINITION_STORE_API_BASE_URL'
+
+# secrets
+echo 'unset CCD_API_GATEWAY_S2S_SECRET'
```
akomisarek: Any reason why this doesn't clear all of the secrets? I.e. CCD_DEFINITION_STORE_S2S_SECRET this one is not cleared.

bin/configurer/utils/idam-add-role.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -eu
+
+name=${1}
+description=${2}
+
+docker run --rm --network ${DATABASE_NETWORK:-compose_default} postgres:11-alpine psql --host ${DATABASE_HOST:-ccd-shared-database} --username postgres --command "INSERT INTO role(name, display_name) VALUES('${name}', '${description}') ON CONFLICT DO NOTHING" idam
```
akomisarek: Probably silly question, but don't we have some kind of API to do it? Going directly to the IDAM db looks a bit smelly. Is it something being changes as part of new IdAM solution as well?

PR #405, opened 2019-08-13
refactor: removed old respondent definitions

swalker125:
This commit removes old "respondents" definitions and any reference in the code base to the old definitions.


service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
```diff
@@ -22,29 +22,23 @@
...
-                            // PartyId is a hidden field so setting a value will not persist to the db
-                            "partyId", UUID.randomUUID().toString()
-                        )
-                    ))
-                );
-                data.put("respondents1", populatedRespondent);
-            }
-        } else {
-            data.put("respondentsMigrated", "No");
+        if (!caseDetails.getData().containsKey("respondents1")) {
```
akomisarek: @swalker125 - you modified that code, however by removing the existing test those lines of code are not longer tested in the below test (RespondentServiceTest). Are you happy with the coverage provided by the controller level test?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -134,8 +101,11 @@ void shouldAddPartyIDAndPartyTypeValuesToMultipleRespondents() {
...
     void shouldNotAddPartyIDAndPartyTypeValuesToDataStructureIfRespondents1IsNotPresent() {
+        Map<String, Object> data = new HashMap<>();
+        data.put("respondent", "data");
+
         CaseDetails caseDetails = CaseDetails.builder()
-            .data(createData("respondent", "data"))
+            .data(data)
             .build();
 
         AboutToStartOrSubmitCallbackResponse response = service.addHiddenValues(caseDetails);
```
akomisarek: Maybe that should invoke the expand method? Doesn't look like anything in the test is invoking it, thus it is not tested properly.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Respondents.java
```diff
@@ -1,49 +1,13 @@
...
-
-import java.util.Arrays;
-import java.util.List;
-import java.util.UUID;
-import java.util.stream.Collectors;
 
 @Data
 @Builder
-@JsonIgnoreProperties(ignoreUnknown = true)
+@AllArgsConstructor
```
akomisarek: Are we happy with losing the @JsonProperty constructors?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RespondentController.java
```diff
@@ -68,35 +65,22 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
     private List<String> validate(CaseDetails caseDetails) {
         ImmutableList.Builder<String> errors = ImmutableList.builder();
 
-        Map<String, Object> respondentsData =
-            (Map<String, Object>) defaultIfNull(caseDetails.getData().get("respondents"), null);
-
         if (caseDetails.getData().containsKey("respondents1")) {
```
akomisarek: I don't like we removed else statement, but still checking for the respondents1.

What if respondents1 is not there? Should we fail? 
If we are sure all cases are migrated, do we need this if statement?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RespondentController.java
```diff
@@ -68,35 +65,22 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
...
-
             List<Map<String, Object>> migratedRespondentObject =
                 (List<Map<String, Object>>) caseDetails.getData().get("respondents1");
 
-            List<MigratedRespondent> migratedRespondents = migratedRespondentObject.stream()
+            List<Respondents> respondents = migratedRespondentObject.stream()
```
akomisarek: Do we still want to call them "migratedRespondent" - the class is changed after all.

akomisarek:
This change fails in Jenkins - can you please double check what's wrong? Might be reason why my PR is failing as well.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
geek-soft: Good spot Adam. At this point we just vaguely ensure that `respondents1` is populated while it does not really ensure that returned value has what we expect it to have. The purpose of the method is to return list structure with respondent that has partyID populated to make CCD render expanded collection.

Since it is 5m effort to write such test I'd suggest adding unit test that checks returned collection has:
- size of 1
- element 1 has party with party ID populated

swalker125: just written the same tests for applicants, shall add a sad and happy test now 👍

swalker125: Just a heads up - this branch is awaiting a refactoring to move away from operating on maps which is being addressed separately. I shall write the test using the current methods and update in future

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RespondentController.java
swalker125: I have removed the if. When this callback is fired respondents1 will always be populated

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RespondentController.java
swalker125: removed all reference to migration in controller

ccd-definition/CaseEventToFields.json
```diff
@@ -325,42 +325,16 @@
...
-    "LiveFrom": "01/01/2017",
-    "CaseTypeID": "CARE_SUPERVISION_EPO",
-    "CaseEventID": "enterRespondents",
-    "CaseFieldID": "respondentsMigrated",
-    "PageFieldDisplayOrder": 1,
-    "DisplayContext": "OPTIONAL",
-    "PageID": 1,
-    "PageDisplayOrder": 1,
-    "PageColumnNumber": 1,
-    "CallBackURLMidEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/enter-respondents/mid-event",
```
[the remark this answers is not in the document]
swalker125: done

PR #414, opened 2019-08-19
Upgrade codeceptjs, remove conflicting reporters and stop storing state in page objects

geek-soft:

### Change description ###

```
[ ] Yes
[ ] No
```


e2e/pages/enterOtherProceedings/enterOtherProceedings.js
```diff
@@ -1,73 +1,75 @@
...
-      throw new Error('Adding more proceedings is not supported in the test');
-    }
-
     I.click(this.addProceedingButton);
-    this.state.contextPrefix = 'additionalProceedings_0_';
   },
 
   selectYesForProceeding() {
-    I.click(this.fields().onGoingProceeding.yes);
+    I.click(this.fields(undefined).onGoingProceeding.yes);
```
akomisarek: Do we need to pass the undefined explicitly? It would be undefined if we don't pass anything, wouldn't it?

geek-soft: You are right, I just preferred to be explicit.

PR #417, opened 2019-08-20
Refactor E2E tests

geek-soft:

### Change description ###

```
[ ] Yes
[ ] No
```


e2e/pages/events/enterOthersEvent.page.js
```diff
@@ -47,7 +47,11 @@ module.exports = {
     I.selectOption(this.fields(elementIndex).gender, other.gender);
     I.fillField(this.fields(elementIndex).birthPlace, other.birthPlace);
     within(this.fields(elementIndex).address, () => {
-      postcodeLookup.lookupPostcode(other.address);
+      if (elementIndex === 'firstOther') {
```
akomisarek: A bit inconsistent - in the previous file (enterChildrenEvent.page.js) the elementIndex was index and compared to 0. Here we have got some string, shouldn't this be named differently? (I know you haven't introduced this variable, but as this is refactoring change, maybe would be worth improving?)

e2e/steps_file.js
```diff
@@ -85,34 +102,39 @@ module.exports = function () {
       this.wait(2); // in seconds
```
akomisarek: Maybe we should wait for login page to appear instead of two secs?

e2e/tests/accessSegregation_test.js
```diff
@@ -2,51 +2,62 @@ const config = require('../config.js');
...
-  I.signOut();
+  if (!caseId) {
+    await I.logInAndCreateCase(config.swanseaLocalAuthorityEmailUserOne, config.localAuthorityPassword);
+    await I.enterMandatoryFields();
+    caseViewPage.goToNewActions(config.applicationActions.submitCase);
+    submitApplicationEventPage.giveConsent();
+    I.continueAndSubmit();
+
+    // eslint-disable-next-line require-atomic-updates
+    caseId = await I.grabTextFrom('.heading-h1');
```
akomisarek: Hmm, what about reading caseId from URL? I guess two options are not perfect, but maybe URL more consistent?

e2e/tests/accessSegregation_test.js
```diff
@@ -2,51 +2,62 @@ const config = require('../config.js');
...
+    console.log(`Case ${caseId} has been submitted`);
```
akomisarek: Do we need the console loging?

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -0,0 +1,84 @@
...
+  if (!caseId) {
+    await I.logInAndCreateCase(config.swanseaLocalAuthorityEmailUserOne, config.localAuthorityPassword);
+    await I.enterMandatoryFields();
+    caseViewPage.goToNewActions(config.applicationActions.submitCase);
+    submitApplicationEventPage.giveConsent();
+    I.continueAndSubmit();
+
+    // eslint-disable-next-line require-atomic-updates
+    caseId = await I.grabTextFrom('.heading-h1');
+    console.log(`Case ${caseId} has been submitted`);
```
akomisarek: Do we need it?

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -0,0 +1,84 @@
...
+
+    I.wait(2); // in seconds; time needed for access grant calls to complete
```
akomisarek: Do we need to wait for access calls here? We are not testing the access segregation here. If we don't does it make this test flaky?

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -0,0 +1,84 @@
...
+  uploadStandardDirectionsDocumentEventPage.uploadAdditionalDocuments(config.testFile);
+  I.continueAndSave();
+  I.seeEventSubmissionConfirmation(config.applicationActions.uploadDocuments);
+  caseViewPage.selectTab(caseViewPage.tabs.documents);
+  I.see('mockFile.txt');
+  I.seeAnswerInTab('1', 'Other documents 1', 'Document name', 'Document 1');
+  I.seeAnswerInTab('2', 'Other documents 1', 'Upload a file', 'mockFile.txt');
+  I.seeAnswerInTab('1', 'Other documents 2', 'Document name', 'Document 2');
+  I.seeAnswerInTab('2', 'Other documents 2', 'Upload a file', 'mockFile.txt');
+});
```
akomisarek: Didn't we lose coverage of old test 'Local authority can see standard directions in documents' or I am not able to find it. Was it unnecessary?

e2e/tests/localAuthorityDeletesApplication_test.js
```diff
@@ -0,0 +1,40 @@
...
+  I.waitForElement('.error-summary-list');
+  I.see('Tell us the status of all documents including those that you haven\'t uploaded');
+  I.see('You need to add details to orders and directions needed');
+  I.see('You need to add details to children');
+  I.see('You need to add details to applicant');
+  I.see('You need to add details to hearing needed');
+  I.see('You need to add details to grounds for the application');
+});
+
+Scenario('local authority deletes application', (I, caseViewPage, deleteApplicationEventPage, eventSummaryPage) => {
```
akomisarek: Aren't you introducing order dependency in those tests now? If for any reason the second test runs first, the first one is going to fail as the created case would be deleted. I generally believe rely on ordering of the tests is bad.

e2e/tests/localAuthorityMaintainsCaseAfterSubmission_test.js
```diff
@@ -0,0 +1,52 @@
...
+  I.seeDocument('Social work chronology', '', 'To follow', 'mock reason');
+  I.seeDocument('Social work statement and genogram', 'mockFile.txt', 'Attached');
+  I.seeDocument('Social work assessment', 'mockFile.txt', 'Attached');
+  I.seeDocument('Care plan', 'mockFile.txt', 'Attached');
+  I.seeDocument('Social work evidence template (SWET)', 'mockFile.txt', 'Attached');
+  I.seeDocument('Threshold document', 'mockFile.txt', 'Attached');
+  I.seeDocument('Checklist document', 'mockFile.txt', 'Attached');
+});
+
+Scenario('local authority uploads court bundle', (I, uploadDocumentsEventPage, submitApplicationEventPage, caseViewPage) => {
```
akomisarek: Hmm, wasn't this blocked for local authority prior the submission? 
```
Scenario('Ability for a local authority to upload court bundle only after case is submitted', async (I, uploadDocumentsEventPage, submitApplicationEventPage, caseViewPage) => {
  I.dontSee('Court bundle');
  I.continueAndSave();
  I.seeEventSubmissionConfirmation(config.applicationActions.uploadDocuments);

```

e2e/tests/localAuthoritySubmitsApplication_test.js
```diff
@@ -0,0 +1,462 @@
...
+});
+
+Scenario('local authority tries to submit without giving consent', (I, caseViewPage) => {
+  caseViewPage.goToNewActions(config.applicationActions.submitCase);
+  I.see(`I, ${config.swanseaLocalAuthorityUserOne.forename} ${config.swanseaLocalAuthorityUserOne.surname}, believe that the facts stated in this application are true.`);
+  I.click('Continue');
+  I.seeInCurrentUrl('/submitApplication');
+});
+
+Scenario('local authority submits after giving consent', (I, caseViewPage, submitApplicationEventPage) => {
```
akomisarek: We rely on the order, are we 100% sure runner is not going to change order in which tests are run? Are we sure the last one is going to be run last?

e2e/helpers/puppeter_helper.js
```diff
@@ -8,4 +8,29 @@ module.exports = class PuppeteerHelpers extends Helper {
...
+   * Grabs text from a element specified by locator.
+   *
+   * Note: When error is not found undefined is returned. That behaviour is bit different from a behaviour of
+   * default `grabTextFrom` function which throws an error and fails test when element is not found.
+   *
+   * @param locator - element locator
+   * @returns {Promise<string|undefined>}
+   */
+  async grabText(locator) {
+    const elements = await this.helpers['Puppeteer']._locate(locator);
```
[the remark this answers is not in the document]
geek-soft: Not in foreseen future.

If so that one puppeteer helper needs to be reimplemented, hopefully just one.

e2e/helpers/puppeter_helper.js
```diff
@@ -8,4 +8,29 @@ module.exports = class PuppeteerHelpers extends Helper {
...
+
+    const texts = elements.map(async (element) => {
+      return await (await element.getProperty('innerText')).jsonValue();
```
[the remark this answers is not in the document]
geek-soft: I do not think so, will be removed in a moment.

e2e/pages/events/enterOthersEvent.page.js
geek-soft: A `firstOther` forms part of the locator which is driven by data structures in CCD. This is old structure with `{ firstX: {}, additionalX: [{}, {}] }`. In two other places we have just a collection that when CCD UI is rendered uses item position (0-x) to uniquely name DOM elements.

Things like `first / additional` should disappear with time.

e2e/steps_file.js
geek-soft: Will try it and comment finding back.

e2e/tests/accessSegregation_test.js
geek-soft: Not a good reason for change for me. Reading heading worked well from the beginning.

e2e/tests/accessSegregation_test.js
geek-soft: I thought it will give some additional information in case someone wants to grab it and investigate state of the case. I used it few times but it can sometimes be deducted from execution steps (if enabled - not enabled on Jenkins).

It is not ideal when many workers are utilised because you do not know what tests is using created case. I do not feel strong, can remove if you want.

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
geek-soft: Logically seems not needed, I might have left it by mistake. Will remove and report back.

It should not have any negative impact, just 2 more secs to exec time.

e2e/steps_file.js
geek-soft: After adding regular wait for content tests started failing on:

```
Sign in as local authority:
     Text "Sign in" was not found on page after 7.5 sec
Evaluation failed: TypeError: Cannot read property 'innerText' of null
```

Not sure, screenshots show login page correctly. Anyway that wait of 2 seconds seemed to be fairly robust ;)

e2e/tests/localAuthorityDeletesApplication_test.js
geek-soft: I did that on purpose, you can call it design decision. I am aware that generally dependencies between tests are bad thing but here I think I managed to use it to our advantage. The way I see this is that benefits are greater than cost.

Described scenario will not happen in current implementation of framework. Tests are always executed from top to bottom.

e2e/tests/localAuthorityMaintainsCaseAfterSubmission_test.js
geek-soft: I found `I.dontSee('Court bundle')` assertion useless so removed it. Good spot though.

e2e/tests/localAuthoritySubmitsApplication_test.js
geek-soft: Bit like trying to predict future. I only know it was never the case in the past for a long time. 

Framework is based on mocha which does not support random test ordering.

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
geek-soft: I removed it as we do not test visibility of elements in other roles, we just test the fact people have access to case as a whole. I am open for discussion whether we should be testing such things or not.

akomisarek: Something we can discuss together with Sam and Raghu.

e2e/tests/localAuthorityMaintainsCaseAfterSubmission_test.js
akomisarek: That's right the case is already submitted so the test is ok

e2e/tests/localAuthoritySubmitsApplication_test.js
akomisarek: Ok that's fine I was not aware.

e2e/tests/accessSegregation_test.js
akomisarek: I am happy with those.

PR #423, opened 2019-08-23
Move dependency check and linting of E2E tests to Jenkins and disable Travis buils

geek-soft:

None

### Change description ###

Travis builds take a lot of time due to queue size. Moving to GitHub actions for purpose of integrity check and listing does not make much sense. Integrity check should also happen on Jenkins - it does not make sense to run this on external CI only. 

This PR moves dependency check and listing to Jenkins and disables Travis builds.

```
[ ] Yes
[x] No
```


service/build.gradle
```diff
@@ -80,11 +88,11 @@ task runFunctionalTest(type: Exec, description: 'Runs E2E tests.') {
 }
 
 task smoke(description: 'Runs the smoke tests.') {
-  dependsOn(awaitApplicationReadiness, installDependencies, runSmokeTests)
+  dependsOn(awaitApplicationReadiness, checkDependenciesIntegrity, installDependencies, lintUserInterfaceTests, runSmokeTests)
```
akomisarek: Those are checking the JS dependencies and doing JS linting. Is it correct to have them in that level as this is service level abstraction?

geek-soft: Probably not ideal. We had to do it this way because by convention CI for Java projects expects to run `gradle smoke` or `gradle functional` to run these tests. 

Alternatively we could sort:
1. implement functional tests in Java that would use rest assured for example and run E2E tests in after smoke / after functional callbacks as an addition.
2. move CCD configuration outside of FPL callback service to separate repo alongside with E2E tests (something that we planned one day)

akomisarek: Sure, thanks. Make sense for now then!

PR #429, opened 2019-08-27
improvement: align with master ccd-docker and use sidam locally 

akomisarek:

### Change description ###

```
[ ] Yes
[ ] No
```


bin/configurer/utils/idam-authenticate.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -eu
+
+USERNAME=$1
+PASSWORD=$2
+
+curl --silent --show-error --header 'Content-Type: application/x-www-form-urlencoded' --header 'Accept: application/json' -d "username=${USERNAME}&password=${PASSWORD}" "${IDAM_API_BASE_URL:-http://localhost:5000}/loginUser" | docker run --rm --interactive stedolan/jq -r .api_auth_token
```
geek-soft: Isn't that similar / simpler version of `idam-lease-user-token.sh`? I wonder whether we could use that one over `idam-lease-user-token.sh`?

bin/configurer/setup-local-idam.sh
```diff
@@ -0,0 +1,21 @@
...
+
+set -eu
+
+dir=$(dirname ${0})
+
+# Defaults are ok and create require ccd_gateway service in idam
+#${dir}/utils/create-ccd-service.sh
+
+# citizen / solicitor created by default
+${dir}/utils/idam-add-role.sh "ccd-import"
```
geek-soft: There is `add-roles.sh` script already that is supposed to register roles that service needs with IDAM and CCD. I'd recommend moving these few lines there.

I think having one 'go-to' file in case new role is added is better than having two files.

bin/configurer/setup-local-idam.sh
```diff
@@ -0,0 +1,21 @@
...
+${dir}/utils/idam-add-role.sh "caseworker"
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw"
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw-solicitor"
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw-courtadmin"
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw-cafcass"
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw-gatekeeper"
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw-systemupdate"
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw-judiciary"
+
+${dir}/utils/idam-create-caseworker.sh ccd.docker.default@hmcts.net ccd-import CCD_Docker
```
geek-soft: Same here, it could be moved to `add-users.sh` which manages users needed by the service regardless how that user is used.

bin/configurer/setup-local-idam.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -eu
+
+dir=$(dirname ${0})
+
+# Defaults are ok and create require ccd_gateway service in idam
+#${dir}/utils/create-ccd-service.sh
```
geek-soft: Is that needed at any point?

bin/configurer/utils/idam-authenticate.sh
akomisarek: Ok let me test that!

akomisarek: Unfortunately I was not able to use it, I have asked for support IDAM guys but using the token obtained from the user login gave me 403. 

However the oath2/authorize is deprecated so we should use probably something else, but it doesn't seem loginUser is way to go...

bin/configurer/setup-local-idam.sh
akomisarek: Ok, I separated the IDAM stuff here, but I think you are right I will refactor a bit and put all roles in single place.

bin/configurer/setup-local-idam.sh
akomisarek: Ok, I will add this to the users part.

bin/configurer/setup-local-idam.sh
akomisarek: Yes, I have uncommented it's the initial service we need, refactor a bit though.

bin/configurer/add-roles.sh
```diff
@@ -4,13 +4,22 @@ set -eu
...
+${dir}/utils/idam-add-role.sh "caseworker-publiclaw"
 
-${dir}/utils/ccd-add-role.sh caseworker-publiclaw-solicitor
-${dir}/utils/ccd-add-role.sh caseworker-publiclaw-courtadmin
-${dir}/utils/ccd-add-role.sh caseworker-publiclaw-cafcass
-${dir}/utils/ccd-add-role.sh caseworker-publiclaw-gatekeeper
-${dir}/utils/ccd-add-role.sh caseworker-publiclaw-systemupdate
-${dir}/utils/ccd-add-role.sh caseworker-publiclaw-judiciary
+# User used during the CCD import and ccd-role creation
+${dir}/utils/idam-create-caseworker.sh ccd.docker.default@hmcts.net ccd-import CCD_Docker
```
geek-soft: Any reason why args are not in quotes? Looks bit inconsistent.

geek-soft: A `CCD_Docker` is meaningless here. Can we remove? I tend to hide things that does not matter.

bin/configurer/add-services.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -eu
+
+dir=$(dirname ${0})
+
+# Defaults are ok and create require ccd_gateway service in idam
```
geek-soft: I am not sure how much value that comment adds. I'd remove it.

akomisarek: Done.

bin/configurer/add-services.sh
```diff
@@ -0,0 +1,11 @@
...
+${dir}/utils/idam-create-service.sh ccd_gateway false ccd_gateway ccd_gateway_secret [\"http://localhost:3451/oauth2redirect\"]
```
geek-soft: Can we move `false` inside the script and make it hardcoded there? It will be always false for us.

bin/configurer/add-roles.sh
akomisarek: Done

bin/configurer/add-services.sh
geek-soft: Can we make redirect URL a single value? That would allow us have:

`${dir}/utils/idam-create-service.sh ccd_gateway false ccd_gateway ccd_gateway_secret http://localhost:3451/oauth2redirect`.

It is easier to read.

bin/configurer/utils/idam-create-service.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -eu
+
+dir=$(dirname ${0})
+
+LABEL="${1:-ccd_gateway}"
```
geek-soft: Script is reused for two services now. IMO it should not have defaults for arguments that matter and require change. I'd rather fail fast

bin/configurer/utils/idam-lease-user-token.sh
```diff
@@ -2,21 +2,11 @@
 
 set -eu
 
-if [[ "${IDAM_VERSION:-tactical}" == 'tactical' ]]; then
-  id=${1}
-  role=${2}
+importerUsername=${CCD_CONFIGURER_IMPORTER_USERNAME:-ccd.docker.default@hmcts.net}
```
geek-soft: The same here, IMO it should not use defaults for critical arguments to limit potential surprise especially that scripts does not really report what was created.

README.md
```diff
@@ -24,11 +24,20 @@ Creating and starting containers:
 ./ccd compose up -d
 ```
 
-Add users and roles:
+Add services, roles and users (needs to be run in below order).
+
+Make sure the `IDAM_ADMIN_USER` `IDAM_ADMIN_PASSWORD` env variables are set to IDAM initial user.
```
geek-soft:
```suggestion
Make sure the `IDAM_ADMIN_USER` and `IDAM_ADMIN_PASSWORD` env variables are set to IDAM initial user.
```

bin/configurer/utils/idam-create-service.sh
```diff
@@ -0,0 +1,37 @@
...
+SELF_REG="${2:-false}"
+CLIENT_ID="${3:-ccd_gateway}"
+CLIENT_SECRET="${4:-ccd_gateway_secret}"
+REDIRECT_URLS="${5:-[\"http://localhost:3451/oauth2redirect\"]}"
+
+apiToken=$(${dir}/idam-authenticate.sh "${IDAM_ADMIN_USER}" "${IDAM_ADMIN_PASSWORD}")
+
+echo -e "\nCreating service with:\nLabel: ${LABEL}\nClient ID: ${CLIENT_ID}\nClient Secret: ${CLIENT_SECRET}\nRedirect URL: ${REDIRECT_URLS}\n"
+
+STATUS=$(curl -s -o /dev/null -w '%{http_code}' -X POST -H 'Content-Type: application/json' -H "Authorization: AdminApiAuthToken ${apiToken}" \
```
geek-soft: To be honest I prefer longer versions of switches such as `-s`, `-o` or `-w`. It make it more readable IMO. Short versions mean sometimes different things in different tools :(

bin/configurer/add-services.sh
akomisarek: Hardocded false.

Redirect URL single value, support only for single redirect,  I agree it is easier.

bin/configurer/utils/idam-create-service.sh
akomisarek: Done.

README.md
akomisarek: Done.

bin/configurer/utils/idam-create-service.sh
akomisarek: Done.

bin/configurer/utils/idam-lease-user-token.sh
akomisarek: I have extracted all parameters for consistency

PR #433, opened 2019-08-29
Bump version of AppInsights lib to take advantage of improvements for Java 11+

geek-soft:

None

### Change description ###

Uses version of library that supports Java 9+ better (including new HTTP client added in Java 11).

Change introduces new version of the client only in Docker environments. There is generally more hassle in doing same for ASE environments and since we are moving to AKS anyway I thought there is no point.

```
[ ] Yes
[x] No
```


geek-soft:
https://github.com/hmcts/cnp-plum-recipes-service/pull/54

Dockerfile
```diff
@@ -1,10 +1,21 @@
...
+
+# Build image
+
+FROM busybox as downloader
+
+RUN wget -P /tmp https://github.com/microsoft/ApplicationInsights-Java/releases/download/2.5.0-BETA.3/applicationinsights-agent-2.5.0-BETA.3.jar
+
+# Application image
+
+FROM hmctspublic.azurecr.io/base/java:openjdk-11-distroless-debug-1.0
```
akomisarek: I guess we need debug version of image for the AI?

Dockerfile
```diff
@@ -1,10 +1,21 @@
-FROM hmctspublic.azurecr.io/base/java:openjdk-11-distroless-1.0
+ARG APP_INSIGHTS_AGENT_VERSION=2.5.0-BETA.3
```
akomisarek: Is it correct to rely on the beta version for the prod stuff?

Dockerfile
```diff
@@ -1,10 +1,21 @@
...
+
+RUN wget -P /tmp https://github.com/microsoft/ApplicationInsights-Java/releases/download/2.5.0-BETA.3/applicationinsights-agent-2.5.0-BETA.3.jar
+
+# Application image
+
+FROM hmctspublic.azurecr.io/base/java:openjdk-11-distroless-debug-1.0
 
 EXPOSE 4000
 
 HEALTHCHECK --interval=10s --timeout=10s --retries=10 CMD http_proxy="" curl --silent --fail http://localhost:4000/health
```
akomisarek: Do we want to remove health check at the same time? Something which was requested as part of the AKS.

geek-soft: Separate PR.

Dockerfile
geek-soft: AppInsights evolves at slow peace and we needed to start to get all features we needed.

At current we already relay on `applicationinsights-agent-2.4.0-BETA-SNAPSHOT.jar`, this PR just bumps to another BETA.

Generally we faced no issue with this approach in the past, obviously I am not saying we won't in the future.

Dockerfile
geek-soft: My mistake, it should not be there, my mistake.

I put it there for troubleshooting times.

Will remove.

geek-soft: Done

PR #437, opened 2019-08-30
Reject submission attempt when case belongs to smoke test LA

geek-soft:

None

### Change description ###

This is to prevent accidental submission of an open case in prod while being logged in as smoke test user.

It gives:

image

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -76,9 +78,20 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
 
         return AboutToStartOrSubmitCallbackResponse.builder()
             .data(data)
+            .errors(validate(mapper.convertValue(data, CaseData.class)))
             .build();
     }
 
+    private List<String> validate(CaseData caseData) {
+        ImmutableList.Builder<String> builder = ImmutableList.builder();
```
swalker125:
```suggestion
        ImmutableList.Builder<String> errors = ImmutableList.builder();
```

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -76,9 +78,20 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
...
+
+        if ("FPLA".equals(caseData.getCaseLocalAuthority())) {
```
swalker125: I would prefer this if statement the other way around, I think that makes more logical sense.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -76,9 +78,20 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
 
         return AboutToStartOrSubmitCallbackResponse.builder()
             .data(data)
+            .errors(validate(mapper.convertValue(data, CaseData.class)))
             .build();
     }
 
+    private List<String> validate(CaseData caseData) {
```
swalker125: validateLocalAuthority?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
geek-soft: This is trick to avoid null checks. Obviously I am aware that `caseData.getCaseLocalAuthority` should be populated at the time.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
geek-soft: I made it generic in case more validation is needed in the future. I am not sure it is worth wait for another build.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
geek-soft: I am not sure it is worth wait for another build especially that context is very narrow here.

PR #446, opened 2019-09-02
Read secrets from vault volumes in AKS

geek-soft:

None

### Change description ###

Later...

```
[ ] Yes
[x] No
```


Jenkinsfile_parameterized
```diff
@@ -9,8 +9,7 @@ properties([
 ])
 
 List<Map<String, Object>> secrets = [
-  secret('address-lookup-api-key', 'ADDRESS_LOOKUP_TOKEN'),
-  secret('notify-api-key', 'NOTIFY_API_KEY')
+  secret('address-lookup-api-key', 'ADDRESS_LOOKUP_TOKEN')
```
adusumillipraveen: This is deprecated way of defining secrets, but this is used only in sandbox jenkins.

charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -7,13 +7,17 @@ java:
...
-    FPL_LOCAL_AUTHORITY_CODE_TO_NAME_MAPPING: ${FPL_LOCAL_AUTHORITY_CODE_TO_NAME_MAPPING}
-    FPL_LOCAL_AUTHORITY_USER_MAPPING: ${FPL_LOCAL_AUTHORITY_USER_MAPPING}
-    FPL_LOCAL_AUTHORITY_CODE_TO_HMCTS_COURT_MAPPING: ${FPL_LOCAL_AUTHORITY_CODE_TO_HMCTS_COURT_MAPPING}
-    FPL_LOCAL_AUTHORITY_CODE_TO_CAFCASS_MAPPING: ${FPL_LOCAL_AUTHORITY_CODE_TO_CAFCASS_MAPPING}
+  keyVaults:
+    fpl:
+      resourceGroup: fpl-case-service
+      secrets:
+        - notify-api-key
+        - fpl-case-service-s2s-secret
```
adusumillipraveen: Unwanted whitespace

Jenkinsfile_parameterized
geek-soft: Agree, whole file is trash as we do not use sandbox env. I might remove it in next PR.

PR #454, opened 2019-09-05
Disable retry in E2E tests

geek-soft:

None

### Change description ###

Retry has some negative effects which is even more noticeable in new test approach. This PR disables retry to fail fast and focus on flaky elements in the future.

Some of unwanted side effects of retries are:
 - screenshot are published despite the fact that test retry might make test pass introducing unnecessary confusion
 - execution time is increased and real problems might be hidden - often wait for element is better long term solution
- if retry happens in test exercising draft event with collections amendments are made in places not initially intended (e.g, details of first respondent are entered as second respondent while details of second respondent are entered as third one)

```
[ ] Yes
[x] No
```


geek-soft:
Hold done with this. I expected more passes while there were only failures.

PR #466, opened 2019-09-12
cleanup ase's after aks migration

naikajah:
Once the ASE to AKS migration is completed. This PR should be marked ready for review and then merged updating the value of
enable_ase to be false for the environment that needs ASE's to be deleted .
e.g.

aat.tfvars
enable_ase = false


geek-soft:
PR needs manual merge resolution. I will look into that next week.

PR #472, opened 2019-09-16
Use local domains for test LA users

geek-soft:

None

### Change description ###

Changes domain names used by LA users to local domains.

```
[ ] Yes
[x] No
```


geek-soft:
Closing for now as I faced difficulties and it is less of a priority right now.

PR #473, opened 2019-09-17
Generate production version of CCD definitions after build promotion

geek-soft:

None

### Change description ###

...

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -55,7 +55,6 @@ withPipeline(type, product, component) {
   installCharts()
   enableSlackNotifications('#fpla-tech')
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile_CNP
```diff
@@ -77,7 +76,11 @@ withPipeline(type, product, component) {
     }
   }
 
-  before('smoketest:saat-staging') {
+  after('prod:promotion') {
+    archiveCoreCaseDataDefinitions('prod')
```
geek-soft: Pipeline is changed, there no longer smoke test phase in prod part of the pipeline so this tries to generate XLS straight after build promotion to prod.

PR #481, opened 2019-09-19
Disable CCD activity auditing in preview env

geek-soft:

None

### Change description ###

As per title

```
[ ] Yes
[x] No
```


charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -34,6 +34,7 @@ ccd:
       NODE_TLS_REJECT_UNAUTHORIZED: 0
       DM_URL: https://gateway-${SERVICE_FQDN}/documents
       DM_URL_REMOTE: https?://(gateway-${SERVICE_FQDN}:443|dm-store-aat.service.core-compute-aat.internal)/documents
+      CCD_ACTIVITY_URL: null
```
geek-soft: It utilises https://github.com/hmcts/ccd-case-ui-toolkit/blob/master/src/shared/services/activity/activity.service.ts#L52

PR #500, opened 2019-09-27
FPLA:755 fix access segregation

toireasam:

### Change description ###

```
[ ] Yes
[ ] No
```


bin/configurer/utils/get-user-ids.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -eu
+
+userids=$(docker run -e PGPASSWORD='openidm' --rm --network ccd-network postgres:11-alpine psql --host shared-db  --username openidm --tuples-only  --command "SELECT string_agg(fullobject->>'_id',',') from managedObjects WHERE fullobject->>'sn' = '(local-authority)';" openidm)
+
+useridsformatted=$(echo $userids | tr -d ' ')
+
+sed -i '' "s/31,32,33/$useridsformatted/" "../../../service/src/main/resources/application-user-mappings.yaml"
+sed -i '' "s/31,32,33/$useridsformatted/" "../../../build/resources/main/application-user-mappings.yaml"
```
geek-soft: This is not needed. We should only modify source files. Build files will be copied over at build time.

service/src/main/resources/application-user-mappings.yaml
```diff
@@ -0,0 +1,6 @@
+spring:
+  profiles: local,user-mappings
```
geek-soft: I see no value in having `user-mappings` profile, a `local` profile is only thing we need.

service/src/main/resources/application-user-mappings.yaml
```diff
@@ -0,0 +1,6 @@
+spring:
```
geek-soft: The problem I see with this file is that after we start tracking it in repository anyone who generated their own local mapping will see that file as modified and some day someone will push it.

I suggest that:
1. script `get-user-ids.sh` (renamed to `generate-local-user-mappings.sh`) is reimplemented to generate `application-user-mappings.yaml` from scratch
2. file `application-user-mappings.yaml` is added to `.git-ignore` so that no-one will be able to commit and push it

toireasam: cool sounds good & makes sense :)

service/src/main/resources/application-user-mappings.yaml
toireasam: https://stackoverflow.com/questions/35902616/starting-spring-application-by-merging-yml-files i think we must set two different profiles for this as shown here ^ let me know what you think

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -0,0 +1,19 @@
...
+useridsformatted=$(echo $userids | tr -d ' ')
+
+cd ../../../service/src/main/resources
+cat > application-user-mappings.yaml <<EOL
+spring:
+  profiles: ${SPRING_PROFILES}
+
+fpl:
+  local_authority_user:
+    mapping: 'FPLA=>0;SA=>${useridsformatted};HN=>34,35'
```
geek-soft: This is not good because all LA users are mapped as Swansea users which means that users within Hillingdon LA are miss-assigned and their access segregation will not work as expected.

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+set -eu
+
+SPRING_PROFILES=user-mappings
```
geek-soft: That could be inlined as for me there is no sufficient reason to have it here. It won't be changed in the future and significance of this in context of file generation is minor IMO.

toireasam: yes makes sense :)

bin/configurer/utils/generate-local-user-mappings.sh
toireasam: fixed :)

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+set -eu
+
+swanseauserids=$(docker run -e PGPASSWORD='openidm' --rm --network ccd-network postgres:11-alpine psql --host shared-db  --username openidm --tuples-only  --command "SELECT string_agg(fullobject->>'_id',',') FROM managedObjects WHERE fullobject->>'userName' LIKE '%swansea%';" openidm)
```
geek-soft: Can I ask you to use either snake case or camel case for the names? It is hard to read at the moment.

service/src/main/resources/application-user-mappings.yaml
toireasam: closing as you are happy with approach discussed

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -0,0 +1,19 @@
...
+hillingdonuserids=$(docker run -e PGPASSWORD='openidm' --rm --network ccd-network postgres:11-alpine psql --host shared-db  --username openidm --tuples-only  --command "SELECT string_agg(fullobject->>'_id',',') FROM managedObjects WHERE fullobject->>'userName' LIKE '%hillingdon%';" openidm)
+
+swanseauseridsformatted=$(echo $swanseauserids | tr -d ' ')
+hillingdonuseridsformatted=$(echo $hillingdonuserids | tr -d ' ')
+
+cd ../../../service/src/main/resources
```
geek-soft: This is a bit fragile. It won't work when script is executed from location other than scripts dir. As a counter measure all other scripts have:

```
dir=$(dirname ${0})
```

and then use script location to calculate other paths:

```
${dir}/...
```

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+set -eu
+
+swanseauserids=$(docker run -e PGPASSWORD='openidm' --rm --network ccd-network postgres:11-alpine psql --host shared-db  --username openidm --tuples-only  --command "SELECT string_agg(fullobject->>'_id',',') FROM managedObjects WHERE fullobject->>'userName' LIKE '%swansea%';" openidm)
+hillingdonuserids=$(docker run -e PGPASSWORD='openidm' --rm --network ccd-network postgres:11-alpine psql --host shared-db  --username openidm --tuples-only  --command "SELECT string_agg(fullobject->>'_id',',') FROM managedObjects WHERE fullobject->>'userName' LIKE '%hillingdon%';" openidm)
+
+swanseauseridsformatted=$(echo $swanseauserids | tr -d ' ')
```
geek-soft:
```suggestion
swanseauseridsformatted=$(echo ${swanseauserids} | tr -d ' ')
```

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -0,0 +1,19 @@
...
+hillingdonuseridsformatted=$(echo $hillingdonuserids | tr -d ' ')
```
geek-soft:
```suggestion
hillingdonuseridsformatted=$(echo ${hillingdonuserids} | tr -d ' ')
```

bin/configurer/utils/generate-local-user-mappings.sh
geek-soft: To take it to next level you could use local function with one argument which is LA name. Function would return same outcome as it does now and function would be invoked 2 times, one time with `swansea` argument and second time with `hillingdon` argument.

It would allow us to cut on duplicates.

NICE TO HAVE.

toireasam: yes of course :)

bin/configurer/utils/generate-local-user-mappings.sh
toireasam: done

bin/configurer/utils/generate-local-user-mappings.sh
toireasam: done

geek-soft:
I fall in love with that little script ❤️

PR #508, opened 2019-10-01
Upgrade Gradle with plugins + service dependencies

geek-soft:

None

### Change description ###

As per title...

```
[ ] Yes
[x] No
```


config/checkstyle/checkstyle.xml
```diff
@@ -234,7 +234,6 @@
       <property name="allowMissingParamTags" value="true"/>
       <property name="allowMissingThrowsTags" value="true"/>
       <property name="allowMissingReturnTag" value="true"/>
-      <property name="minLineCount" value="2"/>
```
geek-soft: Deprecated property got removed in latest version.

service/build.gradle
```diff
@@ -177,7 +177,7 @@ repositories {
 // it is important to specify logback classic and core packages explicitly as libraries like spring boot
 // enforces it's own (older) version which is not recommended.
 def versions = [
-  reformLogging: '5.0.1',
+  reformLogging: '5.1.0',
```
akomisarek: I think thanks for that change, we can now remove the netty.io exclusion which I have added in the owasp config.

PR #513, opened 2019-10-04
fix: new upload sdo docs in gatekeeping event

swalker125:
new event for uploading sdo in gatekeeping state for court admin users


swalker125:
@geek-soft / @akomisarek can I get a force merge on this please? I can't seem to get onto the aks dashboard to work out what is going on. I get an error on max retries for port 8888. This PR is just definition changes, they were manually tested on demo and AAT before manual deployment to PROD

PR #539, opened 2019-10-17
Point preview environments to team own IDAM instance in AKS

geek-soft:

None

### Change description ###

This switches FPL preview environment to team IDAM instance that is deployed in AKS cluster. 

Team IDAM is deployed in preview and configured with all required services, roles and users.

In case switch is required the following steps are required:

1. merge this PR
2. run `bin/configurer/utils/generate-preview-user-mappings.sh` script and update `local-authority-user-mapping` Key Vault key with the result of the script to update user mappings
3. ~~set `fpl_case_service_secret` as value of `fpl-case-service-idam-client-secret` in Key Vault~~ (for now the code is using fpl-case-service-idam-client-secret-preview) seems a bit more elegant) 

```
[ ] Yes
[x] No
```


charts-other/sidam/charts/sidam-database/templates/service.yaml
```diff
@@ -0,0 +1,33 @@
...
+    app.kubernetes.io/name: {{ .Release.Name }}-{{ .Chart.Name }}
+    app.kubernetes.io/instance: {{ .Release.Name }}-{{ .Chart.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+spec:
+  selector:
+    app.kubernetes.io/name: {{ .Release.Name }}-{{ .Chart.Name }}
+  ports:
+    - port: 5432
+      targetPort: 5432
```
akomisarek: Do we need to duplicate this? By default (and for convenience), the target port is the same as port.

charts-other/sidam/charts/sidam-fr-am/templates/service.yaml
```diff
@@ -0,0 +1,41 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name }}-{{ .Chart.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+spec:
+  selector:
+    app.kubernetes.io/name: {{ .Release.Name }}-{{ .Chart.Name }}
+  ports:
+    - name: ldap
+      port: 1389
+      targetPort: 1389
```
akomisarek: Fine if you want to have this to be explicit :)

charts-other/sidam/charts/sidam-fr-am/templates/service.yaml
```diff
@@ -0,0 +1,41 @@
...
+      port: 8080
+      targetPort: 8080
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: fr-am
+  labels:
```
akomisarek: It's me not knowing Helm and chart (catching up!), but why we need to repeat it this? It's like the same as above but without name having placeholders...

charts-other/sidam/charts/sidam-fr-idm/templates/service.yaml
```diff
@@ -0,0 +1,41 @@
...
+      targetPort: 18080
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: fr-idm
+  labels:
+    app.kubernetes.io/name: {{ .Release.Name }}-{{ .Chart.Name }}
```
akomisarek: Hmm, don't understand why this service duplication is required? 🤔

charts-other/sidam/charts/sidam-fr-idm/templates/service.yaml
```diff
@@ -0,0 +1,41 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name }}-{{ .Chart.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+spec:
+  selector:
+    app.kubernetes.io/name: {{ .Release.Name }}-{{ .Chart.Name }}
+  ports:
+    - name: jxm
```
akomisarek: Probalby doesn't really matter, but it should be JMX :)

charts-other/sidam/charts/sidam-web/values.yaml
```diff
@@ -0,0 +1,21 @@
+java:
```
akomisarek: Do we need sidam-web? Not sure just trying to minimise our preview envs being overloaded :(512 mem and 250 cpu per image sounds a lot :()

charts-other/sidam/charts/sidam-fr-idm/templates/service.yaml
geek-soft: Yeah, removed in most recent version.

charts-other/sidam/charts/sidam-fr-idm/templates/service.yaml
geek-soft: This is funny bit. ForgeRock images are not parametrised and they have some host names hardcoded so nothing other that fr-idm won't work. In final version I will keep just one version. I started my own naming convention but later I learnt that it cannot be used because of these hardcoded host names.

charts-other/sidam/charts/sidam-api/values.yaml
```diff
@@ -0,0 +1,40 @@
+java:
+  releaseNameOverride: '{{ .Release.Name }}-sidam-api'
+  image: hmctspublic.azurecr.io/idam/api:preview
```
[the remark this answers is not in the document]
geek-soft: You must have seen outdated version. It uses stable tag now.

charts-other/sidam/charts/sidam-database/templates/service.yaml
geek-soft: I would need to double check but with services I think you need to provide source and target ports.

geek-soft: Nope, I might be wrong. If there is no target port then port is used. Will remove.

charts-other/sidam/charts/sidam-fr-am/templates/service.yaml
geek-soft: Duplicate will be removed as stated in other comment.

charts-other/sidam/charts/sidam-web/values.yaml
geek-soft: Yes, this is where user enters username and password and internally it points to API instance. We would not be able to reuse that frontend from AAT.

charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -7,9 +7,10 @@ java:
   ingressHost: ${SERVICE_FQDN}
   livenessDelay: 50
   environment:
+    IDAM_API_URL: http://api-idam-fpl.service.core-compute-preview.internal
     CORE_CASE_DATA_API_URL: http://${SERVICE_NAME}-data-store-api
     IDAM_CLIENT_ID: fpl_case_service
-    IDAM_CLIENT_REDIRECT_URI: https://fpl-case-service-*.service.core-compute-aat.internal/oauth2/callback
+    IDAM_CLIENT_REDIRECT_URI: https://localhost:9000/oauth2/callback
```
akomisarek: Hmm why this is localhost?
Is it overriden? maybe we shouldn't put the value here then? I find it confusing.

bin/configurer/add-services.sh
```diff
@@ -4,7 +4,7 @@ set -eu
 
 dir=$(dirname ${0})
 
-${dir}/utils/idam-create-service.sh "ccd_gateway" "ccd_gateway" "ccd_gateway_secret" "http://localhost:3451/oauth2redirect"
+${dir}/utils/idam-create-service.sh "CCD" "ccd_gateway" "ccd_gateway_secret" "http://localhost:3451/oauth2redirect"
```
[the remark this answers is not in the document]
geek-soft: This is just label change to use same one locally as in AAT. To be honest I can revert that change if you want as I took different approach and this script is not used in preview.

bin/configurer/utils/generate-preview-user-mappings.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -eu
+
+function get_user_ids() {
+  kubectl exec service/shared-db -- psql --username openidm --tuples-only --command "SELECT string_agg(fullobject->>'_id',',') FROM managedObjects WHERE fullobject->>'userName' LIKE '%$1%';" openidm
```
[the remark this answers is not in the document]
geek-soft: It has lots in common. Process of extracting user IDs is different. Another difference is that this script does not generate mapping YAML file, it just prints something that can be easily put in key vault.

Finally hopefully this is just workaround until we have organisation access.

For all above reasons I did spend much time to find better place.

charts/fpl-case-service/values.preview.template.yaml
geek-soft: That redirect URL doesn't matter as redirect never happens. It is just to match whitelisted URL on IDAM side. It matches https://github.com/hmcts/cnp-flux-config/blob/master/k8s/preview/family-public-law/idam.yaml#L36

bin/configurer/add-services.sh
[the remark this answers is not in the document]
akomisarek: Ah that's ok then, sorry should check the parameters.

charts/fpl-case-service/values.preview.template.yaml
akomisarek: Yeah this bit is something I don't understand in those authentication flows, why the parameters which doesn't matter are required to be passed, yeah the redirect neve happens as this is not browser login, should there be another endpoint? 🤔 need to read more about it to understand it properly!

geek-soft:
Parked for now as AAT IDAM works OK.

PR #540, opened 2019-10-18
FPLA 836: Document Upload Adjustment

lewisbirks:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/FPLA-836

### Change description ###
Merged the uploadDocuments events into one that can now occur throughout the case lifecycle and ensure that users still have the correct permissions to view and add what they should.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #542, opened 2019-10-21
Fpla-604: upload c2 notification

Gogetter:
### Notify HMCTS of C2 upload ###

```
[ ] Yes
[x] No
```


ccd-definition/CaseEvent.json
```diff
@@ -349,10 +349,11 @@
     "PreConditionState(s)": "Submitted",
     "PostConditionState": "Submitted",
     "SecurityClassification": "Public",
+    "CallBackURLAboutToSubmitEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/upload-c2/about-to-submit",
```
swalker125: just had a chat with you re: using one event for c2. Check out https://github.com/hmcts/fpl-ccd-configuration/pull/540/files as I believe it's a similar sort of thing implementation wise

@akomisarek, separate story to combine these? sounds like c2 can and will be uploaded anywhere from submitted onwards

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/UserRole.java
```diff
@@ -0,0 +1,51 @@
...
+
+    GATEKEEPER {
+        public List<String> getRoles() {
+            return addRoles(DEFAULT.getRoles(), "caseworker-publiclaw-gatekeeper");
+        }
+    },
+
+    JUDICIARY {
+        public List<String> getRoles() {
+            return addRoles(DEFAULT.getRoles(), "caseworker-publiclaw-gatekeeper");
```
swalker125:
```suggestion
            return addRoles(DEFAULT.getRoles(), "caseworker-publiclaw-judiciary");
```

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +78,66 @@
...
+    @MockBean
+    private UserDetailsService userDetailsService;
+
     @InjectMocks
     private NotificationHandler notificationHandler;
 
+    private final DateFormatterService dateFormatterService = new DateFormatterService();
+
+    @Test
+    void shouldNotNotifyHmctsPAdminOnC2Upload() throws IOException, NotificationClientException {
```
swalker125:
```suggestion
    void shouldNotNotifyHmctsAdminOnC2Upload() throws IOException, NotificationClientException {
```

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserDetailsService.java
```diff
@@ -23,4 +24,8 @@ public UserDetailsService(IdamApi idamApi) {
     public String getUserName(String authorization) {
         return idamApi.retrieveUserDetails(authorization).getFullName();
     }
+
+    public UserDetails getUserDetails(final String authorization) {
```
swalker125: seems silly to have a getUserName method and a getUserDetials method when they are doing the same thing... how much effort is it to refactor the use of the method above to return UserDetails?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/HmctsEmailContentProvider.java
```diff
@@ -30,4 +46,35 @@ public HmctsEmailContentProvider(LocalAuthorityNameLookupConfiguration localAuth
...
+            caseDetails.getData().get("respondents1"), new TypeReference<>() {});
+
+        List<Respondent> respondents = respondents1.stream()
+            .map(respondent ->
+                mapper.getObjectMapper().convertValue(respondent.get("value"), Respondent.class))
+            .collect(toList());
+
+        return Map.of(
+            "court", hmctsCourtLookupConfiguration.getCourt(localAuthorityCode).getName(),
+            "lastNameOfRespondent", respondents
```
swalker125: this will only ever get the lastName of the first entry in respondents1 right? Is this correct?

Gogetter: currently yes, as per the jira.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: thanks for spotting this @swalker125

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserDetailsService.java
Gogetter: you are right. Making the change here will affect 9-classes. Not sure if we want to make as part of this PR?

swalker125: you want to create a tech debt ticket in jira for this?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/HmctsEmailContentProvider.java
swalker125: discussed, need to investigate with BA

service/src/main/java/uk/gov/hmcts/reform/fpl/service/UserDetailsService.java
Gogetter: Done [here](https://tools.hmcts.net/jira/browse/FPLA-850)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/HmctsEmailContentProvider.java
Gogetter: Mia, has confirmed we will need the first respondent.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/HmctsEmailContentProvider.java
```diff
@@ -30,4 +54,56 @@ public HmctsEmailContentProvider(LocalAuthorityNameLookupConfiguration localAuth
...
+        return super.getCasePersonalisationBuilder(caseDetails)
+            .put("court", hmctsCourtLookupConfiguration.getCourt(localAuthorityCode).getName())
+            .put("lastNameOfRespondent", getRespondent1Lastname(respondents))
+            .put("familyManCaseNumber", StringUtils.defaultIfBlank(caseData.getFamilyManCaseNumber(), ""))
+            .put("hearingDate", dateFormatterService.formatLocalDateToString(
+                getHearingBookingDate(caseData), FormatStyle.MEDIUM))
+            .build();
+    }
+
+    private LocalDate getHearingBookingDate(final CaseData caseData) {
```
toireasam: are you able to use the get most urgent hearing booking method in hearing booking service class?

Gogetter: thanks for the spot @toireasam, this field has now been taken out.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsController.java
```diff
@@ -49,13 +54,23 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(
...
         return AboutToStartOrSubmitCallbackResponse.builder().data(data).build();
     }
 
+    @PostMapping("/submitted")
+    public void handleSubmittedEvent(
+        @RequestHeader(value = "authorization") String authorization,
+        @RequestHeader(value = "user-id") String userId,
+        @RequestBody CallbackRequest callbackRequest) {
+
+        applicationEventPublisher.publishEvent(new C2UploadNotifyEvent(callbackRequest, authorization, userId,
```
akomisarek: I am not fan of the name `C2UploadNotifyEvent`. This should know nothing about any notification being send. The event should simply state what happened, i.e. C2Uploaded.

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/UserRole.java
```diff
@@ -0,0 +1,51 @@
...
+        public List<String> getRoles() {
+            return Arrays.asList("caseworker", "caseworker-publiclaw");
+        }
+    },
+
+    LOCAL_AUTHORITY {
+        public List<String> getRoles() {
+            return addRoles(DEFAULT.getRoles(), "caseworker-publiclaw-solicitor");
+        }
+    },
```
akomisarek: Why not simply using constructor?
```suggestion
         LOCAL_AUTHORITY("caseworker-publiclaw-solicitor")        
```
And build the list in the constructor?

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/UserRole.java
```diff
@@ -0,0 +1,51 @@
+package uk.gov.hmcts.reform.fpl.enums;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public enum UserRole {
+    DEFAULT {
```
akomisarek: This role shouldn't be accessible in the code.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -62,6 +68,23 @@ public void sendNotificationToHmctsAdmin(SubmittedCaseEvent event) {
         sendNotification(HMCTS_COURT_SUBMISSION_TEMPLATE, email, parameters, reference);
     }
 
+    @EventListener
+    public void sendNotificationForC2Upload(final C2UploadNotifyEvent caseEvent) {
+        UserDetails userDetails = caseEvent.getUserDetails();
+        List<String> roles = userDetails.getRoles();
+
+        if (!roles.containsAll(UserRole.HMCTS_ADMIN.getRoles())) {
```
akomisarek: I would probably just check contain HMCTS_ADMIN. No reason to check others.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/HmctsEmailContentProvider.java
```diff
@@ -30,4 +44,22 @@ public HmctsEmailContentProvider(LocalAuthorityNameLookupConfiguration localAuth
             .put("localAuthority", localAuthorityNameLookupConfiguration.getLocalAuthorityName(localAuthorityCode))
             .build();
     }
+
+    public Map<String, Object> buildC2UploadNotification(final CaseDetails caseDetails) {
+        CaseData caseData = objectMapper.convertValue(caseDetails.getData(), CaseData.class);
+
+        List<Map<String, Object>> respondents1 = (ObjectUtils.isEmpty(caseDetails.getData().get("respondents1"))
```
akomisarek: Not sure why caseData was created above if it's not used (or rather used to get the family man?)

Couldn't the code to get respondents to be refactored to use caseData? I find it confusing and difficult to read, at least extract it to another method please.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,38 @@
...
+    private static String getRespondent1Lastname(final List<Respondent> respondents) {
+        Optional<Respondent> optionalRespondent =
+            (CollectionUtils.isEmpty(respondents) ? Optional.empty() : respondents
+                .stream()
+                .filter(Objects::nonNull)
+                .findFirst());
+
+        if (optionalRespondent.isPresent()) {
+            return StringUtils.defaultIfBlank(optionalRespondent.get().getParty().getLastName(), "");
+        }
```
akomisarek: optionalRospondent.map with filters and another map? not sure if it's more readable, just suggestion :) 

You could return orElse "" as well.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,38 @@
...
+
+        return StringUtils.EMPTY;
```
akomisarek: Why not simply ""? No value in using constant here in my opinion just clutters the code. :( It's not like "" is going to be modified..

If you really want to return constant you should define value which is default when respondent not provided and use it. Otherwise it just doesn't make sense.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +75,63 @@
...
+            .willReturn(new UserDetails("1", "hmcts-admin@test.com",
+                "Hmcts", "Test", UserRole.HMCTS_ADMIN.getRoles()));
+
+        given(hmctsEmailContentProvider.buildC2UploadNotification(callbackRequest().getCaseDetails()))
+            .willReturn(parameters);
+
+        notificationHandler.sendNotificationForC2Upload(new C2UploadNotifyEvent(callbackRequest(), AUTH_TOKEN, USER_ID,
+            userDetailsService.getUserDetails(AUTH_TOKEN)));
+
+        verify(notificationClient, times(0)).sendEmail(
```
akomisarek: I don't like this verification:

* If you remove your if statement, your code fails with NPE.
* If you fix npe by adding mocking of the court, this tests PASSES, as the values are not null.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +75,63 @@
...
+            .willReturn("Example Local Authority");
+
+        given(hmctsEmailContentProvider.buildC2UploadNotification(callbackRequest().getCaseDetails()))
+            .willReturn(parameters);
+
+        notificationHandler.sendNotificationForC2Upload(new C2UploadNotifyEvent(callbackRequest(), AUTH_TOKEN, USER_ID,
+            userDetailsService.getUserDetails(AUTH_TOKEN)));
+
+        verify(notificationClient, times(1)).sendEmail(
+            eq(C2_UPLOAD_SUBMISSION_TEMPLATE), notNull(), eq(parameters), eq("12345"));
```
akomisarek: Couldn't we check what's the email in the second parameter? You mocked it couple lines above.

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelperTest.java
```diff
@@ -0,0 +1,21 @@
...
+import uk.gov.hmcts.reform.fpl.model.CaseData;
+
+import java.util.ArrayList;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class EmailNotificationHelperTest {
+
+    @Test
+    void subjectLineShouldBeEmpty() {
```
akomisarek: Not sure what's this test is doing:

* we have got two separates method in the class
* The method has got some logic with ? : or if statemetns and we are just checking if the subjectLine is not empty.

I don't think this test is good enough.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: good spot

Gogetter: updated

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelperTest.java
Gogetter: updated

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: updated

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/HmctsEmailContentProvider.java
Gogetter: updated

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/UserRole.java
Gogetter: now updated.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
Gogetter: yes we could, but the roles where returned as List, why we moved it into a UserRole enum

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsController.java
Gogetter: updated.... naming can be hard 😄

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: you are right on this. Was just preference. Now updated to use orElse

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: Hmmm... taking a look at this again. thanks for spotting

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private static final String COMMA = ",";
+
+    private EmailNotificationHelper() {
+    }
+
+    public static String buildSubjectLine(final CaseDetails caseDetails) {
+        final List<Respondent> respondents = getRespondents1(caseDetails);
+
+        final String lastName = (StringUtils.isNotBlank(getRespondent1Lastname(respondents))
```
swalker125: I would prefer static method imports, just isNotBlank etc

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsController.java
```diff
@@ -49,13 +54,23 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(
...
     }
 
+    @PostMapping("/submitted")
+    public void handleSubmittedEvent(
+        @RequestHeader(value = "authorization") String authorization,
+        @RequestHeader(value = "user-id") String userId,
+        @RequestBody CallbackRequest callbackRequest) {
+
+        applicationEventPublisher.publishEvent(new C2UploadedEvent(callbackRequest, authorization, userId,
+            userDetailsService.getUserDetails(authorization)));
```
swalker125: let's just get userDetails in the eventHandler rather than passing it through? We need to set it in the handler anyway so 🤷‍♂

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -25,6 +28,8 @@
 @Data
 @Builder(toBuilder = true)
 @AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
swalker125: why?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: yeah, it could be that. Left this way as it gives us visibility as to where the particular method is coming from. do we have a convention for this?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import static java.util.stream.Collectors.toList;
+
+public class EmailNotificationHelper {
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    private static final String COMMA = ",";
```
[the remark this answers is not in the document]
Gogetter: yep, should be local. Nice spot 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsController.java
Gogetter: yes we could. Will update

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: upodated

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.springframework.util.CollectionUtils.isEmpty;
+
+public class EmailNotificationHelper {
+
+    private EmailNotificationHelper() {
+    }
+
+    public static String buildSubjectLine(final CaseData caseData) {
+        final String lastName = (isNotBlank(getRespondents1Lastname(caseData))
```
swalker125: why can't this just be defaultIfNull(getRespondets1(LastName(caseData), "") ?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,38 @@
...
+            ? String.format("%1$s", getRespondents1Lastname(caseData)) : "");
+
+        final String familyManCaseNumber = (isNotBlank(caseData.getFamilyManCaseNumber())
```
swalker125: same as above

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,38 @@
...
+            ? String.format("%1$s", caseData.getFamilyManCaseNumber()) : "");
+
+        return String.format("%1$s%2$s%3$s", lastName,
+            (isNoneBlank(lastName, familyManCaseNumber) ? ", " : ""),
+            familyManCaseNumber);
+    }
+
+    private static String getRespondents1Lastname(final CaseData caseData) {
+        return isEmpty(caseData.getRespondents1()) ? "" : caseData.getRespondents1()
```
swalker125: looking at this method, you're actually dealing with the null check here, so either you can edit your above methods to set lastName = this method, which then returns the last name or an empty string back. OR you can use defaultIfNull and have this method return null

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,38 @@
...
+            .stream().filter(Objects::nonNull)
+            .map(Element::getValue).filter(Objects::nonNull)
+            .findFirst().map(Respondent::getParty)
+            .map(RespondentParty::getLastName).orElse("");
```
swalker125: can we space these out properly please, each . on a new line I reckon

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +76,81 @@
...
+            .put("hearingDetailsCallout", SUBJ_LINE)
+            .put("reference", "12345")
+            .put("caseUrl", "null/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345")
+            .build();
+
+        given(idamApi.retrieveUserDetails(AUTH_TOKEN))
+            .willReturn(new UserDetails("1", "hmcts-admin@test.com",
+                "Hmcts", "Test", UserRole.HMCTS_ADMIN.getRoles()));
+
+        given(userDetailsService.getUserDetails(AUTH_TOKEN))
```
swalker125: why do you need to mock this method?

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +76,81 @@
...
+    void shouldNotifyNonHmctsAdminOnC2Upload() throws IOException, NotificationClientException {
+        final String SUBJ_LINE = "Lastname, SACCCCCCCC5676576567";
+        final Map<String, Object> parameters = ImmutableMap.<String, Object>builder()
+            .put("subjectLine", SUBJ_LINE)
+            .put("hearingDetailsCallout", SUBJ_LINE)
+            .put("reference", "12345")
+            .put("caseUrl", "null/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345")
+            .build();
+
+        given(userDetailsService.getUserDetails(AUTH_TOKEN))
```
swalker125: why do you need to mock this method?

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +76,81 @@
...
+
+    @Mock
+    private UserDetailsService userDetailsService;
+
     @InjectMocks
     private NotificationHandler notificationHandler;
 
+    @Test
+    void shouldNotNotifyHmctsAdminOnC2Upload() throws IOException, NotificationClientException {
+        final String SUBJ_LINE = "Lastname, SACCCCCCCC5676576567";
```
swalker125: what are your thoughts on using @Nested for these two tests here. Then setting SUBJ line and parameters outside the test

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProvider.java
```diff
@@ -0,0 +1,38 @@
...
+        super(uiBaseUrl);
+        this.objectMapper = objectMapper;
+    }
+
+    public Map<String, Object> buildC2UploadNotification(final CaseDetails caseDetails) {
+        CaseData caseData = objectMapper.convertValue(caseDetails.getData(), CaseData.class);
+        final String subjectLine = EmailNotificationHelper.buildSubjectLine(caseData);
+        return Map.of(
+            "subjectLine", subjectLine,
+            "hearingDetailsCallout", subjectLine,
```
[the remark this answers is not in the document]
Gogetter: yes it is. Had to separate as we tried to decide if (or not) to add _hearingDate_ to _ hearingDetailsCallout_.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: yes it can be, actually makes it shorter 😄

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: to have value to use in _NotificationHandler.sendNotificationForC2Upload(...)_

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: to have value to use in _NotificationHandler.sendNotificationForC2Upload(...)_

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: we could use that actually

ccd-definition/AuthorisationCaseEvent.json
```diff
@@ -345,8 +338,43 @@
   {
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
-    "CaseEventID": "createNoticeOfProceedings",
+    "CaseEventID": "uploadC2",
```
Jamesnelson117: Upload C2 goes live right after this PR?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -62,6 +75,23 @@ public void sendNotificationToHmctsAdmin(SubmittedCaseEvent event) {
         sendNotification(HMCTS_COURT_SUBMISSION_TEMPLATE, email, parameters, reference);
     }
 
+    @EventListener
+    public void sendNotificationForC2Upload(final C2UploadedEvent caseEvent) {
+        List<String> roles = userDetailsService.getUserDetails(caseEvent.getAuthorization()).getRoles();
+
+        if (!roles.containsAll(UserRole.HMCTS_ADMIN.getRoles())) {
+            CaseDetails caseDetailsFromEvent = caseEvent.getCallbackRequest().getCaseDetails();
```
Jamesnelson117: Consider using ObjectMapper to map this to CaseData. That way we do not need to force the cast on line 84.

`String localAuthorityCode = caseData.getCaseLocalAuthority()`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
```diff
@@ -17,14 +17,14 @@
...
+    final String uiBaseUrl;
 
     protected AbstractEmailContentProvider(String uiBaseUrl) {
         this.uiBaseUrl = uiBaseUrl;
     }
 
     @SuppressWarnings("unchecked")
-    protected ImmutableMap.Builder<String, Object> getCasePersonalisationBuilder(CaseDetails caseDetails) {
+    ImmutableMap.Builder<String, Object> getCasePersonalisationBuilder(CaseDetails caseDetails) {
         List<String> ordersAndDirections = buildOrdersAndDirections((Map<String, Object>) caseDetails.getData().get("orders"));
```
Jamesnelson117: Again can we use CaseData class to remove supression?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
```diff
@@ -17,14 +17,14 @@
...
 
         Optional<String> timeFrame = Optional.ofNullable((Map<String, Object>) caseDetails.getData().get("hearing"))
```
Jamesnelson117: Maybe i'm missing something but what do you expect to see on hearing?

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/UserRole.java
```diff
@@ -0,0 +1,45 @@
...
+
+    JUDICIARY {
+        public List<String> getRoles() {
+            return addRoles("caseworker-publiclaw-judiciary");
+        }
+    };
+
+    public abstract List<String> getRoles();
+
+    private static List<String> addRoles(final String roleToAdd) {
```
lewisbirks: Would it not be simpler to just save the role of the enum as a private string called `role` or something and then instead of having `addRoles(...)`, `getRoles()` can just be `return new ArrayList<>(Arrays.asList("caseworker", "caseworker-publiclaw", this.role));`

Just seems a bit simpler

Gogetter: we could but we will have to repeat `"caseworker", "caseworker-publiclaw"` for every `getRoles(...)` which is what was avoiding

ccd-definition/AuthorisationCaseEvent.json
```diff
@@ -6,13 +6,6 @@
     "UserRole": "caseworker-publiclaw-solicitor",
     "CRUD": "CRU"
   },
-  {
```
swalker125: you've removed perms here. Please add back. It's easier for me to add just C and U when we toggle on

ccd-definition/AuthorisationCaseEvent.json
```diff
@@ -345,8 +338,43 @@
   {
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
-    "CaseEventID": "createNoticeOfProceedings",
```
swalker125: you have removed perms here

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelperTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.junit.jupiter.api.Test;
+import uk.gov.hmcts.reform.fpl.model.CaseData;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class EmailNotificationHelperTest {
+
+    @Test
+    void subjectLineShouldBeEmptyWhenNoRespondentOrCaseNumberEmpty() {
```
Jamesnelson117: Do need to a test to ensure that if respondent[0] has no name but respondents[1] has a name, it will return back empty string?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProvider.java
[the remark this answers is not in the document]
swalker125: so why are we using two variables? can we not use subject line twice in the template?

ccd-definition/AuthorisationCaseEvent.json
Jamesnelson117: Which is fine, we do not want to go live with createNoticeOfProceedings yet.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,39 @@
...
+        final String lastName = getRespondents1Lastname(caseData);
+        final String familyManCaseNumber = defaultIfNull(caseData.getFamilyManCaseNumber(), "");
+
+        return String.format("%1$s%2$s%3$s", lastName,
+            (isNoneBlank(lastName, familyManCaseNumber) ? ", " : ""),
+            familyManCaseNumber);
+    }
+
+    private static String getRespondents1Lastname(final CaseData caseData) {
+        return isEmpty(caseData.getRespondents1()) ? "" : caseData.getRespondents1()
```
swalker125: this still isn't quite right. You don't need default if null up above as your getRespondentLast name returns either the name or "". i.e    

```
final String familyManCaseNumber = caseData.getFamilyManCaseNumber();
```

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +77,84 @@
...
+        @Test
+        void shouldNotifyNonHmctsAdminOnC2Upload() throws IOException, NotificationClientException {
+            final Map<String, Object> parameters = ImmutableMap.<String, Object>builder()
+                .put("subjectLine", subjectLine)
+                .put("hearingDetailsCallout", subjectLine)
+                .put("reference", "12345")
+                .put("caseUrl", "null/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345")
+                .build();
+
+            given(userDetailsService.getUserDetails(AUTH_TOKEN))
```
swalker125: I still don't understand why we mock this. All this method does is called idamApi.retrieveUserDetails. Surely one mock is all we need

ccd-definition/AuthorisationCaseEvent.json
swalker125: C and U are easier for me to add back than the full permission

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: we do need the `defaultIfNull` for `familyManCaseNumber` as it will return a `null` if not initialised - and we want an empty string when `null`. 

Also, `isEmpty(caseData.getRespondents1())` checks if the list is empty or `null`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
Gogetter: this is existing code. Do we want to make this change as part of current pr?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelperTest.java
Gogetter: added

ccd-definition/AuthorisationCaseEvent.json
swalker125: @Gogetter please add these back

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -32,23 +41,11 @@
...
-                               CafcassEmailContentProvider cafcassEmailContentProvider,
-                               GatekeeperEmailContentProvider gatekeeperEmailContentProvider) {
-        this.hmctsCourtLookupConfiguration = hmctsCourtLookupConfiguration;
-        this.cafcassLookupConfiguration = cafcassLookupConfiguration;
-        this.notificationClient = notificationClient;
-        this.hmctsEmailContentProvider = hmctsEmailContentProvider;
-        this.cafcassEmailContentProvider = cafcassEmailContentProvider;
-        this.gatekeeperEmailContentProvider = gatekeeperEmailContentProvider;
-    }
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
swalker125: @Slf4j annotation instead of this please. Then your logger is log

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProvider.java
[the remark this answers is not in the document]
swalker125: marked as resolved but not resolved @Gogetter

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +77,65 @@
...
+            .put("reference", "12345")
+            .put("caseUrl", "null/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345")
+            .build();
+
+        @Test
+        void shouldNotNotifyHmctsAdminOnC2Upload() throws IOException, NotificationClientException {
+            given(idamApi.retrieveUserDetails(AUTH_TOKEN))
+                .willReturn(new UserDetails("1", "hmcts-admin@test.com",
+                    "Hmcts", "Test", HMCTS_ADMIN.getRoles()));
+
```
swalker125: remove blank line

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +77,65 @@
...
+            given(c2UploadedEmailContentProvider.buildC2UploadNotification(callbackRequest().getCaseDetails()))
```
swalker125: this mock can be extracted to a before and removed from the tests

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Objects;
+
+import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
+import static org.apache.commons.lang3.StringUtils.isNoneBlank;
+import static org.springframework.util.CollectionUtils.isEmpty;
+
+public class EmailNotificationHelper {
+
+    private EmailNotificationHelper() {
```
swalker125: I seem to remember sonar cloud will complain about an empty constructor

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
swalker125: @Gogetter

Gogetter: @swalker125 yes it can but they use two different emails and different UserRole

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: yes it did but the bug has been fixed [Sonar end](https://jira.sonarsource.com/browse/SONARJAVA-1138)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
```diff
@@ -17,14 +17,14 @@
 @SuppressWarnings({"LineLength", "VariableDeclarationUsageDistance"})
 public abstract class AbstractEmailContentProvider {
 
-    private final String uiBaseUrl;
+    final String uiBaseUrl;
```
lewisbirks: protected instead?

Gogetter: as discussed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProvider.java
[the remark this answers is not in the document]
akomisarek: Ok I think only one variable now so it is probably fine to resolve.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public static String buildSubjectLine(final CaseData caseData) {
+        final String lastName = getRespondents1Lastname(caseData);
+        final String familyManCaseNumber = defaultIfNull(caseData.getFamilyManCaseNumber(), "");
+
+        return String.format("%1$s%2$s%3$s", lastName,
```
akomisarek: Why not simply, Stream.of(lastName, familyManCaseNumber).collect(joining(",")?

akomisarek: Or remove 1$, as we are using them in order they are just cluttering.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -67,9 +78,65 @@
...
+                .willReturn(new UserDetails("1", "hmcts-admin@test.com",
+                    "Hmcts", "Test", HMCTS_ADMIN.getRoles()));
+
+            notificationHandler.sendNotificationForC2Upload(new C2UploadedEvent(callbackRequest(), AUTH_TOKEN, USER_ID));
+
+            SendEmailResponse response = verify(notificationClient, times(0))
+                .sendEmail(eq(C2_UPLOAD_NOTIFICATION_TEMPLATE), eq("hmcts-admin@test.com"),
+                    eq(parameters), eq("12345"));
+
+            assertThat(response).isNull();
```
swalker125: I would prefer an assertion here that notificationClient is called never() I think

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -0,0 +1,41 @@
...
+    public static String buildSubjectLine(final CaseData caseData) {
+        final String lastName = getRespondents1Lastname(caseData);
+        final String familyManCaseNumber = defaultIfNull(caseData.getFamilyManCaseNumber(), "");
+
+        return Stream.of(lastName, familyManCaseNumber)
+            .filter(StringUtils::isNotBlank)
+            .collect(joining(", "));
+    }
+
+    private static String getRespondents1Lastname(final CaseData caseData) {
```
lewisbirks: LastName instead of Lastname

swalker125: Actually whilst we are here, let's drop the 1 in respondents. 

"getFirstRespondentLastName"??

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: updated

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
Gogetter: updated

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProvider.java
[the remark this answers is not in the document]
Gogetter: @swalker125, there is currently an ongoing discussion as to what the callout should be (with or without the hearing date). Left this in as at the moment its without the hearingDate, but this could change and `hearingDetailsCallout` value will be different from `subjectLine`

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
Gogetter: done

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
swalker125: @Gogetter I still see assertThat(response).isNull();

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsControllerTest.java
```diff
@@ -100,12 +113,16 @@ void shouldAppendAnAdditionalC2DocumentBundleWhenAC2DocumentBundleIsPresent() th
         assertC2BundleDocument(existingC2Document, "C2 document one");
         assertC2BundleDocument(appendedC2Document, "C2 document two");
 
-        assertThat(appendedC2Document.getAuthor()).isEqualTo("Emma Taylor");
-        assertThat(appendedC2Document.getUploadedDateTime()).isEqualTo(dateFormatterService
-            .formatLocalDateTimeBaseUsingFormat(ZONE_DATE_TIME.toLocalDateTime(), "h:mma, d MMMM yyyy"));
+        assertThat(appendedC2Document.getAuthor()).isEqualTo(USER_NAME);
+
+        // updated to use LocalDate to avoid 1-minute issue
+        LocalDateTime uploadedDateTime = LocalDateTime.parse(appendedC2Document.getUploadedDateTime(), FORMATTER);
```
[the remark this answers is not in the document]
akomisarek: Yeah that's something I had in mind, we can use different bean for the test purposes and inject itas the spring service, good find :)

PR #550, opened 2019-10-22
clean: using DOCMOSIS_KEY as env variable

akomisarek:
### Change description ###

Update README file and use DOCMOSIS_KEY from env variable

```
[ ] Yes
[X] No
```


README.md
```diff
@@ -89,6 +89,15 @@ To enable retry upon test failure please set `TEST_RETRIES` environment variable
 TEST_RETRIES=2 yarn test
 ```
 
+### Docmosis Tornado
```
geek-soft: I am not sure it is the right place. With `H3` it makes it part of `Testing` section which IMO is not related to running tests. I mean not directly.

geek-soft: On second thought whole readme is bit wired because in getting started it focuses on E2E tests.

akomisarek: Yeah to be honest I was not sure where to put it, would use some cleaning.

akomisarek: Did second level header above testing.,

PR #557, opened 2019-10-25
FPLA-872: Make-Respondent1-Firstname-And-Surname-Mandatory

Jamesnelson117:

- [x] commit messages are meaningful and follow good commit message guidelines

https://tools.hmcts.net/jira/browse/FPLA-872

### Change description ###
Respondent 1 first name and last name is now mandatory

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/enums/Section.java
```diff
@@ -4,15 +4,16 @@
     APPLICANT("applicant", "applicant"),
     CHILDREN("children", "children"),
     ORDERS("orders", "orders and directions needed"),
+    RESPONDENTS("respondent", "respondents"),
     GROUNDS("grounds", "grounds for the application"),
     HEARING("hearing", "hearing needed"),
-    DOCUMENTS("Document", "documents"),
-    CASENAME("caseName", "case name");
+    DOCUMENTS("document", "documents"),
+    CASENAME("casename", "case name");
```
lewisbirks: Is there a reason why this is no longer camel case? Also should document and respondent be singular or plural?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -44,7 +46,18 @@
     @Valid
     private final List<@NotNull(message = "You need to add details to applicant")
         Element<Applicant>> applicants;
-    private final List<Element<Respondent>> respondents1;
+    @NotNull(message = "You need to add details to respondents")
+    private final List<@NotNull(message = "You need to add details to respondents")Element<Respondent>> respondents1;
+
+    @Valid
+    private final Respondent getFirstRespondent() {
+        if (isEmpty(respondents1) || isEmpty(respondents1.get(0))) {
```
lewisbirks: Is there a chance that the value of the element could be null? If so then a check for that here might be useful instead of returning a null value.

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/Section.java
Jamesnelson117: I'm using JSR for validation and it uses the name of the class as a validation key i.e package `uk.gov.hmcts.reform.fpl.model.Proceeding`. Since my new validation class was validating from `RespondentParty` the camel casing would no longer work.

In the second arg in the enum is display text. So it would render "You must add details to case name" etc

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
Jamesnelson117: No but `isEmpty(respondents1.get(0)` is redundant. As we can assume if respondents is not empty it will have at least one item in the collection.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorService.java
```diff
@@ -47,7 +48,7 @@ public CaseValidatorService(Validator validator) {
         List<String> errorList;
 
         errorList = caseData.stream()
```
lewisbirks: Would rename `caseData` to `constraintViolations` or something like that, `caseData` isn't accurate in describing what that variable contains.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorServiceTest.java
```diff
@@ -30,6 +32,10 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createPopulatedApplicants;
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createPopulatedChildren;
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createRespondents;
+
 @ExtendWith(SpringExtension.class)
 class CaseValidatorServiceTest {
     private CaseValidatorService caseValidatorService;
```
lewisbirks: Does this need to be reset before each method? Could the `@BeforeEach` be replaced with `@BeforeAll`?

Jamesnelson117: We are asserting for different levels of errors. We have assertions for high level errors (case errors) i.e 'you have not entered respondents'. We then have assertions for lower level errors i.e in "respondents - you did not enter first name". We need a fresh CaseValidatorService for this unfortunately.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/RespondentParty.java
```diff
@@ -27,6 +29,16 @@
     private final String litigationIssues;
     private final String litigationIssuesDetails;
 
+    @NotBlank(message = "Enter the respondent's full name")
```
swalker125: the error message right here? First name?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorServiceTest.java
```diff
@@ -182,8 +192,26 @@ void shouldNotReturnErrorsIfMandatoryFieldsHaveBeenCompleted() {
             .orders(Orders.builder()
                 .orderType(ImmutableList.of(OrderType.EMERGENCY_PROTECTION_ORDER))
                 .build())
-            .children1(initChildren())
-            .applicants(initApplicants())
+            .children1(createPopulatedChildren())
```
swalker125: I would maybe suggest extracting this to private methods, 30 lines of setup isn't cool

service/src/main/java/uk/gov/hmcts/reform/fpl/model/RespondentParty.java
Jamesnelson117: I'd agree but it's the same way we handle children and applicant's name. Mia want's it to continue to read as above.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorServiceTest.java
```diff
@@ -270,22 +234,53 @@ void shouldNotReturnAnErrorWhenGroundsAndGroundsForEPOAreCompleted() {
...
+                .id(UUID.randomUUID())
+                .value(Respondent.builder().party(
+                    RespondentParty.builder()
+                        .firstName("Sarah")
+                        .build())
+                    .build())
+                .build());
+    }
+
+    private List<Element<Applicant>> initApplicants(Boolean hasCompletedAddress) {
```
lewisbirks: boolean instead of Boolean

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorServiceTest.java
```diff
@@ -270,22 +234,53 @@ void shouldNotReturnAnErrorWhenGroundsAndGroundsForEPOAreCompleted() {
...
+        Address address;
```
lewisbirks: I would suggest that instead of having an Address variable using and Address.Builder variable and extracting the common logic of the if to before it (i.e. the addressLine2). Then when it needs to be used you can just build it.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorServiceTest.java
```diff
@@ -65,24 +67,8 @@ void shouldReturnErrorsIfMandatoryCaseSectionsHaveNotBeenCompleted() {
...
-            .applicants(List.of(Element.<Applicant>builder()
-                .id(UUID.randomUUID())
-                .value(Applicant.builder()
-                    .party(ApplicantParty.builder().build())
-                    .build())
-                .build()))
-            .build();
-
+    void shouldReturnErrorsWhenMandatoryFieldsHaveNotBeenCompleted() {
+        CaseData caseData = initEmptyMandatoryCaseData();
```
swalker125: I don't like init, just remove everywhere. emptyMandatoryCaseData() etc...

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorServiceTest.java
```diff
@@ -299,17 +289,22 @@ void shouldNotReturnAnErrorWhenGroundsAndGroundsForEPOAreCompleted() {
...
+        return Hearing.builder()
+            .timeFrame("Within 18 days")
+            .build();
+    }
+
+    private Grounds initGrounds() {
+        return Grounds.builder()
+            .thresholdDetails("details")
+            .thresholdReason(ImmutableList.of("reason"))
+            .build();
```
swalker125: I don't like this but I won't block the story because of it. To be refactored 👍

PR #563, opened 2019-10-28
Increase liveness delay to prevent flakey pod startup

akomisarek:
### Change description ###

As in title 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/fpl-case-service/values.preview.template.yaml
+++ b/charts/fpl-case-service/values.preview.template.yaml
@@ -5,6 +5,7 @@ tags:
 java:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
+  readinessDelay: 60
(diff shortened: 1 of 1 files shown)


charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -5,6 +5,7 @@ tags:
 java:
   image: ${IMAGE_NAME}
   ingressHost: ${SERVICE_FQDN}
+  readinessDelay: 60
```
geek-soft: Shouldn't that be done for liveness probe? Readiness does not restart pod, does it? Is't readiness used just to dec idea whether traffic should go to the pod or not?

akomisarek: Of course you are eright that should be liveness!

PR #566, opened 2019-10-29
Bump java charts version to use 250m and configure 250m for ccd pods

akomisarek:
### Change description ###

As in title. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/charts/fpl-case-service/values.preview.template.yaml
+++ b/charts/fpl-case-service/values.preview.template.yaml
@@ -30,6 +30,7 @@ ccd:
   ingressHost: ${SERVICE_FQDN}
   ingressIP: ${INGRESS_IP}
   consulIP: ${CONSUL_LB_IP}
+  cpuRequests: 250m
   idam-pr:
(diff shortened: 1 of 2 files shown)


charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -30,6 +30,7 @@ ccd:
   ingressHost: ${SERVICE_FQDN}
   ingressIP: ${INGRESS_IP}
   consulIP: ${CONSUL_LB_IP}
+  cpuRequests: 250m
```
geek-soft: @akomisarek why is that needed here since pipeline sets `--set global.devMode=true`?

akomisarek: I think the ccd-charts templates are not using the devcpu. 

After my change the CCD pod is getting:
    Restart Count:  0
    Limits:
      cpu:     2500m
      memory:  1Gi
    Requests:
      cpu:      250m
      memory:   512Mi

PR #571, opened 2019-10-30
Add SQL script that projects event permissions from last uploaded CCD definitions

geek-soft:

None

### Change description ###

Projects event permissions per user ans state from last version of uploaded CCD definitions.

Running `bin/utils/project-event-permissions.sh` script will give the following result:

state | user_role | event_id | event_name
-- | -- | -- | --
Open | caseworker-publiclaw-gatekeeper | otherProposal | Allocation proposal
Open | caseworker-publiclaw-solicitor | ordersNeeded | Orders and directions needed
Open | caseworker-publiclaw-solicitor | hearingNeeded | Hearing needed
Open | caseworker-publiclaw-solicitor | enterChildren | Children
Open | caseworker-publiclaw-solicitor | enterRespondents | Respondents
'*' | caseworker-publiclaw-solicitor | openCase | Start application
'*' | caseworker-publiclaw-solicitor | uploadDocuments | Documents

Optionally it allows narrowing down results to specific role by using `-s Open` flag or to specific role by using `-r caseworker-publiclaw-solicitor` flag.

```
[ ] Yes
[x] No
```

### DIFF
--- a/bin/utils/project-event-permissions.sh
+++ b/bin/utils/project-event-permissions.sh
@@ -2,22 +2,27 @@
 
 set -eu
 
+case_type_clause="WHERE reference = 'CARE_SUPERVISION_EPO'"
 state_where_clause=""
 role_where_clause=""
 
-while getopts ":s:r:h" option; do
+while getopts "c::s:r:h" option; do
   case ${option} in
+    c)
+      case_type_clause="WHERE reference = '${OPTARG}'"
+      ;;
     s)
-      state_where_clause="AND s.reference='${OPTARG}'"
+      state_where_clause="AND s.reference = '${OPTARG}'"
       ;;
     r)
-      role_where_clause="AND r.reference='${OPTARG}'"
+      role_where_clause="AND r.reference = '${OPTARG}'"
       ;;
     h)
       echo "
-        Usage: ${0} [-s state] [-r role]
+        Usage: ${0} [-c case type] [-s case state] [-r user role]
 
         Options:
+          -c: allows to select case type (optional, defaults to 'CARE_SUPERVISION_EPO')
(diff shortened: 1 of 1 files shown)


bin/utils/project-event-permissions.sh
```diff
@@ -0,0 +1,50 @@
...
+  SELECT COALESCE(s.reference, '*') as state,
+    r.reference as user_role,
+    e.reference as event_id,
+    e.name as event_name
+  FROM event e
+    LEFT JOIN event_pre_state eps ON e.id = eps.event_id
+    LEFT JOIN state s ON eps.state_id = s.id
+    JOIN event_acl ea ON e.id = ea.event_id
+    JOIN role r ON ea.role_id = r.id
+  WHERE e.case_type_id = (SELECT MAX(id) FROM case_type)
```
akomisarek: I am not biggest fan of that MAX(id), as it's not going to work i we have multiple case types, but no biggie, maybe we could flag somewhere it's not supported for multiple case_types?

geek-soft: Good point, I assumed we will have just one case type. I can add extra where clause so that only our case type is taken to consideration.

Generally when new case type is added that script should be changed anyway otherwise it will yield unexpected / confusing results.

geek-soft: I also head idea to make it as an external tool that could visualise results in a better way. Such tool could contain other projectors that for example render case state progression. Let's say this is MVP. I am not sure how useful would extra tool be to be honest so it is just an idea in my head ;)

akomisarek: Yeah good for now definitely! It improves our life :)

PR #600, opened 2019-11-07
FPLA 893: Solicitor email and name mandatory in application

lewisbirks:
https://tools.hmcts.net/jira/browse/FPLA-893

### Change description ###

Made the solicitor name and email mandatory through validation upon case submission.
Introduced the ability for an error message to contain multiple points spread over multiple lines.

```
[ ] Yes
[x] No
```


service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseValidatorServiceTest.java
```diff
@@ -174,13 +181,30 @@ void shouldNotReturnAnErrorWhenFirstRespondentHasFullNameButNotSecondRespondent(
...
+            .solicitor(solicitor())
             .respondents1(respondents())
             .build();
 
         List<String> errors = caseValidatorService.validateCaseDetails(caseData);
         assertThat(errors).isEmpty();
     }
 
+    @Test
+    void shouldReturnAnErrorWhenSolicitorDoesNotHaveNameAndEmail() {
```
Jamesnelson117: Do we need to test this separately? You've asserted this twice above no?

lewisbirks: Good point will remove

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -58,6 +58,9 @@ private final Respondent getFirstRespondent() {
     }
 
     private final Proceeding proceeding;
+
+    @NotNull(message = "Enter the solicitor's full name" + VALIDATION_SEPARATOR + "Enter the solicitor's email")
```
swalker125: I think the case data level validation is normally "you need to add details for solicitor" and then the Solicitor level validation is a "you need to fill in these fields"

lewisbirks: Usually yes but it wasn't mentioned on the ticket so I kept it consistent.

Jamesnelson117: Confirm with Mia, for consistency we should really have - 

In the applicant section :
- You need to add details to applicants
- You need to add details to solicitor (or similar)

lewisbirks: After discussion with Mia, have made it consistent with the rest of the messages when the event hadn't been accessed

PR #616, opened 2019-11-14
feat: parties can comply with own directions (FPLA-33)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-33

### Change description ###

Adds directions belonging to a party to a list of directions that they can comply with 

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -167,10 +167,12 @@ private boolean responseExists(DirectionResponse response, Element<Direction> di
     public void addAssigneeDirectionKeyValuePairsToCaseData(String assignee,
                                                             List<Element<Direction>> directions,
                                                             CaseDetails caseDetails) {
-        if (assignee.equals("courtDirections")) {
-            caseDetails.getData().put(assignee.concat("Custom"), extractPartyResponse(assignee, directions));
-        } else {
-            caseDetails.getData().put(assignee, extractPartyResponse(assignee, directions));
+        if (!assignee.equals(ALL_PARTIES.getValue())) {
```
lewisbirks: Is there a reason why we are always passing around the String representation of the enum? Would it be simpler to just use the enum or would that change if it is better extend outside the scope of the ticket?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -167,10 +167,12 @@ private boolean responseExists(DirectionResponse response, Element<Direction> di
...
+            if (assignee.equals(COURT.getValue())) {
+                caseDetails.getData().put(assignee.concat("Custom"), extractPartyResponse(assignee, directions));
```
lewisbirks: Haven't had a look in the rest of the code base yet but this appears to be the only place you are taking advantage of the fact that you have a String instead of an enum. Every other time you are comparing you are having to check the value of the enum. This just seems like extra work unless there is something that I am missing.

swalker125: opened up a tech debt ticket to address https://tools.hmcts.net/jira/browse/FPLA-964

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ComplyWithDirectionsController.java
```diff
@@ -40,14 +40,15 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
...
             directionHelperService.sortDirectionsByAssignee(caseData.getStandardDirectionOrder().getDirections());
 
-        sortedDirections.forEach((assignee, directions) -> directionHelperService
-            .addAssigneeDirectionKeyValuePairsToCaseData(assignee, allPartyDirections, caseDetails));
+        sortedDirections.forEach((assignee, directions) -> {
+            if (!assignee.equals(ALL_PARTIES.getValue())) {
+                directions.addAll(sortedDirections.get(ALL_PARTIES.getValue()));
+                directionHelperService.addAssigneeDirectionKeyValuePairsToCaseData(assignee, directions, caseDetails);
+            }
+        });
```
akomisarek: If that would be enum, the code would be a lot nicer with filters and regular streams, but we can leave it.

PR #618, opened 2019-11-14
impr: update notification for c21 (FPLA-936)

Gogetter:
### Update notification for C21 [XS] ###

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C21OrderEmailContentProvider.java
```diff
@@ -43,24 +48,34 @@ protected C21OrderEmailContentProvider(@Value("${ccd.ui.base.url}")String uiBase
...
         String hearingDate = "";
-        if (!ObjectUtils.isEmpty(caseData.getHearingDetails())) {
+        if (!isEmpty(caseData.getHearingDetails())) {
             hearingDate = " hearing " + dateFormatterService.formatLocalDateToString(
                 hearingBookingService.getMostUrgentHearingBooking(caseData.getHearingDetails()).getDate(),
                 FormatStyle.MEDIUM);
         }
         return Stream.of(subjectLine, hearingDate)
             .filter(StringUtils::isNotBlank)
-            .collect(joining(", "));
```
Jamesnelson117: Was this change intentional? I'd imagine we'd want to have a space between our collections no?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
```diff
@@ -196,4 +204,71 @@ public static Document createDocumentWithAttachedFile() {
...
+    public static List<Element<C21Order>> createC21Orders() {
+        return ImmutableList.of(
+            Element.<C21Order>builder()
+                .value(C21Order.builder()
+                    .orderTitle("Example Order")
+                    .orderDetails(
+                        "Example order details here - Lorem ipsum dolor sit amet, consectetur adipiscing elit")
+                    .orderDate(DATE_FORMATTER_SERVICE.formatLocalDateTimeBaseUsingFormat(
+                        LocalDateTime.now().plusDays(57), FORMAT_STYLE))
+                    .judgeAndLegalAdvisor(JudgeAndLegalAdvisor.builder()
```
Jamesnelson117: It might be of value splitting out building judgeAndLegalAdvisor into it's own method - `createJudgeAndLegalAdvisor`

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
```diff
@@ -196,4 +204,71 @@ public static Document createDocumentWithAttachedFile() {
...
+                    .orderDetails("Westeros")
+                    .orderDate(DATE_FORMATTER_SERVICE.formatLocalDateTimeBaseUsingFormat(
+                        LocalDateTime.now().plusDays(59), FORMAT_STYLE))
+                    .judgeAndLegalAdvisor(JudgeAndLegalAdvisor.builder()
+                        .legalAdvisorName("Baratheon")
+                        .judgeFullName("Tyrion Lannister")
+                        .judgeLastName("Lannister")
+                        .judgeTitle(HIS_HONOUR_JUDGE)
+                        .build())
+                    .document(DocumentReference.builder()
```
Jamesnelson117: Again might be of value splitting out into `createDocument`

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+                                    randomUUID().toString(), "binary"))
+                                .build())
+                            .build())
+                        .build()),
+                "respondents1", createRespondents(),
+                "familyManCaseNumber", familyManCaseNumber))
+            .build();
+    }
+
+    private CaseDetails populateCaseDetails() {
```
Jamesnelson117: Not a great method name. Let's be more descriptive.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+        return CaseDetails.builder()
+            .id(167888L)
+            .data(ImmutableMap.of("hearingDetails", createHearingBookings(LocalDate.now()),
+                "c21Orders", ImmutableList.of(
+                    Element.<C21Order>builder()
+                        .value(C21Order.builder()
+                            .orderTitle("Example Order")
+                            .orderDetails(
+                                "Example order details here - Lorem ipsum dolor sit amet, consectetur adipiscing elit")
+                            .judgeAndLegalAdvisor(JudgeAndLegalAdvisor.builder()
```
Jamesnelson117: If we split out our previous JudgeAndLegalAdvisor method into a separate static method, we could reuse here.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -0,0 +1,148 @@
...
+                                .legalAdvisorName("Peter Parker")
+                                .judgeLastName("Judy")
+                                .judgeTitle(HER_HONOUR_JUDGE)
+                                .build())
+                            .document(DocumentReference.builder()
```
Jamesnelson117: Re-use static method if you choose to extract.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C21OrderEmailContentProvider.java
Gogetter: yes it was as _hearingDate_ already had a space.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C21OrderEmailContentProvider.java
```diff
@@ -43,24 +48,34 @@ protected C21OrderEmailContentProvider(@Value("${ccd.ui.base.url}")String uiBase
...
             "subjectLine", subjectLine,
             "localAuthorityOrCafcass", localAuthorityNameLookupConfiguration.getLocalAuthorityName(localAuthorityCode),
             "hearingDetailsCallout", buildSubjectLineWithHearingBookingDateSuffix(subjectLine, caseData),
+            "linkToDocStore", mostRecentUploadedC21DocumentUrl(caseData.getC21Orders()),
             "reference", String.valueOf(caseDetails.getId()),
             "caseUrl", uiBaseUrl + "/case/" + JURISDICTION + "/" + CASE_TYPE + "/" + caseDetails.getId()
         );
     }
 
+    private String mostRecentUploadedC21DocumentUrl(final List<Element<C21Order>> c21Orders) {
```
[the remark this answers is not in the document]
Gogetter: @Jamesnelson117 given we will be sending out the notification once the event occurs, will using the last upload not be preferred?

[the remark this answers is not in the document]
Jamesnelson117: Yep, ofc

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
Gogetter: good suggestion, updated

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
Gogetter: updated

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C21OrderEmailContentProvider.java
```diff
@@ -43,24 +48,34 @@ protected C21OrderEmailContentProvider(@Value("${ccd.ui.base.url}")String uiBase
         // Validation within our frontend ensures that the following data is present
         CaseData caseData = objectMapper.convertValue(caseDetails.getData(), CaseData.class);
         final String subjectLine = buildSubjectLine(caseData);
-        return ImmutableMap.of(
+        return Map.of(
             "subjectLine", subjectLine,
             "localAuthorityOrCafcass", localAuthorityNameLookupConfiguration.getLocalAuthorityName(localAuthorityCode),
             "hearingDetailsCallout", buildSubjectLineWithHearingBookingDateSuffix(subjectLine, caseData),
+            "linkToDocStore", mostRecentUploadedC21DocumentUrl(caseData.getC21Orders()),
```
akomisarek: I don't like this implementation to be honest. The event should simply send the C21 Order which was attached, we shouldn't reach to the case and find out which one was added. 

Probably logic will be the same but in different place - event itself should be descriptive and know what happened.

akomisarek: Why linkToDocStore and not simply linkToDocument? Just curious not 100% sure which one is better.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -0,0 +1,139 @@
...
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createDocumentReference;
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createHearingBookings;
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createJudgeAndLegalAdvisor;
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createRespondents;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {JacksonAutoConfiguration.class, C21OrderEmailContentProvider.class,
+    HearingBookingService.class, LocalAuthorityNameLookupConfiguration.class, DateFormatterService.class})
+class C21OrderEmailContentProviderTest {
+    private final LocalDate date = LocalDate.now();
```
akomisarek: I would be more specific than "date", TODAY or something like that would make the tests more readable.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -0,0 +1,139 @@
...
+        assertThat(returnedParameters)
+            .extracting("subjectLine", "localAuthorityOrCafcass", "hearingDetailsCallout", "reference",
+                 "caseUrl")
+            .containsExactly(subjectLine, "Example Local Authority",
+                subjectLine + ", hearing " + dateFormatterService.formatLocalDateToString(date, FormatStyle.MEDIUM),
+                "167888", "/case/" + JURISDICTION + "/" + CASE_TYPE + "/167888");
+
+        /*  extracting separately as document_binary_url in populateCaseDetailsWithSingleC21Element()
+            method uses UUID.randomUUID()
+         */
```
akomisarek: Don't use randomUUID? Or pass externally the UUID value which should be used? If our test is checking not equal to empty on unit test level it's always smell to me. I change the implementation to always generate ABC and is it going to be correct implementation? Unit tests should spot that.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C21OrderEmailContentProvider.java
Gogetter: good suggestion on moving into the event itself in place of reaching into the _CaseData_. With regards the _linkToDocStore_, is it ok if I update the template on Gov dot notify and make this _linkToDocument_ as its more readable? @akomisarek

Gogetter:
Hi @akomisarek, requested changes made. Please could you help take another look? thanks

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -38,71 +39,102 @@
...
+            cafcassLookupConfiguration);
 
         given(localAuthorityNameLookupConfiguration.getLocalAuthorityName(LOCAL_AUTHORITY_CODE))
             .willReturn("Example Local Authority");
 
+        given(cafcassLookupConfiguration.getCafcass(LOCAL_AUTHORITY_CODE))
+            .willReturn((new CafcassLookupConfiguration.Cafcass(CAFCASS_NAME, CAFCASS_EMAIL)));
+
         familyManCaseNumber = RandomStringUtils.randomAlphabetic(8);
+        documentId = randomUUID().toString();
```
akomisarek: Really picky, but if we have got documentId I believe it should be in UUID format and defer toString when it's required.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -38,71 +39,102 @@
...
-        /*  extracting separately as document_binary_url in populateCaseDetailsWithSingleC21Element()
-            method uses UUID.randomUUID()
-         */
-        assertThat(returnedParameters).extracting("linkToDocStore").isNotEqualTo("");
+        assertThat(returnedLocalAuthorityParameters)
+            .extracting("subjectLine", "localAuthorityOrCafcass", "hearingDetailsCallout",
+                "linkToDocument", "reference", "caseUrl")
+            .containsExactly(subjectLine, "Example Local Authority",
+                (subjectLine + ", hearing " + dateFormatterService.formatLocalDateToString(today, FormatStyle.MEDIUM)),
+                documentUrl, "167888", "/case/" + JURISDICTION + "/" + CASE_TYPE + "/167888");
```
akomisarek: Isn't this another test as we are checking something else? The unit test could assert one thing.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -38,71 +39,102 @@
...
-
-        Map<String, Object> returnedParameters = c21OrderEmailContentProvider.buildC21OrderNotification(caseDetails,
-            LOCAL_AUTHORITY_CODE);
-
-        final String subjectLine = "Jones, " + familyManCaseNumber;
-
-        assertThat(returnedParameters)
-            .extracting("subjectLine", "localAuthorityOrCafcass", "hearingDetailsCallout", "linkToDocStore",
-                "reference", "caseUrl")
+        final String mostRecentDocumentUrl = "http://dm-store:8080/documents/79ec80ec-7be6-493b-b4e6-f002f05b7079/binary";
```
akomisarek: Similar here, we are checking multiple scenarios.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CreateC21OrderService.java
```diff
@@ -96,4 +106,13 @@ private String getCourtName(String courtName) {
                     .formatLocalDateToString(child.getDateOfBirth(), FormatStyle.LONG) : ""))
             .collect(toList());
     }
+
+    private String mostRecentUploadedC21DocumentUrl(final List<Element<C21Order>> c21Orders) {
+        return getLast(c21Orders.stream()
```
akomisarek: Hmm I am not sure if I like the approach here, getLast relies on the order provided to us by CCD, I think it's bad practice and makes code a lot more difficult in future to change than required. I know it's probably correct in this instance (and it will be 99% of the time in future), but maybe we should return really what mostRecent means? order by date and take the latest one?

akomisarek: Probbaly before collecting mapping would be a bit nicer :)

akomisarek: @joez-f what do you think about the above?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/C21OrderControllerTest.java
```diff
@@ -137,4 +176,58 @@ private AboutToStartOrSubmitCallbackResponse makeRequest(CallbackRequest request
...
+                    "c21Orders", createC21Orders(),
+                    "hearingDetails", createHearingBookings(dateIn3Months),
+                    "respondents1", createRespondents(),
+                    "caseLocalAuthority", LOCAL_AUTHORITY_CODE,
+                    "familyManCaseNumber", FAMILY_MAN_CASE_NUMBER))
+                .build())
+            .build();
+    }
+
+    private Map<String, Object> similarParameters() {
```
akomisarek: Not sure if I like the name similar, similar to what? commonNotification?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CreateC21OrderService.java
joez-f: I like sorting by orderDate and then taking the latest one - good to do it explicitly (although I do agree that in this instance I don't see a way that the list of c21Orders could ever be out of order. so not a huge problem IMO, but let's explicitly return mostRecent if it's not too difficult)

service/src/main/java/uk/gov/hmcts/reform/fpl/events/C21OrderEvent.java
```diff
@@ -1,10 +1,25 @@
...
+
+    public C21OrderEvent(CallbackRequest callbackRequest, String authorization, String userId,
+                         C21OrderEventData c21OrderEventData) {
         super(callbackRequest, authorization, userId);
+        this.c21OrderEventData = c21OrderEventData;
+    }
+
+    @Value
+    @Builder
+    public static class C21OrderEventData {
```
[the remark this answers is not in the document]
Gogetter: yes was thinking that myself after writing. The idea was to split the document link out. But given its only one field we can pass as a param

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CreateC21OrderService.java
Gogetter: I picked this option as the _orderDate_ is currently stored in string

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/C21OrderControllerTest.java
Gogetter: updated

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/C21OrderControllerTest.java
```diff
@@ -219,14 +219,14 @@ private CallbackRequest buildCallbackRequest() {
 
     private Map<String, Object> c21CafcassParameters() {
```
akomisarek: Wouldn't it be beneficial to name them "expected" that looks like they are used in beforeEach and used for setup.

service/src/main/java/uk/gov/hmcts/reform/fpl/config/DocumentConfiguration.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts.reform.fpl.config;
+
+import lombok.Data;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Data
+@Configuration
+public class DocumentConfiguration {
```
akomisarek: I don't like the name DocumentConfiguration as it doesn't have to do anything with the document - it's the GatewayConfiguration and are configuring the gatewayUrl. 

At least DocumentGateway? but still the use is more broader than just documents.

service/src/main/java/uk/gov/hmcts/reform/fpl/config/DocumentConfiguration.java
```diff
@@ -0,0 +1,15 @@
...
+    private final String documentHostUrl;
```
akomisarek: As above.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CreateC21OrderService.java
```diff
@@ -107,12 +107,22 @@ private String getCourtName(String courtName) {
             .collect(toList());
     }
 
-    private String mostRecentUploadedC21DocumentUrl(final List<Element<C21Order>> c21Orders) {
-        return getLast(c21Orders.stream()
+    public String mostRecentUploadedC21DocumentUrl(final List<Element<C21Order>> c21Orders) {
```
akomisarek: Please implement unit test for this method now.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CreateC21OrderService.java
```diff
@@ -107,12 +107,22 @@ private String getCourtName(String courtName) {
...
+
+        final String documentHostUrl = documentConfiguration.getDocumentHostUrl();
+
+        try {
+            URL url = new URL(mostRecentUploadedC21DocumentUrl);
+            mostRecentUploadedC21DocumentUrl = documentHostUrl + url.getPath();
+        } catch (MalformedURLException e) {
+            log.error("mostRecentUploadedC21DocumentUrl url incorrect.", e);
+        }
+        return mostRecentUploadedC21DocumentUrl;
```
akomisarek: Could we extract to sdeparatee method? A bit messy. Could allow some mapping. 

Can you swap from URL to URI? UIR is a bit more lightweight class. 

Not sure about returning original URL if it fail to parse, probably would be better to return some placeholder text? It would be invalid URL so not sure.
Are you sure the "e" contains the URL for the debugging purposes?

service/src/main/resources/application.yaml
```diff
@@ -104,6 +105,7 @@ docmosis:
 
 document_management:
   url: http://localhost:3453
+  dm_host_url: https://gateway-${SERVICE_FQDN}
```
akomisarek: This cannot contain variable + it's not dm host.

service/src/main/resources/application.yaml
```diff
@@ -44,6 +44,7 @@ docmosis:
 
 document_management:
   url: http://fake-document-management-store-api
+  dm_host_url: http://fake-document-gateway-host
```
akomisarek: It's not DM host.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -118,7 +118,7 @@
 
     @Nested
     class C2UploadedNotificationChecks {
-        final String documentUrl = "http://dm-store:8080/documents/79ec80ec-7be6-493b-b4e6-f002f05b7079/binary";
+        final String mostRecentUploadedDocumentUrl = "http://dm-store:8080/documents/79ec80ec-7be6-493b-b4e6-f002f05b7079/binary";
```
akomisarek: Really picky but it would be better if we do gateway URL here to avoid confusion.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CreateC21OrderServiceTest.java
```diff
@@ -39,12 +40,14 @@
 
     private DateFormatterService dateFormatterService = new DateFormatterService();
     private HmctsCourtLookupConfiguration hmctsCourtLookupConfiguration = new HmctsCourtLookupConfiguration(CONFIG);
+    private DocumentConfiguration documentConfiguration = new DocumentConfiguration("");
```
akomisarek: Not sure why this needs to be field if neverused in class (apart from beforeEach)

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -139,6 +157,24 @@ void shouldReturnExactC21NotificationParametersWithMostRecentUploadedDocumentUrl
                 "/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345");
     }
 
+    @Test
+    void shouldReturnExactC21LocalAuthorityNotificationParametersWithMostRecentUploadedDocumentUrl() {
```
akomisarek: What is different in this test from the one declared above in line 115? Multi c21 are not relevant on that level as the URL is just provided in the contentProvider so why we need to test separate scenario? Is it even useful?

charts/fpl-case-service/values.aat.template.yaml
```diff
@@ -13,6 +13,7 @@ java:
     SPRING_SECURITY_ENABLED: true
     SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_ISSUERURI: https://forgerock-am.service.core-compute-idam-aat.internal:8443/openam/oauth2/hmcts
     SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWKSETURI: https://idam-api.aat.platform.hmcts.net/o/jwks
+    DOCUMENT_MANAGEMENT_GATEWAY_URL: https://gateway-${SERVICE_FQDN}
```
akomisarek: Not sure why you insist of using the document management, I am not sure if we should say it's document management, as this is just generic gateway? Different purpose?

charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -13,6 +13,7 @@ java:
     SPRING_SECURITY_ENABLED: true
     SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_ISSUERURI: https://forgerock-am.service.core-compute-idam-aat.internal:8443/openam/oauth2/hmcts
     SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWKSETURI: https://idam-api.aat.platform.hmcts.net/o/jwks
+    DOCUMENT_MANAGEMENT_GATEWAY_URL: https://gateway-${SERVICE_FQDN}
```
akomisarek: doesn't we need to have -ccd? 
The above will result in
gateway-fpl-case-service-pr-618.service.core-compute-preview.internal. Is this correct URL? Can you access it?

infrastructure/prod.tfvars
```diff
@@ -9,3 +9,4 @@ docmosis_vault = "docmosisiaasprodkv"
 
 idam_client_id = "fpl_case_service"
 idam_client_redirect_uri = "https://fpl-case-service-prod.service.core-compute-prod.internal/oauth2/callback"
+document_management_gateway_url = "https://gateway-ccd.prod.platform.hmcts.net"
```
akomisarek: Apart from that you need to configure it for the demo (which is only real environment using the terraform to that extend)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/C21OrderControllerTest.java
```diff
@@ -155,11 +155,11 @@ void shouldTriggerC21EventWhenSubmitted() throws Exception {
 
         verify(notificationClient, times(1)).sendEmail(
             eq(C21_ORDER_NOTIFICATION_TEMPLATE), eq(LOCAL_AUTHORITY_EMAIL_ADDRESS),
-            eq(c21LocalAuthorityParameters()), eq(expectedCaseReference));
+            eq(expectedC21LocalAuthorityParameters()), eq(expectedCaseReference));
```
akomisarek: 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/config/GatewayConfiguration.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts.reform.fpl.config;
+
+import lombok.Data;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+
+@Data
+@Configuration
+public class GatewayConfiguration {
+    private final String gatewayUrl;
```
akomisarek: Isn't it enough to name it url? it's obvious its gateway as we have it in the Class name :)

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
Gogetter: yes looks very similar. Line 115 bit checks for Cafcass parameters and this checks for LA params

service/src/main/java/uk/gov/hmcts/reform/fpl/config/GatewayConfiguration.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public GatewayConfiguration(@Value("${document_management.gateway_url}") String gatewayUrl) {
```
akomisarek: Maybe gateway? And not document management, but not that fussed if you really like the doc_management prefix.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/C21OrderController.java
```diff
@@ -121,4 +129,17 @@ private Document getDocument(String authorization,
...
+        final String gatewayUrl = gatewayConfiguration.getGatewayUrl();
+
+        try {
+            URI uri = new URI(mostRecentUploadedC21DocumentUrl);
+            return gatewayUrl + uri.getPath();
+        } catch (URISyntaxException e) {
+            log.error(mostRecentUploadedC21DocumentUrl + " url incorrect.", e);
+        }
+        return "";
+    }
```
akomisarek: I am not sure if the controller is however the good place for this logic? maybe could be done in notification content provider? This is the place which really needs to transform the URL, on the controller level it's ok, I think?

akomisarek: Would be easier to test as well - and this piece of code should be unit tested.

charts/fpl-case-service/values.aat.template.yaml
Gogetter: for now was trying to make it as specific as possible. But you have made a good

Gogetter: Lets keep as gateway given the class name 👍

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/C21OrderEmailContentProviderTest.java
```diff
@@ -0,0 +1,184 @@
...
+                "linkToDocument", "reference", "caseUrl")
+            .containsExactly(subjectLine, "Example Cafcass",
+                subjectLine + ", hearing " + dateFormatterService.formatLocalDateToString(today, FormatStyle.MEDIUM),
+                 mostRecentDocumentUrl, "12345", "/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345");
+    }
+
+    @Test
+    void shouldReturnExactC21LocalAuthorityNotificationParametersWithMostRecentUploadedDocumentUrl() {
+        final String mostRecentDocumentUrl = "http://dm-store:8080/documents/79ec80ec-7be6-493b-b4e6-f002f05b7079/binary";
+        CaseDetails caseDetails = createCaseDetailsWithMultipleC21Elements();
```
akomisarek: As agreed this test should be removed.

PR #628, opened 2019-11-18
refactor: make use of DirectionAssignee enum values across CMO and SDO

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-964

### Change description ###

Refactor so that instead of calling getValue() on enum to get String value, make use of the DirectionAssignee enum directly (tech debt cleanup)

```
[ ] Yes
[X] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ComplyWithDirectionsControllerTest.java
```diff
@@ -52,7 +53,7 @@
     private static final String USER_ID = "1";
 
     @Test
-    void aboutToStartCallbackShouldSplitAllPartiesDirectionsIntoSeparateCollections() throws Exception {
+    void aboutToStartCallbackShouldAddAllPartiesDirectionsIntoSeparateRoleCollections() throws Exception {
```
swalker125: no, please revert Joe. This PR should just address using enums.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ComplyWithDirectionsControllerTest.java
```diff
@@ -64,11 +65,7 @@ void aboutToStartCallbackShouldSplitAllPartiesDirectionsIntoSeparateCollections(
 
         CaseData caseData = makeRequest(request, "about-to-start");
 
-        assertThat(roleDirectionsContainExpectedAllPartiesDirection(caseData.getLocalAuthorityDirections()));
-        assertThat(roleDirectionsContainExpectedAllPartiesDirection(caseData.getCafcassDirections()));
-        assertThat(roleDirectionsContainExpectedAllPartiesDirection(caseData.getCourtDirectionsCustom()));
-        assertThat(roleDirectionsContainExpectedAllPartiesDirection(caseData.getOtherPartiesDirections()));
-        assertThat(roleDirectionsContainExpectedAllPartiesDirection(caseData.getParentsAndRespondentsDirections()));
+        assertThat(collectionsContainDirectionsForRoleAndAllParties(caseData));
```
swalker125: again, please revert the assertion change

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ComplyWithDirectionsControllerTest.java
```diff
@@ -143,12 +140,24 @@ private CaseData makeRequest(CallbackRequest request, String endpoint) throws Ex
...
+    @SuppressWarnings("LineLength")
+    private boolean collectionsContainDirectionsForRoleAndAllParties(CaseData caseData) {
+        return roleDirectionsContainExpectedDirections(caseData.getLocalAuthorityDirections(), LOCAL_AUTHORITY)
+            && roleDirectionsContainExpectedDirections(caseData.getCafcassDirections(), CAFCASS)
+            && roleDirectionsContainExpectedDirections(caseData.getCourtDirectionsCustom(), COURT)
+            && roleDirectionsContainExpectedDirections(caseData.getOtherPartiesDirections(), OTHERS)
+            && roleDirectionsContainExpectedDirections(caseData.getParentsAndRespondentsDirections(), PARENTS_AND_RESPONDENTS)
+            && roleDirectionsContainExpectedDirections(emptyList(), ALL_PARTIES);
+    }
+
```
swalker125: why

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ComplyWithDirectionsController.java
```diff
@@ -40,14 +40,15 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
...
             directionHelperService.sortDirectionsByAssignee(caseData.getStandardDirectionOrder().getDirections());
 
-        sortedDirections.forEach((assignee, directions) -> directionHelperService
-            .addAssigneeDirectionKeyValuePairsToCaseData(assignee, allPartyDirections, caseDetails));
+        sortedDirections.forEach((assignee, directions) -> {
+            if (assignee != ALL_PARTIES) {
+                directions.addAll(sortedDirections.get(ALL_PARTIES));
+                directionHelperService.addAssigneeDirectionKeyValuePairsToCaseData(assignee, directions, caseDetails);
+            }
+        });
```
swalker125: why?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -195,14 +196,16 @@ private String getFirstApplicantName(CaseData caseData) {
             .orElse("");
     }
 
-    private Map<String, List<Map<String, String>>> getGroupedDirections(CaseData caseData) throws IOException {
+    private ImmutableMap<String, List<Map<String, String>>> getGroupedDirections(CaseData caseData)
+        throws IOException {
```
swalker125: why?

swalker125: should be Map<DirectionAssignee, List...> here no?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -195,14 +196,16 @@ private String getFirstApplicantName(CaseData caseData) {
...
         if (caseData.getStandardDirectionOrder() == null) {
             return ImmutableMap.of();
         }
 
-        Map<String, List<Element<Direction>>> groupedDirections = directionHelperService.sortDirectionsByAssignee(
+        Map<DirectionAssignee, List<Element<Direction>>> groupedDirections =
+            directionHelperService.sortDirectionsByAssignee(
             directionHelperService.numberDirections(caseData.getStandardDirectionOrder().getDirections()));
 
         ImmutableMap.Builder<String, List<Map<String, String>>> formattedDirections = ImmutableMap.builder();
```
swalker125: should this not be 
```suggestion
        ImmutableMap.Builder<DirectionAssignee, List<Map<String, String>>> formattedDirections = ImmutableMap.builder();
```

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperServiceTest.java
```diff
@@ -681,9 +682,10 @@ void shouldSortDirectionsIntoSeparateEntriesInMapWhenManyAssignees() {
                 .addAll(buildDirections(COURT))
                 .build();
 
-            Map<String, List<Element<Direction>>> sortedDirections = service.sortDirectionsByAssignee(directions);
+            Map<DirectionAssignee, List<Element<Direction>>> sortedDirections = service.sortDirectionsByAssignee(
+                directions);
```
swalker125: I don't think this is spaced down correctly

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
joez-f: the thing we return from this is getting added to case data with data.putAll(), so don't we need our key of the map to be a string? if we use Map<DIrectionAssignee, ... then we need to stream it so we can getKey().getValue() inside putAll(). can do either approach

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperServiceTest.java
joez-f: will move onto new line after the =

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
lewisbirks: Could we possibly annotate the `getValue` method in the enum with `@JsonValue`?

lewisbirks: I think this would then serialise the enum using the result of that method when it is being converted in json

joez-f: this is the method for getting the data for the docmosis template - the docmosis service expects a string, so if we change to DirectionAssignee then docmosis cannot render it. there's no JSON conversion so we can't use JsonValue. I think we have to keep it as is

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
joez-f: see above

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperServiceTest.java
joez-f: never mind - think this is actually spaced correctly
image

PR #634, opened 2019-11-19
refactor: custom directions match a standard pattern

swalker125:
ParentsAndRespondentsDirections -> RespondentDirections

The benefits of this PR will be huge in FPL-894, when using custom direction collections to comply on behalf of others.

** Reverted changes to enum value, this should no longer be a breaking change **

```
[] Yes
[x] No
```


swalker125:
@akomisarek if you would rather not have the breaking change I can change the ENUM back to PARENTS_AND_RESPONDENTS. Let me know.

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/DirectionAssignee.java
```diff
@@ -3,7 +3,7 @@
 public enum DirectionAssignee {
     ALL_PARTIES("allParties"),
     LOCAL_AUTHORITY("localAuthorityDirections"),
-    PARENTS_AND_RESPONDENTS("parentsAndRespondentsDirections"),
+    PARENTS_AND_RESPONDENTS("respondentDirections"),
```
Jamesnelson117: Shall we rename this enum to `RESPONDENTS`?

Jamesnelson117:
Why exactly is this needed @swalker125

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/DirectionAssignee.java
swalker125: breaking change to cases in AAT and demo, I had it like that, but have reverted. Been open for a few days and no reviews so I thought the BREAKING CHANGE flag scared people off

swalker125:
> Why exactly is this needed @swalker125

I want to be able to do put.(key.concat("Custom"), ...) without having an if to catch this case that doesn't match

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/DirectionAssignee.java
lewisbirks: Can confirm was very scared 😨

PR #653, opened 2019-11-27
feat: make judge input system in hearing details consistent with system (FPLA-931)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-931

### Change description ###

Update hearing details to use the judge input system we use in other places within the system. Update SDO to capture judge and legal advisor now entered in hearing details. Also fixed issues with SDO document not correctly handling situation with no legal advisor

```
[ ] Yes
[X] No
```


e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -113,8 +115,9 @@ Scenario('HMCTS admin enters hearing details and submits', async (I, caseViewPag
   I.seeAnswerInTab(5, 'Hearing 1', '', hearingDetails[0].type.welsh);
   I.seeAnswerInTab(5, 'Hearing 1', '', hearingDetails[0].type.somethingElse);
   I.seeAnswerInTab(6, 'Hearing 1', 'Give details', hearingDetails[0].giveDetails);
-  I.seeAnswerInTab(7, 'Hearing 1', 'Judge or magistrate\'s title', hearingDetails[0].judgeTitle);
-  I.seeAnswerInTab(8, 'Hearing 1', 'Judge or magistrate\'s last name', hearingDetails[0].lastName);
+  I.seeAnswerInTab(1, 'Judge and legal advisor', 'Judge or magistrate\'s title', 'Her Honour Judge');
```
akomisarek: As discussed you should use your test fixture to assert not hardcoded strings.

e2e/tests/judiciaryAdministersCaseAfterSubmission_test.js
```diff
@@ -52,8 +54,9 @@ Scenario('Judiciary enters hearing details and submits', async (I, caseViewPage,
   I.seeAnswerInTab(5, 'Hearing 1', '', hearingDetails[0].type.welsh);
   I.seeAnswerInTab(5, 'Hearing 1', '', hearingDetails[0].type.somethingElse);
   I.seeAnswerInTab(6, 'Hearing 1', 'Give details', hearingDetails[0].giveDetails);
-  I.seeAnswerInTab(7, 'Hearing 1', 'Judge or magistrate\'s title', hearingDetails[0].judgeTitle);
-  I.seeAnswerInTab(8, 'Hearing 1', 'Judge or magistrate\'s last name', hearingDetails[0].lastName);
+  I.seeAnswerInTab(1, 'Judge and legal advisor', 'Judge or magistrate\'s title', 'Her Honour Judge');
```
akomisarek: Fixture should be used.

e2e/tests/judiciaryAdministersCaseAfterSubmission_test.js
```diff
@@ -36,8 +36,10 @@ Before(async (I, caseViewPage, submitApplicationEventPage, enterFamilyManCaseNum
 Scenario('Judiciary enters hearing details and submits', async (I, caseViewPage, loginPage, addHearingBookingDetailsEventPage) => {
   await caseViewPage.goToNewActions(config.administrationActions.addHearingBookingDetails);
   await addHearingBookingDetailsEventPage.enterHearingDetails(hearingDetails[0]);
+  await addHearingBookingDetailsEventPage.enterJudgeAndLegalAdvisor('Reed', 'Rupert Robert');
```
akomisarek: Entering judgeAndLegal should be embedded in the method above.

e2e/tests/localAuthorityMaintainsCaseAfterSDO_test.js
```diff
@@ -38,8 +38,10 @@ Before(async (I, caseViewPage, submitApplicationEventPage, enterFamilyManCaseNum
     await I.navigateToCaseDetails(caseId);
     await caseViewPage.goToNewActions(config.administrationActions.addHearingBookingDetails);
     await addHearingBookingDetailsEventPage.enterHearingDetails(hearingDetails[0]);
+    await addHearingBookingDetailsEventPage.enterJudgeAndLegalAdvisor('Reed', 'Rupert Robert');
```
akomisarek: same comment as in another place

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelperTest.java
```diff
@@ -35,6 +35,18 @@ void shouldReturnEmptyJudgeNameWhenJudgeTitleIsNotProvided() {
         assertThat(judgeTitleAndName).isEqualTo("");
     }
 
+    @Test
+    void shouldReturnBlankPlaceholderWhenNoJudgeEnteredForDraftSDO() {
+        JudgeAndLegalAdvisor judgeAndLegalAdvisor = JudgeAndLegalAdvisor.builder()
+            .judgeTitle(null)
+            .build();
+
+        String judgeTitleAndName = JudgeAndLegalAdvisorHelper.formatJudgeTitleAndNameForDraftSDO(judgeAndLegalAdvisor,
```
akomisarek: (0) what do you think about static import? Would it increase readability?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
```diff
@@ -113,11 +109,8 @@ void shouldMapCaseDataWhenEmptyListValues() throws IOException {
         Map<String, Object> templateData = caseDataExtractionService
             .getStandardOrderDirectionData(caseData);
 
-        assertThat(templateData.get("judgeAndLegalAdvisor")).isEqualTo(ImmutableMap.of(
```
akomisarek: (0) I know it's not your code, but not sure why this test scenario is different than above - what benefit it gives us to test half of the fields empty? And repeat the test for all of them empty.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
```diff
@@ -75,10 +73,7 @@ void shouldMapEmptyCaseDataForDraftSDO() throws IOException {
         Map<String, Object> templateData = caseDataExtractionService
             .getStandardOrderDirectionData(CaseData.builder().build());
 
-        assertThat(templateData.get("judgeAndLegalAdvisor")).isEqualToComparingFieldByField(ImmutableMap.of(
-            "judgeTitle", EMPTY_PLACEHOLDER,
-            "legalAdvisorName", EMPTY_PLACEHOLDER
-        ));
+        assertThat(templateData.get("judgeTitleAndName")).isEqualTo(EMPTY_PLACEHOLDER);
```
akomisarek: We lost the check for legalAdvisorName, is it intended?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
```diff
@@ -24,7 +25,19 @@ public static String formatJudgeTitleAndName(JudgeAndLegalAdvisor judgeAndLegalA
             .orElse("");
     }
 
+    public static String formatJudgeTitleAndNameForDraftSDO(JudgeAndLegalAdvisor judgeAndLegalAdvisor,
```
akomisarek: This shouldn't be died to SDO I think, better would be if method suggest why it's different behaviour. 

What if we would like to use it in another place?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
```diff
@@ -24,7 +25,19 @@ public static String formatJudgeTitleAndName(JudgeAndLegalAdvisor judgeAndLegalA
...
+                                                            String emptyPlaceholder) {
+        return Optional.ofNullable(judgeAndLegalAdvisor)
+            .filter(judge -> judge.getJudgeTitle() != null)
+            .map(JudgeAndLegalAdvisorHelper::mapJudgeOrAdvisor)
+            .orElse(emptyPlaceholder);
```
akomisarek: (0) Just a thought - this could return Optional.empty if not present and the client would need to format if required with blank.

(-1) Or/and - Above methods should be refactored to make use of this one (the one we are doing map + orElse "")

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
joez-f: for now I will add in assertion that it is empty string - I haven't set legal advisor to be EMPTY_PLACEHOLDER anywhere because it is OK to have no legal advisor. In next story when magistrate is selected this may be handled differently

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelperTest.java
joez-f: agree 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
joez-f: actually a redundant method - removed 👍

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
joez-f: agree that it doesn't really have much benefit but we will leave it in as don't want to touch this code in this PR

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
joez-f: decided not to make any changes to this at the moment due to tight deadline for this story

e2e/pages/events/addHearingBookingDetailsEvent.page.js
```diff
@@ -1,4 +1,5 @@
 const { I } = inject();
+const judgeAndLegalAdvisor = require('../../fragments/judgeAndLegalAdvisor');
```
geek-soft: I tend to name these with 'Fragment' suffix e.g. `judgeAndLegalAdvisorFragment`.

Having `judgeAndLegalAdvisor` looks bit wired to me because it looks a bit like a something that represents person (judge and / or legal advisor).

I am curious your opinion.

e2e/tests/judiciaryAdministersCaseAfterSubmission_test.js
```diff
@@ -52,8 +52,9 @@ Scenario('Judiciary enters hearing details and submits', async (I, caseViewPage,
   I.seeAnswerInTab(5, 'Hearing 1', '', hearingDetails[0].type.welsh);
   I.seeAnswerInTab(5, 'Hearing 1', '', hearingDetails[0].type.somethingElse);
   I.seeAnswerInTab(6, 'Hearing 1', 'Give details', hearingDetails[0].giveDetails);
-  I.seeAnswerInTab(7, 'Hearing 1', 'Judge or magistrate\'s title', hearingDetails[0].judgeTitle);
-  I.seeAnswerInTab(8, 'Hearing 1', 'Judge or magistrate\'s last name', hearingDetails[0].lastName);
+  I.seeAnswerInTab(1, 'Judge and legal advisor', 'Judge or magistrate\'s title', hearingDetails[0].judgeAndLegalAdvisor.judgeTitle);
```
geek-soft: IMO `hearingDetails[0].judgeAndLegalAdvisor` should be split into `hearingDetails[0].judge` and `hearingDetails[0].legalAdvisor`.

It might be just comment for the future. Cost of refactoring it now might be too big and might be actually better to keep it consistent.

Just a note.

e2e/pages/events/addHearingBookingDetailsEvent.page.js
joez-f: Good idea - this was an existing fragment used in other tests already so I did not alter the name but in future I agree that we could suffix these with 'Fragment'

e2e/tests/judiciaryAdministersCaseAfterSubmission_test.js
geek-soft: Not your fault but the fact you had to make same change in 3 different files is smell of a bad design / implementation.

Just a note.

joez-f: Might be nice to split them out in the future, but to be consistent with our java models/ccd definitions/rest of the code I kept it together (we keep them together everywhere)

service/src/main/java/uk/gov/hmcts/reform/fpl/model/HearingBooking.java
```diff
@@ -27,8 +28,7 @@
     private final LocalDateTime endDate;
     private final List<String> hearingNeedsBooked;
     private final String hearingNeedsDetails;
-    private final String judgeTitle;
-    private final String judgeName;
+    private final JudgeAndLegalAdvisor judgeAndLegalAdvisor;
```
geek-soft: I am not fan of these `XandY` classes. They do too much, they have too many reasons to change. They are bad pattern, bad code smell.

Again not issue directly with this PR as you just apply existing solution in consistent manner.

e2e/tests/judiciaryAdministersCaseAfterSubmission_test.js
joez-f: agree - we do the same test multiple times for different user in different state etc, and we simply copy paste this test in different files. e2e tests are full of duplicate code 😸

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -80,12 +79,14 @@ public CaseDataExtractionService(DateFormatterService dateFormatterService,
...
-                caseData.getStandardDirectionOrder().getJudgeAndLegalAdvisor()));
-        } else {
-            data.put("judgeAndLegalAdvisor", prepareJudgeAndLegalAdvisor(JudgeAndLegalAdvisor.builder().build()));
-        }
+        JudgeAndLegalAdvisor judgeAndLegalAdvisor = isNotEmpty(caseData.getStandardDirectionOrder())
+            ? caseData.getStandardDirectionOrder().getJudgeAndLegalAdvisor() : caseData.getJudgeAndLegalAdvisor();
+
+        data.put("judgeTitleAndName", defaultIfBlank(JudgeAndLegalAdvisorHelper.formatJudgeTitleAndName(
+            judgeAndLegalAdvisor), EMPTY_PLACEHOLDER));
+        //Legal advisor will be tied to magistrate in following story - this + tests will need updating then
```
joez-f: actually this comment is incorrect now after discussions with BA - legal advisor now optional and not conditional on magistrate. following story still addresses this though 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
```diff
@@ -25,6 +26,10 @@ public static String formatJudgeTitleAndName(JudgeAndLegalAdvisor judgeAndLegalA
     }
 
     private static String mapJudgeOrAdvisor(JudgeAndLegalAdvisor judgeAndLegalAdvisor) {
+        if (isBlank(judgeAndLegalAdvisor.getJudgeLastName()) && isBlank(judgeAndLegalAdvisor.getJudgeFullName())) {
```
geek-soft: I could not find it in the story, is that something business wanted? It hides judge title if name is missing, is that alright?

If this is OK please update tests.

e2e/tests/judiciaryAdministersCaseAfterSubmission_test.js
geek-soft: @akomisarek, you might look into these duplicates in e2e tests. I question whether it makes sense to test same thing many times for each role eligible for action. Wouldn't it be enough to test it once to make sure UI it correct and then separately make sure that configuration has right permissions for roles or that roles inheritance is correct?

Not trivial problem but wanted to highlight it as I am not convinced with current approach.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -80,12 +79,14 @@ public CaseDataExtractionService(DateFormatterService dateFormatterService,
     public Map<String, Object> getStandardOrderDirectionData(CaseData caseData) throws IOException {
         ImmutableMap.Builder data = ImmutableMap.<String, Object>builder();
 
-        if (isNotEmpty(caseData.getStandardDirectionOrder())) {
-            data.put("judgeAndLegalAdvisor", prepareJudgeAndLegalAdvisor(
-                caseData.getStandardDirectionOrder().getJudgeAndLegalAdvisor()));
-        } else {
-            data.put("judgeAndLegalAdvisor", prepareJudgeAndLegalAdvisor(JudgeAndLegalAdvisor.builder().build()));
-        }
+        JudgeAndLegalAdvisor judgeAndLegalAdvisor = isNotEmpty(caseData.getStandardDirectionOrder())
```
geek-soft: Why not to use `caseData.getStandardDirectionOrder() != null`? Using `isNotEmpty` makes me wonder how it can be empty. Makes me wonder whether it is a list or a map. Mind that as a ready I have no type information at this point.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
joez-f: It may not be captured in story but it is what business wants (it may be noted in 1004, these stories are linked, it's a bit messy) - there is a situation where I draft SDO, then I redraft it later and remove a judge title. In this case the document will show the judge title with no name I.e "Her Honour Judge", but intended behaviour is that it should give the empty placeholder. If we have judge title we must have judge name. In following 1004 story, we then deal with magistrate situation. If we have magistrate title, then we do not need name and will display something different.

regardless yes I think this needs updated test, I will add 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
joez-f: can do, I used existing implementation for reference so stuck with isNotEmpty. Will change to != null, I like that better too

geek-soft: Someone made bad decision that was not verified :(

geek-soft:
Do we have any data in prod captured for hearing bookings in an old manner? Thinking whether we can loose it as removing fields from complex object will mean they will disappear from live cases.

joez-f:
I believe that there is no hearing data entered in live cases in prod , adam checked recently and there were none. SDO and C6 have been switched off, these are the events that make use of hearing details so I think that no one has entered hearing details yet @akomisarek

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -122,65 +122,36 @@ public CaseDataExtractionService(DateFormatterService dateFormatterService,
...
-                "judgeName", EMPTY_PLACEHOLDER
-            );
+            return ImmutableMap.<String, Object>builder()
+                .put("hearingDate", EMPTY_PLACEHOLDER)
+                .put("hearingVenue", EMPTY_PLACEHOLDER)
+                .put("preHearingAttendance", EMPTY_PLACEHOLDER)
+                .put("hearingTime", EMPTY_PLACEHOLDER)
+                .put("hearingJudgeTitleAndName", EMPTY_PLACEHOLDER)
+                .put("hearingLegalAdvisorName", "")
+                .build();
```
joez-f: ImmutableMap.of() limited to maximum 5 key-value pairs... now 6 so have made the refactor

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
joez-f: test added, please approve changes if happy

PR #654, opened 2019-11-28
feat: Add three new options for Judge titles (FPLA-1004)

akomisarek:
https://tools.hmcts.net/jira/browse/FPLA-1004

### Change description ###

Three new options for judge title:
* District Judge
* District Judge Magistrates Court
* Other (with additional description field)
Additionally, made the magistrate name optional and handled it nicely in the relevant templates. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
+++ b/e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
@@ -88,8 +88,9 @@ Scenario('Gatekeeper enters hearing details and submits', async (I, caseViewPage
   I.seeAnswerInTab(5, 'Hearing 2', '', hearingDetails[1].type.somethingElse);
   I.seeAnswerInTab(6, 'Hearing 2', 'Give details', hearingDetails[1].giveDetails);
   I.seeAnswerInTab(1, 'Judge and legal advisor', 'Judge or magistrate\'s title', hearingDetails[1].judgeAndLegalAdvisor.judgeTitle);
-  I.seeAnswerInTab(2, 'Judge and legal advisor', 'Last name', hearingDetails[1].judgeAndLegalAdvisor.judgeLastName);
-  I.seeAnswerInTab(3, 'Judge and legal advisor', 'Legal advisor\'s full name', hearingDetails[1].judgeAndLegalAdvisor.legalAdvisorName);
+  I.seeAnswerInTab(2, 'Judge and legal advisor', 'Title', hearingDetails[1].judgeAndLegalAdvisor.otherTitle);
+  I.seeAnswerInTab(3, 'Judge and legal advisor', 'Last name', hearingDetails[1].judgeAndLegalAdvisor.judgeLastName);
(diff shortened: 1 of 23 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
```diff
@@ -26,12 +27,16 @@ public static String formatJudgeTitleAndName(JudgeAndLegalAdvisor judgeAndLegalA
     }
 
     private static String mapJudgeOrAdvisor(JudgeAndLegalAdvisor judgeAndLegalAdvisor) {
-        if (isBlank(judgeAndLegalAdvisor.getJudgeLastName()) && isBlank(judgeAndLegalAdvisor.getJudgeFullName())) {
+        if (isBlank(judgeAndLegalAdvisor.getJudgeLastName())
+            && judgeAndLegalAdvisor.getJudgeTitle() != MAGISTRATES) {
             return "";
         }
 
         if (judgeAndLegalAdvisor.getJudgeTitle() == MAGISTRATES) {
```
lewisbirks: Would it be better to use a switch statement in this situation?

e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
```diff
@@ -88,8 +88,9 @@ Scenario('Gatekeeper enters hearing details and submits', async (I, caseViewPage
   I.seeAnswerInTab(5, 'Hearing 2', '', hearingDetails[1].type.somethingElse);
   I.seeAnswerInTab(6, 'Hearing 2', 'Give details', hearingDetails[1].giveDetails);
   I.seeAnswerInTab(1, 'Judge and legal advisor', 'Judge or magistrate\'s title', hearingDetails[1].judgeAndLegalAdvisor.judgeTitle);
```
joez-f: (0) If there was a way to set 'Judge or magistrate's title' in the tab to be otherTitle if we have one, that would be great, but I don't think it's possible - another option would be to hide this field if 'Other' is selected, but this would also happen during input. I don't think we can make this look any better in the tab, at least not right now ⏳

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
joez-f: yeah could use switch on .getJudgeTitle() for nicer code, but not too concerned as this story is fairly pressed for time + needs to be tested still. I'll let adam decide if he wants to refactor this part

e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
akomisarek: Yeah not right now :(,not much options we currently have.

PR #661, opened 2019-11-30
feat: docomsis document generation for CMO (FPLA-648)

Gogetter:
[FPLA-648](https://tools.hmcts.net/jira/browse/FPLA-648)

### Change description ###
Backend implementation for generating CMO case document

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonCaseDataExtractionService.java
```diff
@@ -53,6 +64,29 @@ public String extractPrehearingAttendance(HearingBooking booking) {
         return booking.hasDatesOnSameDay() ? formatTime(time) : formatDateTimeWithYear(time);
     }
 
+    public Map<String, Object> getHearingBookingData(final HearingBooking hearingBooking) {
+        if (isEmpty(hearingBooking)) {
```
geek-soft: I suggest doing simple `!= null` check which is more verbose.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonCaseDataExtractionService.java
```diff
@@ -53,6 +64,29 @@ public String extractPrehearingAttendance(HearingBooking booking) {
         return booking.hasDatesOnSameDay() ? formatTime(time) : formatDateTimeWithYear(time);
     }
 
+    public Map<String, Object> getHearingBookingData(final HearingBooking hearingBooking) {
```
geek-soft: May I ask you not to operate on maps? They are harder to refactor and encourage to bad code practices.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +207,143 @@ public void prepareCustomDirections(Map<String, Object> data) {
         }
     }
 
-    public Map<String, Object> generateCMOTemplateData(Map<String, Object> caseData) {
-        return new HashMap<>();
+    // REFACTOR: 02/12/2019 Refactor this with CaseDataExtractionService and NotifyOfProceedingService to try and
+    //  extract common elements to CommonCaseDataExtractionService (maybe an separate service for docmosis templates?)
+    @SuppressWarnings("unchecked")
+    public Map<String, Object> generateCMOTemplateData(Map<String, Object> caseDataMap) throws IOException {
```
geek-soft: Same here, do not return map.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +207,143 @@ public void prepareCustomDirections(Map<String, Object> data) {
...
+        });
+
+        return formattedDirections.build();
+    }
+
+    private String getFirstRespondentFullName(final CaseData caseData) {
+        return caseData.buildFirstRespondentFullName();
+    }
+
+    private Map<String, Object> getHearingBooking(final CaseData caseData, DynamicList hearingDateList) {
```
geek-soft: Be consistent, either you declare args as final or not.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -224,9 +376,8 @@ private void prePopulateHearingDateSelection(List<Element<HearingBooking>> heari
     }
 
     private String getRespondentFullName(RespondentParty respondentParty) {
-        String firstName = defaultIfNull(respondentParty.getFirstName(), "");
-        String lastName = defaultIfNull(respondentParty.getLastName(), "");
-
+        String firstName = defaultString(respondentParty.getFirstName());
+        String lastName = defaultString(respondentParty.getLastName());
         return String.format("%s %s", firstName, lastName);
```
geek-soft: Easier you added `buildFullName` to `Party` class. Use it here to remove duplicate.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/PopulateStandardDirectionsHandlerTest.java
```diff
@@ -101,7 +101,7 @@ void shouldPopulateStandardDirectionsWhenPopulatedDisplayInConfiguration() throw
                 .token(TOKEN)
                 .build());
 
-        given(ordersLookupService.getStandardDirectionOrder()).willReturn(OrderDefinition.builder()
+        given(ordersLookupService.getDirectionOrder()).willReturn(OrderDefinition.builder()
```
geek-soft: Why did you rename it?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -150,5 +143,19 @@ private final Respondent getFirstRespondent() {
     }
 
     private final CaseManagementOrder caseManagementOrder;
+
     private final Others others;
+
+    public String buildFirstRespondentFullName() {
```
geek-soft: I find this unnecessary and it does two things. I'd rather see composite based approach where that methods finds first respondent. Then `Party.buildFullName()` is called to format name on the outcome of first method.

Please split it.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftCMOController.java
```diff
@@ -67,9 +68,9 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
 
     @PostMapping("/mid-event")
     public AboutToStartOrSubmitCallbackResponse handleMidEvent(
-        @RequestBody CallbackRequest callbackRequest,
-        @RequestHeader("authorization") String authorization,
-        @RequestHeader("userId") String userId) {
+        @RequestHeader(value = "authorization") String authorization,
```
geek-soft: May I ask you to avoid unnecessary changes? They just make code review slower.

Reorder you made brings no value. Exposing attribute name (value) makes it unnecessarily verbose.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
```diff
@@ -21,4 +23,8 @@
     public final Address address;
     public final EmailAddress email;
     public final Telephone telephoneNumber;
+
+    public String buildFullName() {
```
geek-soft: Not sure what your experience but word `builder` does not work well for me. There is no building process, just simple concatenation. A `getFullName` works much better for me. Please rename it.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -107,16 +92,8 @@ public CaseDataExtractionService(DateFormatterService dateFormatterService,
...
-            try {
-                InputStream is = getClass().getResourceAsStream("/assets/images/draft-watermark.png");
-                fileContent = is.readAllBytes();
-                String encodedString = Base64.getEncoder().encodeToString(fileContent);
-
-                data.put("draftbackground", String.format("image:base64:%1$s", encodedString));
-            } catch (IOException e) {
-                log.warn(e.getMessage());
-            }
+            data.put("draftbackground", String.format("image:base64:%1$s",
```
geek-soft: Good idea to move it although I wonder what case data extraction (since name of the service) has to do with watermarks? Looks like bad naming or bad abstraction :(

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocmosisDocumentGeneratorService.java
```diff
@@ -57,4 +62,15 @@ public DocmosisDocument generateDocmosisDocument(Map<String, Object> templateDat
 
         return new DocmosisDocument(docmosisTemplate.getDocumentTitle(), response);
     }
+
+    public String generateDraftWatermarkEncodedString() {
```
geek-soft: I doubt it is good place for it. It adds unnecessary responsibility to the service that was focus on one concrete thing so far. For me there should be separate class that reads assets and encodes them in B64.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftCMOController.java
Gogetter: thanks @geek-soft. Reverted this as the order was initially incorrect and resulted in the attached error
error

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
Gogetter: your right @geek-soft `getFullname` will be preferred but given we are using Lombok, using `get` prefix will try to deserialize a `fullName` field which does not exist and will result in a `Unrecognized field "fullName"`.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/PopulateStandardDirectionsHandlerTest.java
Gogetter: we renamed to reflect reusability. As we needed to use for `CaseManagementOrder` order definitons

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocmosisDocumentGeneratorService.java
Gogetter: We initially had it in a different class but decided to bring in here with the idea that docmosis related stuff be in the same service. But your suggestion is right. We will bring the new service class

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
Gogetter: its currently used within the `getStandardOrderDirectionData` method (an existing method) and we just moved the code out to make more reusable instead of duplicating.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
Gogetter: we returned `Map` here as this is passed onto Docmosis render api

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/Party.java
lewisbirks: Rather than an issue with lombok, it more sounds like you are describing an issue with jackson seeing as lombok doesn't deal with serialisation and deserialisation

Gogetter: just thinking about it again, we can add `@JsonIgnore` to the `getFullName` method

swalker125:
what is the point of this? not clear from PR title, not clear from code changes, and I can't see anything to do with Judge picking options which is what the 648 story is related to?

lewisbirks:
> what is the point of this? not clear from PR title, not clear from code changes, and I can't see anything to do with Judge picking options which is what the 648 story is related to?

This is the document generation part of the story, also it is the LA picking options at this point

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -81,20 +98,36 @@ public DraftCMOService(DateFormatterService dateFormatterService,
 
     public CaseManagementOrder prepareCMO(Map<String, Object> caseData) {
         DynamicList list = mapper.convertValue(caseData.get("cmoHearingDateList"), DynamicList.class);
+
+        String hearingDate = null;
+        UUID id = null;
```
akomisarek: [-0.5] Probably that could be abstracted into Element<String> and use optional. maps instead of ugly nulls and ifs.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
akomisarek: [-1] I am OK (well not really but best what we can do right now) with map being returned.

However this class shouldn't work with the Map<String, Object> as input parameter we should work with our CaseData model immediately. I know you do conversion later, but I would much prefer strongly typed object being passed here.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +223,267 @@ public void prepareCustomDirections(Map<String, Object> data) {
...
+            "respondentName", respondent.getParty().getFullName(),
+            "representativeName", EMPTY_PLACEHOLDER,
+            "representativeEmail", EMPTY_PLACEHOLDER,
+            "representativePhoneNumber", EMPTY_PLACEHOLDER
+        );
+    }
+
+    private Map<String, Object> getSchedule(CaseManagementOrder caseManagementOrder) {
+        final Schedule schedule = caseManagementOrder.getSchedule();
+        Map<String, Object> scheduleMap = mapper.convertValue(schedule, new TypeReference<>() {});
```
akomisarek: Create variables and do operations only when they are required, this scheduleMap is not required in if statement.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +223,267 @@ public void prepareCustomDirections(Map<String, Object> data) {
...
+        Map<String, Object> result;
+
+        if ((schedule == null) || schedule.getIncludeSchedule().equals("No")) {
```
akomisarek: Would probably eextract to if(isScheduleIncluded(schedule)) would read a lot more nicely, or do some optional mapping.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +223,267 @@ public void prepareCustomDirections(Map<String, Object> data) {
...
+            result = getEmptyScheduleMap();
+        } else {
+            result = scheduleMap;
+        }
+
+        result.put("scheduleProvided", schedule != null && "Yes".equals(schedule.getIncludeSchedule()));
```
akomisarek: This small method is confusing, why we doing the check again, couldn't we do necessary operations in same place?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +223,267 @@ public void prepareCustomDirections(Map<String, Object> data) {
...
+
+    private String getCourtName(String localAuthorityCode) {
+        if (isBlank(localAuthorityCode)) {
+            return EMPTY_PLACEHOLDER;
+        }
+        return defaultIfBlank(hmctsCourtLookupConfiguration.getCourt(localAuthorityCode).getName(), EMPTY_PLACEHOLDER);
+    }
+
+    private Map<String, Object> getGroupedCMODirections(
+        final CaseManagementOrder caseManagementOrder) throws IOException {
```
akomisarek: This method is not throwing IOException.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +223,267 @@ public void prepareCustomDirections(Map<String, Object> data) {
         }
     }
 
-    public Map<String, Object> generateCMOTemplateData(Map<String, Object> caseData) {
```
akomisarek: I would probably create this piece of code in separate service, adding document generation template preparation in the same place as logic around CMO object handling baloons it to enormous size. What do you think?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -190,8 +223,267 @@ public void prepareCustomDirections(Map<String, Object> data) {
...
+        if (hearingDetails == null) {
+            return HearingBooking.builder().build();
+        }
+
+        return hearingDetails.stream()
+            .filter(element -> element.getId().equals(hearingDateList.getValue().getCode()))
+            .findFirst()
+            .map(Element::getValue)
+            .orElse(HearingBooking.builder().build());
+    }
```
akomisarek: THat could be extracted to HearingBookingService.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftCMOController.java
```diff
@@ -67,9 +68,9 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
 
     @PostMapping("/mid-event")
     public AboutToStartOrSubmitCallbackResponse handleMidEvent(
-        @RequestBody CallbackRequest callbackRequest,
-        @RequestHeader("authorization") String authorization,
-        @RequestHeader("userId") String userId) {
+        @RequestHeader("authorization") String authorization, @RequestHeader("user-id") String userId,
```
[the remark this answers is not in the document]
lewisbirks: According to @Gogetter it throws errors if we leave it in the order it was in before [see here](https://github.com/hmcts/fpl-ccd-configuration/pull/661#discussion_r352807789). The reason the order is actually changed is that in the base branch we made some changes based on a rough idea of what we were going to need but it was never tested because we then decided to do the changes for document generation in a new branch.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
```diff
@@ -377,4 +405,109 @@ public static Direction createUnassignedDirection() {
...
+            Element.<Recital>builder()
+                .value(Recital.builder()
+                    .title("A title")
+                    .description("A description")
+                    .build())
+                .build()
+        );
+    }
+
+    public static Schedule createSchedule(boolean includeSchedule) {
```
akomisarek: Maybe we could add this to separate helper (if we add all possible data creation here, it's going to grow in size a lot), and the whole helper thing should be moved to tests? Not in your pr but the place seems wrong.

service/src/test/java/uk/gov/hmcts/reform/fpl/model/common/PartyTest.java
```diff
@@ -0,0 +1,71 @@
...
+        builder.firstName("Bob");
+
+        final String fullName = builder.build().getFullName();
+
+        assertThat(fullName).isEqualTo("Bob");
+    }
+
+
+    @Test
+    void shouldReturnFirstNameOnlyWhenLastNameIsBlank() {
```
akomisarek: Really nice scenarios covered here. 👍

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
```diff
@@ -32,8 +34,10 @@
 import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createStandardDirectionOrders;
 
 @ExtendWith(SpringExtension.class)
-@ContextConfiguration(classes = {JacksonAutoConfiguration.class, JsonOrdersLookupService.class,
-    HearingVenueLookUpService.class})
+@ContextConfiguration(classes = {
+    JacksonAutoConfiguration.class, JsonOrdersLookupService.class,
+    HearingVenueLookUpService.class, DocmosisDocumentGeneratorService.class, RestTemplate.class,
```
akomisarek: [-1] What Damian commented on previously if you need to add resttepmlate to get the watermark as this is puling that dependency, I don't like it :(

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
lewisbirks: I think there is a PR to already do this

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
lewisbirks: That would probably be a nice idea

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
lewisbirks: Yeah will extract it to a method

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
lewisbirks: Will move to after

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonCaseDataExtractionService.java
lewisbirks: The one in `CaseDataExtractionService` returns a map which also contains information about the judge and legal advisor where as this one doesn't

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
lewisbirks: Moved to a new abstract class that future docmosis template data providers can then extend

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocmosisDocumentGeneratorService.java
lewisbirks: Moved to a new abstract class that future docmosis template data providers can then extend

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
lewisbirks: However I have left it here for CaseDataExtraction service so as to not change to much, future refactor

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocmosisDocumentGeneratorService.java
```diff
@@ -57,4 +62,15 @@ public DocmosisDocument generateDocmosisDocument(Map<String, Object> templateDat
...
+        InputStream is = getClass().getResourceAsStream("/assets/images/draft-watermark.png");
+        byte[] fileContent = new byte[0];
+        try {
+            fileContent = is.readAllBytes();
+        } catch (IOException e) {
+            log.error("Unable to generate draft water image for template.", e);
```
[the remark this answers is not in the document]
lewisbirks: Now throwing in new method

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
lewisbirks: Created a new set of classes to clean this up as discussed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
lewisbirks: Altered case data so that it can now be passed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
lewisbirks: final for everything 😄

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/PopulateStandardDirectionsHandlerTest.java
lewisbirks: Reverted change as it isn't actually needed for CMO

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
lewisbirks: Removed because data can be added in the previous if statement

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DocmosisTemplateDataGenerationTest.java
```diff
@@ -0,0 +1,37 @@
...
+        @Override
+        public Map<String, Object> getTemplateData(CaseData caseData) {
+            throw new NotImplementedException("Not testing this method");
+        }
+    };
+
+    @Test
+    void shouldGenerateEncodedStringWhenGenerateDraftWatermarkCalled() throws IOException {
+        String encodedImgString = generateDraftWatermarkEncodedString();
+        assertThat(encodedImgString).isNotEqualTo("");
```
akomisarek: Do we need ot have this test if we have the test below? isn't it duplicating?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -149,6 +143,12 @@ private final Respondent getFirstRespondent() {
         return defaultIfNull(c21Orders, new ArrayList<>());
     }
 
+    // Objects related to CMO flow
     private final CaseManagementOrder caseManagementOrder;
+    private final DynamicList cmoHearingDateList;
+    private final Schedule schedule;
+    private final List<Element<Recital>> recitals;
+    private final ReviewCaseManagementOrder reviewCaseManagementOrder;
```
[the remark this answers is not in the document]
lewisbirks: When going through the event the captured data is stored in individual objects before being combined into the cmo at the end. This review object captures the cmoStatus and orderDoc, the latter is being populated in the mid event to display on the last page.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocmosisTemplateDataGeneration.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    protected Map<String, Object> getDraftWaterMarkData() throws IOException {
+        return ImmutableMap.of(
+            "draftbackground", String.format("image:base64:%1$s", generateDraftWatermarkEncodedString())
+        );
+    }
+
+    // QUESTION: 05/12/2019 Should there be some sort of contract to check if the template data will be for a draft
+    //  document? Maybe a boolean parameter or a method of some sort?
```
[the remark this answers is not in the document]
lewisbirks: Only adding the watermark. Could make the `getDraftWaterMarkData` method public and then we don't need to worry about it as that method can then always be called and added to the result of the `getTemplateData` method

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DocmosisTemplateDataGenerationTest.java
lewisbirks: Good point, they are basically doing the same thing, will removed the former and leave the latter

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonCaseDataExtractionService.java
lewisbirks: In our situation it would have overridden existing entires in the map with incorrect data

service/src/main/java/uk/gov/hmcts/reform/fpl/model/configuration/ReviewCaseManagementOrder.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.reform.fpl.model.configuration;
+
+import lombok.Builder;
+import lombok.Data;
+import uk.gov.hmcts.reform.fpl.enums.CMOStatus;
+import uk.gov.hmcts.reform.fpl.model.common.DocumentReference;
+
+@Data
+@Builder
+public class ReviewCaseManagementOrder {
```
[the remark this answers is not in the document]
lewisbirks: Where would you prefer it to be seeing as it needs to exist in the current implementation? Couldn't really think of a suitable place for it

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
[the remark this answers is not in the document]
lewisbirks: Removed the requirement for `ReviewCaseManagementOrder`

service/src/main/java/uk/gov/hmcts/reform/fpl/model/configuration/ReviewCaseManagementOrder.java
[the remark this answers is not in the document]
lewisbirks: Removed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocmosisTemplateDataGeneration.java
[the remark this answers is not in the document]
akomisarek: You meant private? then we wouldn't worry :), but that's fine something to consider in the future, is in good enough state.

PR #684, opened 2019-12-06
Make case submission restriction configurable

geek-soft:

https://tools.hmcts.net/jira/browse/FPLA-1047

### Change description ###

☝️ 

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/main/resources/application.yaml
+++ b/service/src/main/resources/application.yaml
@@ -17,6 +17,9 @@ idam:
   s2s-auth:
     microservice: fpl_case_service
 
+fpl:
+  local_authority_codes_forbidden_case_submission:
+
 ---
 
 spring:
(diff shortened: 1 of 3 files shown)


service/src/main/resources/application.yaml
```diff
@@ -57,6 +60,7 @@ ccd:
       url: http://fake-url
 
 fpl:
+  local_authority_codes_forbidden_case_submission: FPLA
```
akomisarek: Probably it works. Suprised you don't need - to make it a list.

geek-soft: Since it is declared as `List<String>` it does split on `,` with some whitespace trimming.

akomisarek: Oh, that's clever. Thanks I should check the docs!

PR #691, opened 2019-12-10
feat:  automatically share a case with org (FPLA-673)

akomisarek:

https://tools.hmcts.net/jira/browse/FPLA-673

### Change description ###

Adds functionality to share a case automatically within an organisation. Current implementation is as follows:

1. Check the users from the same localAuthority using existing hardcoded mapping.
2. If the LA is not configured for specified code, check the RD (Reference Data) Professional API for all the user within same organisation. 
3. Adds current's userId to the returned list (if missing) to ensure the code LASOLICITOR role is appropriately configured for user creating a case.

```
[ ] Yes
[X] No
```


bin/configurer/add-roles.sh
```diff
@@ -11,15 +11,9 @@ ${dir}/utils/idam-add-role.sh "caseworker-publiclaw"
...
-${dir}/utils/ccd-add-role.sh "caseworker-publiclaw-courtadmin"
-${dir}/utils/idam-add-role.sh "caseworker-publiclaw-cafcass"
-${dir}/utils/ccd-add-role.sh "caseworker-publiclaw-cafcass"
-${dir}/utils/idam-add-role.sh "caseworker-publiclaw-gatekeeper"
-${dir}/utils/ccd-add-role.sh "caseworker-publiclaw-gatekeeper"
-${dir}/utils/idam-add-role.sh "caseworker-publiclaw-systemupdate"
-${dir}/utils/ccd-add-role.sh "caseworker-publiclaw-systemupdate"
-${dir}/utils/idam-add-role.sh "caseworker-publiclaw-judiciary"
-${dir}/utils/ccd-add-role.sh "caseworker-publiclaw-judiciary"
+roles=("-solicitor" "-courtadmin" "-cafcass" "-gatekeeper" "-systemupdate" "-judiciary")
```
geek-soft: For me it would look better if role separating dash was removed from here and kept in lines 17-18.

README.md
```diff
@@ -37,16 +37,14 @@ e.g. [direnv](https://direnv.net).
...
-
 ```bash
 $ ./bin/configurer/add-services.sh
 $ ./bin/configurer/add-roles.sh
 $ ./bin/configurer/add-users.sh
-$ ./bin/configurer/utils/generate-local-user-mappings.sh
-
 ```
 
+Users are defined in `bin/configurer/users.json`. Run the final script each time new LA users are added in order to fix access segregation locally.
```
geek-soft: What do you mean by final script? You removed `./bin/configurer/utils/generate-local-user-mappings.sh` from list of scripts above.

geek-soft: Alright, I saw you moved it into `./bin/configurer/add-users.sh`

bin/configurer/add-users.sh
```diff
@@ -4,17 +4,9 @@ set -eu
...
-${dir}/utils/idam-create-caseworker.sh "sam@wiltshire.gov.uk" "caseworker,caseworker-publiclaw,caseworker-publiclaw-solicitor" "(local-authority)"
-${dir}/utils/idam-create-caseworker.sh "hmcts-admin@example.com" "caseworker,caseworker-publiclaw,caseworker-publiclaw-courtadmin" "(hmcts-admin)"
-${dir}/utils/idam-create-caseworker.sh "cafcass@example.com" "caseworker,caseworker-publiclaw,caseworker-publiclaw-cafcass" "(cafcass)"
-${dir}/utils/idam-create-caseworker.sh "judiciary@mailnesia.com" ",caseworker-publiclaw-courtadmin, caseworker,caseworker-publiclaw,caseworker-publiclaw-judiciary,caseworker-publiclaw-gatekeeper" "(judiciary)"
-${dir}/utils/idam-create-caseworker.sh "gatekeeper@mailnesia.com" "caseworker,caseworker-publiclaw,caseworker-publiclaw-gatekeeper,caseworker-publiclaw-courtadmin" "(gatekeeper)"
-${dir}/utils/idam-create-caseworker.sh "fpl-system-update@mailnesia.com" "caseworker,caseworker-publiclaw,caseworker-publiclaw-systemupdate" "(system-update)"
+jq -r '.[] | .email + " " + .roles + " " +  .lastName' ${dir}/users.json | while read args; do
```
geek-soft: Might be useful to mention in README that `jq` has to be installed (in prerequisites for example). Or maybe use dockerised jq.

bin/configurer/users.json
```diff
@@ -0,0 +1,16 @@
...
+  {"email": "raghu@swindon.gov.uk", "roles": "caseworker,caseworker-publiclaw,caseworker-publiclaw-solicitor", "lastName": "(local-authority)"},
+  {"email": "sam@swindon.gov.uk", "roles": "caseworker,caseworker-publiclaw,caseworker-publiclaw-solicitor", "lastName": "(local-authority)"},
+  {"email": "raghu@wiltshire.gov.uk", "roles": "caseworker,caseworker-publiclaw,caseworker-publiclaw-solicitor", "lastName": "(local-authority)"},
+  {"email": "sam@wiltshire.gov.uk", "roles": "caseworker,caseworker-publiclaw,caseworker-publiclaw-solicitor", "lastName": "(local-authority)"},
+  {"email": "hmcts-admin@example.com", "roles": "caseworker,caseworker-publiclaw,caseworker-publiclaw-courtadmin", "lastName": "(hmcts-admin)"},
+  {"email": "cafcass@example.com", "roles": "caseworker,caseworker-publiclaw,caseworker-publiclaw-cafcass", "lastName": "(cafcass)"},
+  {"email": "judiciary@mailnesia.com", "roles": "caseworker-publiclaw-courtadmin,caseworker,caseworker-publiclaw,caseworker-publiclaw-judiciary,caseworker-publiclaw-gatekeeper", "lastName": "(judiciary)"},
```
geek-soft: List of roles is bit messy here, it starts with more concrete role, then there are more generic roles and then more specific roles again. Every time I need to look into that it gives me headache. Might be an opportunity to clean that up.

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -4,22 +4,50 @@ set -eu
...
+
+function get_users_ids() {
+  query_db "SELECT string_agg(fullobject->>'_id',',') FROM managedObjects WHERE fullobject->>'userName' LIKE '%$1%';"
+}
+
+function create_mock_response() {
+  jq --arg organisation $1 '[.[] | select(.email | contains($organisation))]' $users_file > $mock_file
+  echo $(get_users_email_id_mappings $1) > $users_ids_tmp_file
+
+  for i in `seq 1 $(jq '. | length' $mock_file)`
```
geek-soft: Wound't it be easier to start with 0 and not having to do '$index - 1' later?

bin/configurer/utils/generate-local-user-mappings.sh
```diff
@@ -4,22 +4,50 @@ set -eu
...
+  do
+    email=$(jq -r --argjson index $i '.[$index-1].email' $mock_file)
+    user_id=$(jq -r --arg email $email '.[$email]' $users_ids_tmp_file)
+    jq -r --argjson index $i --arg user_id $user_id '.[$index-1].userIdentifier=$user_id | .[$index-1].firstName=.[$index-1].email | .[$index-1].idamStatus="ACTIVE" | .[$index-1].idamStatusCode="0" | .[$index-1].idamMessage="" | .[$index-1].roles|=split(",")' $mock_file > $mock_tmp_file && mv $mock_tmp_file $mock_file
```
geek-soft: Not a big fan of this solution. First it is not clear what the end result is cause of all these pipes. There is very strong coupling between input list of users ('users.json') and output mock response. That makes it less readable and unnecessarily more fragile.

docker/wiremock/__files/mockResponse.example.json
```diff
@@ -0,0 +1,16 @@
+[
```
geek-soft: Is there any use for that file? I'd easily delete it seeing it is never used. Ideally if script that generates mock response used something like that as a template. That would make it way clearer than modifying JSON objects through 'jq' pipes.

service/src/main/java/uk/gov/hmcts/reform/fpl/config/FeignConfiguration.java
```diff
@@ -12,4 +16,9 @@ public Retryer retryer() {
         return new Retryer.Default(200, 2000, 5);
     }
 
+    @Bean
+    @Primary
+    Decoder feignDecoder(ObjectMapper objectMapper) {
```
geek-soft: Out of curiosity why is that needed now?

service/src/main/java/uk/gov/hmcts/reform/rd/client/OrganisationApi.java
```diff
@@ -0,0 +1,24 @@
...
+import uk.gov.hmcts.reform.fpl.config.FeignConfiguration;
+import uk.gov.hmcts.reform.rd.model.Status;
+import uk.gov.hmcts.reform.rd.model.User;
+
+import java.util.List;
+
+import static org.springframework.http.HttpHeaders.AUTHORIZATION;
+import static uk.gov.hmcts.reform.ccd.client.CoreCaseDataApi.SERVICE_AUTHORIZATION;
+
+@FeignClient(name = "rd-professional-api", url = "${rd_professional.api.url}", configuration = FeignConfiguration.class)
```
geek-soft: If not tight deadline I would have asked you to move it to separate library so that everyone can reuse it.

service/src/main/java/uk/gov/hmcts/reform/rd/client/OrganisationApi.java
```diff
@@ -0,0 +1,24 @@
...
+public interface OrganisationApi {
+    @GetMapping("/refdata/external/v1/organisations/users")
+    List<User> findUsersByOrganisation(
+        @RequestHeader(AUTHORIZATION) String authorisation,
+        @RequestHeader(SERVICE_AUTHORIZATION) String serviceAuthorization,
+        @RequestParam(value = "status") final Status status
```
geek-soft: Make it consistent, either use final keyword or not.

service/src/main/java/uk/gov/hmcts/reform/rd/model/User.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class User {
+    private String userIdentifier;
```
geek-soft: Do you need setter for that class?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+
+@Service
+@Slf4j
+public class OrganisationService {
+    private final LocalAuthorityUserLookupConfiguration localAuthorityUserLookupConfiguration;
+    private final OrganisationApi organisationApi;
+    private final AuthTokenGenerator authTokenGenerator;
+
+    @Autowired
```
geek-soft: Consider `@RequiredArgsConstructor(onConstructor = @__(@Autowired))` which is quite neat.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -0,0 +1,66 @@
...
+    public List<String> findUserIdsInSameOrganisation(String authorisation, String userId, String localAuthorityCode) {
+        List<String> userIds = new ArrayList<>();
+
+        try {
+            addUsersFromSameOrganisationBasedOnAppConfig(localAuthorityCode, userIds);
+        } catch (UnknownLocalAuthorityCodeException ex) {
+            try {
+                addUsersFromSameOrganisationBasedOnReferenceData(authorisation, userIds);
+            } catch (Exception e) {
+                log.warn("Can't find LocalAuthority for code: "
```
geek-soft: Use logging system support for interpolation  "... {} ..." instead of doing manual string concatenation.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -0,0 +1,66 @@
...
+    public OrganisationService(LocalAuthorityUserLookupConfiguration localAuthorityUserLookupConfiguration,
+                               OrganisationApi organisationApi,
+                               AuthTokenGenerator authTokenGenerator) {
+        this.localAuthorityUserLookupConfiguration = localAuthorityUserLookupConfiguration;
+        this.organisationApi = organisationApi;
+        this.authTokenGenerator = authTokenGenerator;
+    }
+
+
+    public List<String> findUserIdsInSameOrganisation(String authorisation, String userId, String localAuthorityCode) {
```
geek-soft: Not big fan of passing user ID only to add it to list if certain condition is not met. It also add extra responsibility to `findUsers..` method that should be searching for users not having extra responsibilities. I'd rather have handling an empty list scenario outside.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -0,0 +1,66 @@
...
+                    + localAuthorityCode
+                    + " in app config and the PRD endpoint call threw an exception.", e);
+            }
+        }
+
+        if (!userIds.contains(userId)) {
+            userIds.add(userId);
+        }
+
+        return userIds;
```
geek-soft: I am not fan of methods that returns mutable lists. Using list builder and returning immutable list might be clearer.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private void addUsersFromSameOrganisationBasedOnAppConfig(String localAuthorityCode, List<String> userIds) {
```
geek-soft: Not big fan of methods like that. IMO better and cleaner is to append users in main method with smaller methods that return a list to be appended e.g.

```
private void getUsersFromSameOrganisationBasedOnAppConfig(String localAuthorityCode) { // if you need to rename it for more clarity
  return localAuthorityUserLookupConfiguration.getUserIds(localAuthorityCode)
}

private void getUsersFromSameOrganisationBasedOnReferenceData(String authorisation) {
         return organisationApi
             .findUsersByOrganisation(authorisation, authTokenGenerator.generate(), Status.ACTIVE)
             .stream()
             .map(User::getUserIdentifier)
     }
 }
```

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -0,0 +1,66 @@
...
+        this.authTokenGenerator = authTokenGenerator;
+    }
+
+
+    public List<String> findUserIdsInSameOrganisation(String authorisation, String userId, String localAuthorityCode) {
+        List<String> userIds = new ArrayList<>();
+
+        try {
+            addUsersFromSameOrganisationBasedOnAppConfig(localAuthorityCode, userIds);
+        } catch (UnknownLocalAuthorityCodeException ex) {
```
geek-soft: So you took one or another approach? How will you test it in ITHC env. Will you remove all mapping for FPLA LA and populate PRD database with it?

service/src/main/java/uk/gov/hmcts/reform/fpl/config/FeignConfiguration.java
akomisarek: To be honest I was not sure why we even needed FeignConfiguration - all the clients we are using (Idam and CaseAccess) is specyfing their own configuration, so I am not sure if it ever worked? Or was it adding the extra retrying functionality to the existing clients? Would need to Debug.

My part is needed as by default it doesn't cope with Json response properly, I would like to explore a bit more as I don't like the NoArgsConstructor still needed, but otherwise it didn't work for me.

bin/configurer/add-roles.sh
pd-dudkiewicz: good point, changed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: Thanks 👍. Implemented

bin/configurer/add-users.sh
pd-dudkiewicz: updated README

bin/configurer/users.json
pd-dudkiewicz: done

bin/configurer/utils/generate-local-user-mappings.sh
pd-dudkiewicz: agree :+1: changed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: I was planning to create new organisation and test the journey e2e for new organisation from scratch (we would still need our mails to be configured but only that).

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: Ok that's good point. I will refactor.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: Can I do both exception and placeholders? The SLF4j log API, documentation informs me the arguments I am passing are just objects for placeholders, which makes me nervous.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: 👍 will refactor.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: Agree, will reefactor.

service/src/main/java/uk/gov/hmcts/reform/rd/model/User.java
akomisarek: Yes, please see my comment in the FeignConfiguration, if time better would investigate a bit more.

bin/configurer/utils/generate-local-user-mappings.sh
pd-dudkiewicz: Added template file used to fill mock response and therefore reduced number of pipes on this line. But this needs to be strongly coupled with _users.json_ as the idea was to have single place for users definition that's used in all user-related scripts.

docker/wiremock/__files/mockResponse.example.json
pd-dudkiewicz: Just put it to show how's _mockResponse.json_ supposed to look like. Removed and added _userTemplate.json_ as suggested.

service/src/main/java/uk/gov/hmcts/reform/rd/client/OrganisationApi.java
akomisarek: I agree, personally I even think the owning teams could provide the clients, but definitely it should be shared.

service/src/main/java/uk/gov/hmcts/reform/rd/client/OrganisationApi.java
akomisarek: 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
geek-soft: From my experience you can. Exception should be last arg.

bin/configurer/utils/generate-local-user-mappings.sh
geek-soft: I understand need for having source list of users, what I do not like is that that list, that is used in two or three different places is very strongly coupled with structure of the API that we mock. In that sense we have no freedom of adding anything to elements of the user list because if we do we will smuggle extra stuff to mocked responses. Same way unlikely change in PRD API structure might force us make change in two or three places that are completely unrelated.

In ideal world list of users should have its own structure and shell script should make that structure to structure of PRD API.

Cause of tight deadline I am not asking more for changes. Happy to accept as is ;)

geek-soft: Here is 🥇 for you @pd-dudkiewicz for becoming our JQ expert :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: Ok thanks, I have refactored the code so this is no longer relevant, but will remember for the future.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: Returning ImmutableList now.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
akomisarek: Refactored.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityUserService.java
```diff
@@ -1,71 +1,75 @@
...
-        List<String> userIds = localAuthorityUserLookupConfiguration.getUserIds(localAuthorityCode);
-
-        if (userIds.isEmpty()) {
-            throw new NoAssociatedUsersException("No users found for the local authority '" + localAuthorityCode + "'");
+    private List<String> findUserIds(String authorisation, String localAuthorityCode) {
+        try {
+            return organisationService
+                .findUserIdsInSameOrganisation(authorisation, localAuthorityCode);
+        } catch (Exception e) {
+            log.warn("Exception while looking for users within the same LA. "
```
geek-soft: I'd say not quite right place to say who gets access but no time to make it better.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -0,0 +1,58 @@
...
+            return ImmutableList
+                .copyOf(getUsersFromSameOrganisationBasedOnAppConfig(localAuthorityCode));
+        } catch (UnknownLocalAuthorityCodeException ex) {
+            try {
+                return
+                    ImmutableList
+                        .copyOf(getUsersFromSameOrganisationBasedOnReferenceData(authorisation));
+            } catch (Exception e) {
+                throw new UserOrganisationLookupException(
+                    format("Can't find users for %s local authority", localAuthorityCode), e
```
geek-soft: Not quite what I asked but template is usually better than concatenation. I am OK with it.

PR #693, opened 2019-12-10
refactor: added nested class for holding methods common event data

pd-dudkiewicz:
### Change description ###
Added nested class in NotificationHandler to hold common event data - avoiding code duplication. Minor changes in other files.

```
[ ] Yes
[x] No
```


toireasam:
LGTM i like the refactoring of the notification handler :)

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -175,4 +168,17 @@ private void sendOrderNotificationForLocalAuthority(final CaseDetails caseDetail
         sendNotification(ORDER_NOTIFICATION_TEMPLATE, localAuthorityEmail, localAuthorityParameters,
             Long.toString(caseDetails.getId()));
     }
+
+    @Getter
+    private static class EventData {
+        private CaseDetails caseDetails;
```
geek-soft: I do not quite get the purpose if having `caseDetails` (whole forest) and `localAuthorityCode` (single tree) or `reference` (another tree) in one class.

Much bigger problem I see is that still in many places data is retrieved using property names e.g. `.getData().get(CASE_LOCAL_AUTHORITY_PROPERTY_NAME)`.

pd-dudkiewicz: The purpose behind having these values in one class was to avoid duplicating the same 3 lines in almost every `NotificationHandler`'s method. I see your point, it may not make sense logically to combine these values, but I believe it's always easier to see what method is actually doing if you don't have to filter out the noise.

Noticed one more method calling `.getData().get(CASE_LOCAL_AUTHORITY_PROPERTY_NAME)`, will get that fixed.

lewisbirks: Another alternative then would be to only store the caseDetails object, and have the get methods perform the same extraction that the constructor does.

pd-dudkiewicz: @lewisbirks that'd make sense as get methods are used once in every EventListener, but I'd rather keep it a little bit shorter with lombok-generated getters.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocumentGeneratorService.java
```diff
@@ -34,7 +34,7 @@ public DocumentGeneratorService(HTMLToPDFConverter converter,
     @SafeVarargs
     public final byte[] generateSubmittedFormPDF(CaseDetails caseDetails, Map.Entry<String, ?>... extraContextEntries) {
         Map<String, Object> context = mapper.convertValue(populateEmptyCollections(caseDetails),
-            new TypeReference<Map<String, Object>>() {});
+            new TypeReference<>() {});
```
Gogetter: 👍

Gogetter:
LGTM going with the idea of removing the duplicates like
```java
CaseDetails caseDetails = ...;
String localAuthorityCode = ...;
String reference = ...;
```
in `NotificationHandler` and wrapping these within `EventData`.

PR #698, opened 2019-12-12
fix: adapt to real PRD model for findUsersByOrg

akomisarek:
### Change description ###

This PR fixes the incorrect PRD model which the client expected.

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/rd/model/Users.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/rd/model/Users.java
@@ -1,0 +1,16 @@
+package uk.gov.hmcts.reform.rd.model;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+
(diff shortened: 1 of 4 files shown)


service/src/main/java/uk/gov/hmcts/reform/rd/model/Users.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class Users {
```
geek-soft: Not a great name. Checking how PRD named it.

geek-soft: `ProfessionalUsersEntityResponse` blah...

geek-soft: I do prefer adding `Response` suffix in cases like that. I also wonder if there is a way to flatten it at Feign level.

PR #699, opened 2019-12-12
refactor: added enum for case management order keys

swalker125:
added enum class for case management order keys.


service/src/main/java/uk/gov/hmcts/reform/fpl/service/CMODocmosisTemplateDataGenerationService.java
```diff
@@ -115,7 +116,7 @@
         }
 
         List<Map<String, String>> recitals = buildRecitals(order.getRecitals());
-        cmoTemplateData.put("recitals", recitals);
+        cmoTemplateData.put(RECITALS.getKey(), recitals);
```
lewisbirks: It works here but I think we might need to do the same for docmosis keys as well, an enum class for each order generated possibly

PR #715, opened 2019-12-17
fix: allow only all party directions to be valid in cmo compliance (FPLA-NA)

swalker125:
### Change description ###

new method in DirectionHelperService to add empty values for DirectionAssignee not present.

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
@@ -20,6 +20,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import static java.util.Arrays.stream;
 import static java.util.Collections.emptyList;
 import static java.util.Objects.isNull;
 import static java.util.UUID.randomUUID;
@@ -402,6 +403,15 @@ public class DirectionHelperService {
         List<Element<Direction>> directions) {
         return directions.stream()
             .collect(groupingBy(directionElement -> directionElement.getValue().getAssignee()));
+    }
+
+    /**
+     * Adds any {@link DirectionAssignee} not present to a map with an empty list of directions.
+     *
+     * @param map assignee, directions key value pairs.
+     */
+    public void addEmptyDirectionsForAssigneeNotInMap(Map<DirectionAssignee, List<Element<Direction>>> map) {
+        stream(DirectionAssignee.values()).forEach(assignee -> map.putIfAbsent(assignee, new ArrayList<>()));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


service/src/test/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperServiceTest.java
```diff
@@ -752,6 +752,36 @@ void shouldSortDirectionsIntoSeparateEntriesInMapWhenManyAssignees() {
...
+    @Nested
+    class AddEmptyDirectionsForAssigneeNotInMap {
+
+        @Test
+        void shouldAddEmptyListValueWhenKeyNotPresentInMap() {
+            Map<DirectionAssignee, List<Element<Direction>>> map = new HashMap<>();
+
+            service.addEmptyDirectionsForAssigneeNotInMap(map);
+
+            assertThat(map).containsOnlyKeys(DirectionAssignee.values());
```
lewisbirks: You could loop over all the values and assert that each entry is an empty list, would make the test follow the name a bit closer

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -404,6 +405,15 @@ public void addResponseElementsToDirections(List<Element<DirectionResponse>> res
             .collect(groupingBy(directionElement -> directionElement.getValue().getAssignee()));
     }
 
+    /**
+     * Adds any {@link DirectionAssignee} not present to a map with an empty list of directions.
+     *
+     * @param map assignee, directions key value pairs.
+     */
+    public void addEmptyDirectionsForAssigneeNotInMap(Map<DirectionAssignee, List<Element<Direction>>> map) {
+        stream(DirectionAssignee.values()).forEach(assignee -> map.putIfAbsent(assignee, new ArrayList<>()));
```
lewisbirks: Was it not going to be `emptyList()` instead of `new ArrayList<>()`

swalker125: I get an issue in test where I am trying to add directions to an immutable list with empty list 😕

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperServiceTest.java
swalker125: I can add to the assert, sure

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
lewisbirks: Thought that might have been the issue

PR #717, opened 2019-12-18
feat: Add helm chart that serves prd mock (FPL-911)

tomaszpowroznik:

```
[ ] Yes
[X] No
```

### DIFF
--- a/charts/fpl-case-service/requirements.yaml
+++ b/charts/fpl-case-service/requirements.yaml
@@ -7,3 +7,7 @@ dependencies:
     repository: '@hmctspublic'
     tags:
       - fpl-ccd-stack
+  - name: fpl-prd-mock
+    version: 1.0.0
+    condition: tags.fpl-prd-mock
+    repository: "file://subcharts/fpl-prd-mock"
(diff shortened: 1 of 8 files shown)


charts/fpl-case-service/charts/fpl-prd-mock/values.yaml
```diff
@@ -0,0 +1,7 @@
+users:
+  - email: damian@swansea.gov.uk
+    identifier: 365465
+  - email: sam@hillingdon.gov.uk
+    identifier: c9166989-84e6-45fe-94c8-0148f908f35b
+prd:
+  url: http://rd-professional-api-aat.service.core-compute-aat.internal
```
akomisarek: Couldn't it use the value from the config somehow? What if we wanted to use the mock on different envs? How could we configure that?

tomaszpowroznik: done, move to parent chart values file for preview env

tomaszpowroznik: still users are aat specific, do you want them to be configured per env?

PR #726, opened 2019-12-19
fix: remove judge review cmo state (FPLA-NA)

swalker125:
### Change description ###

Removed JUDGE_REVIEW CMOstatus.

Replaced logic with eventId. 

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderProgressionService.java
```diff
@@ -34,29 +33,20 @@ public CaseManagementOrderProgressionService(ObjectMapper mapper) {
         this.mapper = mapper;
     }
 
-    public void handleCaseManagementOrderProgression(CaseDetails caseDetails) {
+    public void handleCaseManagementOrderProgression(CaseDetails caseDetails, String eventId) {
         CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
 
-        if (localAuthorityIsDrafting(caseData)) {
+        if (eventId.equals(DRAFT_CASE_MANAGEMENT_ORDER.getId())) {
```
lewisbirks: In the future I would flip this, won't be an issue in this instance but it is better to compare using the constant as a base when using .equals to avoid NPEs

PR #727, opened 2019-12-20
refactor/fix: update 'others to be given notice' event to be consistent with similar events

joez-f:
Will need QA:

- to check that 'others to be given notice' event still works as intended (for LA and admin, who can amend)
- to confirm CMO and SDO "others_label" is still correct

TODO:

- fix DOB not persisting (Others.java, 2_Others.json)
- integration tests for new Others controller (this controller is necessary for upcoming confidentiality story anyway)
- unit tests for new OthersService methods

### Change description ###

Update the 'Other to be given notice' event to just use a collection instead of a 'firstOther' followed by a collection. The old functionality was presumably before 'about-to-start' to expand collections was discovered. We now know how to do this, so we should be consistent with the other two similar events: 'Children' and 'Respondents'. Not only should we be consistent, but as the existing 'firstOther' is not part of a collection, it displays vastly different in the tab view, and users cannot actually delete the first other they have entered.

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
@@ -4,18 +4,23 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 
+import java.time.LocalDate;
+
 @Data
 @Builder
 @AllArgsConstructor
 public class Other {
     @SuppressWarnings("membername")
-    private final String DOB;
+    private final LocalDate DOB;        //date not persisting - please investigate
     private final String name;
     private final String gender;
     private final Address address;
     private final String telephone;
-    private final String birthplace;
+    private final String birthPlace;
     private final String childInformation;
+    private final String detailsHidden;
+    private final String detailsHiddenReason;
     private final String litigationIssues;
+    private final String litigationIssuesDetails;
     private final String genderIdentification;
 }
--- a/ccd-definition/AuthorisationCaseField/la-solicitor.json
+++ b/ccd-definition/AuthorisationCaseField/la-solicitor.json
@@ -221,7 +221,7 @@
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "others",
     "UserRole": "caseworker-publiclaw-solicitor",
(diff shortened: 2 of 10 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
```diff
@@ -4,18 +4,23 @@
...
 
+import java.time.LocalDate;
+
 @Data
 @Builder
 @AllArgsConstructor
 public class Other {
     @SuppressWarnings("membername")
-    private final String DOB;
+    private final LocalDate DOB;        //date not persisting - please investigate
```
swalker125: LocalDateTime ?

joez-f: but it's only a date in the event 😕 String didnt work, Date doesnt work, bit stumped
image

joez-f: `  {
    "LiveFrom": "01/01/2017",
    "ID": "Others",
    "ListElementCode": "DOB",
    "FieldType": "Date",
    "ElementLabel": "Date of birth",
    "HintText": "For example, 31 3 1980",
  },`

thats the ccd, and for reference we also have
`  {
    "LiveFrom": "01/01/2017",
    "ID": "ChildParty",
    "ListElementCode": "dateOfBirth",
    "FieldType": "Date",
    "ElementLabel": "Date of birth",
    "HintText": "For example, 31 3 2016",
  },`

swalker125:
so again, my problem with this is none of it is "needed" and it adds a lot more code to unpick for something where we should be doing data migration. Collections are largely not liked by PO and users, there was talk of asking us to change it back for respondents etc. I would discuss with Jay and Jason before proceeding any further

joez-f:
OK, we can handle 'others' separately if it requires further discussion with jay/jason. I wanted to implement confidentiality in a consistent way for the three events in question (children/respondents/others). can always leave out 'others' for now, or just do our best to implement it into the existing 'Others' (which might result in this refactor being impossible in future).

swalker125:
I would maybe suggest custom serialiser / deserialiser for this also

PR #744, opened 2019-12-30
feat: add directions to order generation flow (FPLA-1105)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1105

### Change description ###

New directions free text added to order flow. Directions appear in generated document but do not appear in tab view.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/FurtherDirections.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.hmcts.reform.fpl.model;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class FurtherDirections {
+    private final String directionsNeeded;
```
akomisarek: I know you started doing the JsonGetters / Setters could we model this as boolean and use the similar fashion to model it as boolean? :)

swalker125: do something like if directionsNeeded = yes, set, if no, set as empty string? definitely possible

swalker125: is it better practice to do optional logic on serialisation?

swalker125: @akomisarek

joez-f:
do we still want to merge 'use-IDAM-preview' as part of this PR? I believe IDAM AAT issues are resolved now - also we can open PR to https://github.com/hmcts/rdo-docmosis-templates development and approve it ourselves (so we can get this into QA stage), just not into prod

PR #747, opened 2019-12-31
imp: use IDAM on preview when AAT idam is down

akomisarek:

None

### Change description ###

This switches FPL preview environment to team IDAM instance that is deployed in AKS cluster. 

Team IDAM is deployed in preview and configured with all required services, roles and users.

In case switch is required the following steps are required:

1. merge this PR
2. run `bin/configurer/utils/generate-preview-user-mappings.sh` script and update `local-authority-user-mapping` Key Vault key with the result of the script to update user mappings
3. ~~set `fpl_case_service_secret` as value of `fpl-case-service-idam-client-secret` in Key Vault~~ (for now the code is using fpl-case-service-idam-client-secret-preview) seems a bit more elegant) 

```
[ ] Yes
[x] No
```

### DIFF
--- a/bin/configurer/utils/generate-preview-user-mappings.sh
+++ b/bin/configurer/utils/generate-preview-user-mappings.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -eu
+
+function get_user_ids() {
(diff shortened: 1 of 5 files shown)


akomisarek:
Not required for the time being - will resurrect when needed.

PR #748, opened 2019-12-31
feat: Inform LA that CMO is under review (FPLA-935)

tomaszpowroznik:

```
[ ] Yes
[X ] No
```

### DIFF
--- a/ccd-definition/AuthorisationCaseField/la-solicitor.json
+++ b/ccd-definition/AuthorisationCaseField/la-solicitor.json
@@ -544,21 +544,21 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "cmoDraftInReviewHeading",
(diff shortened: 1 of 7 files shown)


ccd-definition/AuthorisationCaseField/la-solicitor.json
```diff
@@ -539,5 +539,26 @@
     "CaseFieldID": "cmoEventId",
     "UserRole": "caseworker-publiclaw-solicitor",
     "CRUD": "CRU"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "cmoDraftInReviewHeading",
+    "UserRole": "caseworker-publiclaw-solicitor",
```
akomisarek: Could the permissions be appliedo n the Case role? [LASOLICITOR] I think it doesn't matter in this instance, but ideally we should use the case roles I think

tomaszpowroznik: done

PR #749, opened 2019-12-31
imp: add possibility to inject authorisation header to services via RequestData

akomisarek:
### Change description ###

As in title - you can Inject type RequestData to your services and as long as you are in the request (NOT ASYNC processing) you will get authorization header. 

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/request/AuthorizationConfiguration.java
```diff
@@ -0,0 +1,19 @@
...
+public class AuthorizationConfiguration {
+
+    @Bean
+    @RequestScope
+    public RequestData authorisation(HttpServletRequest request) {
+        String authorization = request.getHeader("authorization");
+
+        return new RequestData(authorization);
+    }
+}
```
tomaszpowroznik: request data does not have to be request scoped, by default injected HttpServletRequest is request scoped.
can be just:
@Service
public class RequestData {
    @Autowired
    HttpServletRequest httpServletRequest;

    public String authorsation() {
        return httpServletRequest.getHeader("authorization");
    }
}

service/src/main/java/uk/gov/hmcts/reform/fpl/request/AuthorizationConfiguration.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.web.context.annotation.RequestScope;
+
+import javax.servlet.http.HttpServletRequest;
+
+@Configuration
+public class AuthorizationConfiguration {
+
+    @Bean
+    @RequestScope
+    public RequestData authorisation(HttpServletRequest request) {
```
tomaszpowroznik: class and method is about authorisation, but returns quite generic RequestData

Gogetter: I think this might be expanded to include `user-id`

akomisarek: Refactored to what we discussed, thanks

service/src/main/java/uk/gov/hmcts/reform/fpl/request/AuthorizationConfiguration.java
akomisarek: Thanks, changing!

PR #752, opened 2020-01-02
fix: disable owasp checks temporarily

tomaszpowroznik:

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -135,6 +135,7 @@ dependencyUpdates.resolutionStrategy = {
 dependencyCheck {
   // Specifies if the build should be failed if a CVSS score above a specified level is identified.
   // range of 0-10 fails the build, anything greater and it doesn't fail the build
+  skip = true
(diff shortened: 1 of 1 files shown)


service/build.gradle
```diff
@@ -132,21 +132,21 @@ dependencyUpdates.resolutionStrategy = {
...
-    assemblyEnabled = false
-  }
-  analyzedTypes = ['jar']
-  scanConfigurations = configurations.findAll {
-    !it.name.startsWith('checkstyle') && !it.name.startsWith('jacoco') && !it.name.startsWith('integrationTest') &&
-      !it.name.startsWith('smokeTest') && !it.name.startsWith('functionalTest')
-  }.collect { it.name }
-}
+//dependencyCheck {
+//  // Specifies if the build should be failed if a CVSS score above a specified level is identified.
```
Gogetter: instead of commenting out, could we skip it?
`dependencyCheck {
  skip = true
...
}
`

tomaszpowroznik: perfect hint, thanks:)

PR #760, opened 2020-01-03
feat: add further directions to order flow (FPLA-1105)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1105

### Change description ###

Added directions text area to order generation flow. 

```
[ ] Yes
[x] No
```


ccd-definition/AuthorisationCaseField/court-admin.json
```diff
@@ -363,6 +363,13 @@
     "UserRole": "caseworker-publiclaw-courtadmin",
     "CRUD": "CRUD"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "orderFurtherDirections",
+    "UserRole": "caseworker-publiclaw-courtadmin",
```
joez-f: Need to update this for judiciary as well, as they can access this event

ccd-definition/CaseEventToFields/createOrder-PREPARE_FOR_HEARING.json
```diff
@@ -37,6 +37,19 @@
     "PageLabel": " ",
     "PageDisplayOrder": 3,
     "PageColumnNumber": 1,
+    "ShowSummaryChangeOption": "Y"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseEventID": "createOrder-PREPARE_FOR_HEARING",
```
joez-f: Need to update createOrderGatekeeping too

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -144,6 +144,8 @@ Scenario('HMCTS admin creates multiple orders for the case', async (I, caseViewP
   await createOrderEventPage.enterC21OrderDetails();
   await I.retryUntilExists(() => I.click('Continue'), '#judgeAndLegalAdvisor_judgeTitle');
   await createOrderEventPage.enterJudgeAndLegalAdvisor(orders[0].judgeAndLegalAdvisor.judgeLastName, orders[0].judgeAndLegalAdvisor.legalAdvisorName);
+  await I.retryUntilExists(() => I.click('Continue'), '#orderFurtherDirections_directionsNeeded');
```
joez-f: as mentioned above, judiciary have permission for Create an order event so need to update judiciaryAdministersCase.. e2e test as well with same changes

e2e/pages/events/createOrderEvent.page.js
```diff
@@ -7,6 +7,11 @@ module.exports = {
     title: '#order_title',
     details: '#order_details',
     orderTypeList: '#orderTypeAndDocument_type',
+    directionsNeeded: {
+      yes: '#orderFurtherDirections_directionsNeeded-Yes',
+      no: '#orderFurtherDirections_directionsNeeded-No',
```
toireasam: Why do we have directionsNeeded.no if we don’t use it here?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderServiceTest.java
```diff
@@ -261,6 +262,7 @@ private void assertCommonC21Fields(GeneratedOrder order) {
         }
 
         expectedMap
+            .put("furtherDirections", "Example Directions")
```
toireasam: we could probably move this expected map out to a private method - I know you are just adding one line here to existing code though so it's up to you :)

service/src/test/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderServiceTest.java
```diff
@@ -288,13 +290,21 @@ private CaseData createPopulatedCaseData(GeneratedOrderType type, LocalDate loca
                     .order(GeneratedOrder.builder()
                         .title("Example Title")
                         .details("Example details")
+                        .build())
```
toireasam: would these case builder options be nicer in private methods too or would you like to keep them in here? - getting bulky

toireasam:
few minor comments :)

e2e/pages/events/createOrderEvent.page.js
swalker125: I like the idea of the page objects being a bit like an API for the page. i.e this is everything you can do... But sure it's not used. You think I should remove?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderServiceTest.java
swalker125: it's already within a private method, so I am just going to leave as is

service/src/test/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderServiceTest.java
swalker125: also already in a private method

e2e/pages/events/createOrderEvent.page.js
toireasam: hmm, maybe that is a good idea then actually. i think leave in :)

ccd-definition/CaseEventToFields/createOrder.json
```diff
@@ -37,6 +37,19 @@
...
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseEventID": "createOrder",
+    "CaseFieldID": "orderFurtherDirections",
+    "PageFieldDisplayOrder": 1,
+    "DisplayContext": "OPTIONAL",
+    "PageID": "FurtherDirections",
+    "PageLabel": " ",
+    "PageDisplayOrder": 4,
+    "PageColumnNumber": 1,
```
joez-f: hmm actually if we have this with no PageShowCondition here it'll be there for every type of order. I don't think we want this page for Blank orders (C21) as there is already a page for C21 where users can enter free text details (2nd page). I think we want this additional directions page for other types of order except C21, so maybe a show condition to make sure the order isn't a BLANK_ORDER? will also have to update e2e; in e2e test first we make a C21, then we make a care order, so this page should only appear for the care order not the C21

lewisbirks: Think that it is now done as you have suggested

lewisbirks:
Reminder that the template needs to be updated in the docmosis repo

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -216,4 +204,85 @@ private CallbackRequest buildCallbackRequest() {
...
+            verify(uploadDocumentService, times(1)).uploadPDF(userId, userAuthToken, pdf, "care_order.pdf");
+
+            final CaseData caseData = mapper.convertValue(callbackResponse.getData(), CaseData.class);
+
+            assertThat(caseData.getOrderTypeAndDocument().getDocument()).isEqualTo(expectedDocument());
+        }
+
+        @Test
+        void midEventShouldNotGenerateOrderDocumentWhenOrderTypeIsCareOrderWithNoFurtherDirections() throws Exception {
+            postMidEvent(generateCallbackRequest(CARE_ORDER, false));
```
Jamesnelson117: postMidEvent accepts CaseDetails. We could do some refactoring here if you wish and remove the needed for CallBackRequest.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -216,4 +204,85 @@ private CallbackRequest buildCallbackRequest() {
...
+            verify(uploadDocumentService, never()).uploadPDF(any(), any(), any(), any());
+        }
+
+        @AfterEach
+        void tearDown() {
+            reset(docmosisDocumentGeneratorService);
+            reset(uploadDocumentService);
+        }
+
+        private CallbackRequest generateCallbackRequest(GeneratedOrderType orderType, boolean withFurtherDirections)
```
Jamesnelson117: I would almost prefer two separate methods with more specific naming as to what i'm getting returned rather than the use of a flag to augment data. But happy as it is.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderService.java
```diff
@@ -129,11 +131,16 @@ public OrderTypeAndDocument buildOrderTypeAndDocument(OrderTypeAndDocument typeA
             .put("legalAdvisorName", JudgeAndLegalAdvisorHelper.getLegalAdvisorName(
                 caseData.getJudgeAndLegalAdvisor()))
             .put("children", getChildrenDetails(caseData))
+            .put("furtherDirections", getFurtherDirections(caseData.getOrderFurtherDirections()))
             .build();
 
         return orderTemplateBuilder.build();
     }
 
+    private String getFurtherDirections(FurtherDirections furtherDirections) {
```
Jamesnelson117: Does it make sense to make this a public method on the furtherDirections object itself?

lewisbirks: Could do, would mean less bloat in the service

lewisbirks: Could call it something like `getFurtherDirectionsText`

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
lewisbirks: Can will change

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
lewisbirks: Changed to accept case details

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -216,4 +204,104 @@ private CallbackRequest buildCallbackRequest() {
...
+        void setUp() {
+            document = document();
+            DocmosisDocument docmosisDocument = new DocmosisDocument("order.pdf", pdf);
+
+            given(docmosisDocumentGeneratorService.generateDocmosisDocument(any(), any())).willReturn(docmosisDocument);
+            given(uploadDocumentService.uploadPDF(any(), any(), any(), any())).willReturn(document);
+        }
+
+        @Test
+        void midEventShouldGenerateOrderDocumentWhenOrderTypeIsBlankOrder() {
```
swalker125: I would maybe remove midEvent at the beginning of these tests. You get the context from the nested class :) what do you reckon? Won't reject but just a thought

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -216,4 +204,104 @@ private CallbackRequest buildCallbackRequest() {
...
+
+            verify(uploadDocumentService, times(1)).uploadPDF(userId, userAuthToken, pdf, "blank_order_c21.pdf");
+
+            assertThat(caseData.getOrderTypeAndDocument().getDocument()).isEqualTo(expectedDocument());
+        }
+
+        @Test
+        void midEventShouldGenerateOrderDocumentWhenOrderTypeIsCareOrderWithFurtherDirections() {
+            final AboutToStartOrSubmitCallbackResponse callbackResponse = postMidEvent(
+                generateCareOrderCaseDetails(true));
```
swalker125: generateCareOrderCaseDetails(true) makes me wanna be sick. What does passing true and false do here? To me it seems like true = I wanna generateCareOrderCaseDetails. If that's the case, I would just call the method. I am sure it works differently to this... but it should be clear so I can use the method without necessarily needing to read how it works

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -216,4 +204,104 @@ private CallbackRequest buildCallbackRequest() {
...
+
+            verify(docmosisDocumentGeneratorService, never()).generateDocmosisDocument(any(), any());
+            verify(uploadDocumentService, never()).uploadPDF(any(), any(), any(), any());
+        }
+
+        @AfterEach
+        void tearDown() {
+            reset(docmosisDocumentGeneratorService);
+            reset(uploadDocumentService);
+        }
```
swalker125: why do we need to tear down?

lewisbirks: A bug with Spring means that mock beans aren't reset after each run, as such this invocation count doesn't reset and the tests fail, this does what is supposed to happen but manually

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
lewisbirks: sure thing

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -231,4 +232,8 @@ private void setCaseManagementOrder_Judiciary(CaseManagementOrder order) {
         return isEmpty(getRespondents1()) || getRespondents1().size() <= seqNo
             ? empty() : Optional.of(getRespondents1().get(seqNo).getValue());
     }
+
+    public String getFurtherDirectionsText() {
+        return Optional.ofNullable(orderFurtherDirections).map(FurtherDirections::getDirections).orElse("");
+    }
```
swalker125: I think I would prefer something like defaultIfNull(orderFurtherDirections, orderFurtherDirections.builder().build())... thoughts?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderService.java
```diff
@@ -129,11 +129,13 @@ public OrderTypeAndDocument buildOrderTypeAndDocument(OrderTypeAndDocument typeA
             .put("legalAdvisorName", JudgeAndLegalAdvisorHelper.getLegalAdvisorName(
                 caseData.getJudgeAndLegalAdvisor()))
             .put("children", getChildrenDetails(caseData))
+            .put("furtherDirections", caseData.getFurtherDirectionsText())
             .build();
 
         return orderTemplateBuilder.build();
     }
 
+
```
swalker125: 👀

service/src/main/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderService.java
```diff
@@ -108,7 +108,7 @@ public OrderTypeAndDocument buildOrderTypeAndDocument(OrderTypeAndDocument typeA
                     .put("orderType", BLANK_ORDER)
                     .put("orderTitle", defaultIfNull(caseData.getOrder().getTitle(), "Order"))
                     .put("childrenAct", "Children Act 1989")
-                    .put("orderDetails", caseData.getOrder().getDetails());
+                    .put("orderDetails", defaultIfNull(caseData.getOrder().getDetails(), ""));
```
swalker125: are order details not mandatory? They might not be, can't rememeber

service/src/test/java/uk/gov/hmcts/reform/fpl/model/CaseDataTest.java
```diff
@@ -164,4 +165,34 @@ public void shouldNotFindNonExistingRespondent() {
 
         assertThat(caseData.findRespondent(1)).isEqualTo(Optional.empty());
     }
+
+    @Nested
+    class GetFurtherDirectionsText {
+
```
swalker125:
```suggestion
```

service/src/main/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderService.java
lewisbirks: Yeah they are, I think I did that due to a test will check

lewisbirks: Reverted

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
lewisbirks: The mapping captures the case of when `FurtherDirections::getDirections` is null as well so I personally prefer the Optional

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
lewisbirks: https://github.com/spring-projects/spring-boot/issues/12470
The provided solutions here didn't work for me either

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
lewisbirks: Updated to be less nauseous

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
swalker125: are directions mandatory? They probably should be if I select yes 🤔

swalker125: and then cool, can leave as is 👍

lewisbirks: Checked and it wasn't mandatory so made it so, left this as is though as I just want to get it into testing

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
lewisbirks: Renamed the method to be a bit more clear, tearDown was just the auto generated name

PR #762, opened 2020-01-03
feat: update application notification subject to indicate urgent hearing (FPLA-1114)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-1114

### Change description ###

Update the subject line for the two emails that are sent out when local authority submits an application (one to admins, one to Cafcass). Also updated the notification when admins send case to gatekeepers so that it is consistent with these changes. This is done in gov.notify. Subject line now should include information showing when the hearing is needed (and URGENT if same day hearing needed), and the respondent's surname.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
```diff
@@ -46,14 +49,18 @@ protected AbstractEmailContentProvider(String uiBaseUrl, DateFormatterService da
             .put("dataPresent", !ordersAndDirections.isEmpty() ? "Yes" : "No")
             .put("fullStop", !ordersAndDirections.isEmpty() ? "No" : "Yes")
             .put("timeFramePresent", timeFrame.isPresent() ? "Yes" : "No")
-            .put("timeFrameValue", timeFrame.orElse(""))
+            .put("timeFrameValue", uncapitalize(timeFrame.orElse("")))
+            .put("urgentHearing", timeFrame.isPresent() && timeFrame.get().equals("Same day") ? "Yes" : "No")
+            .put("nonUrgentHearing", timeFrame.isPresent() && !timeFrame.get().equals("Same day") ? "Yes" : "No")
+            .put("firstRespondentName", getFirstRespondentLastName(caseData))
```
Gogetter: would it be beneficial to move these three into a separate method? So we could have something like 
```java
Map<String, Object> getPartyPersonalisationBuilder(CaseData caseData) {
        Optional<String> timeFrame = Optional.ofNullable(caseData.getHearing().getTimeFrame());

        return ImmutableMap.<String, Object>builder()
            .put("urgentHearing", timeFrame.isPresent() && timeFrame.get().equals("Same day") ? "Yes" : "No")
            .put("nonUrgentHearing", timeFrame.isPresent() && !timeFrame.get().equals("Same day") ? "Yes" : "No")
            .put("firstRespondentName", getFirstRespondentLastName(caseData))
            .build();
    }
```

With this we do not need to update existing method and we have separation of concerns. Thoughts?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/CafcassEmailContentProvider.java
```diff
@@ -16,20 +18,25 @@
...
         this.localAuthorityNameLookupConfiguration = localAuthorityNameLookupConfiguration;
         this.cafcassLookupConfiguration = cafcassLookupConfiguration;
+        this.mapper = mapper;
     }
 
     public Map<String, Object> buildCafcassSubmissionNotification(CaseDetails caseDetails, String localAuthorityCode) {
-        return super.getCasePersonalisationBuilder(caseDetails)
+        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+
+        return super.getCasePersonalisationBuilder(caseDetails, caseData)
```
Gogetter: if you agree its beneficial from above, we could then use as 
```java
.putAll(getPartyPersonalisationBuilder(caseData))
```

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/GatekeeperEmailContentProvider.java
```diff
@@ -14,18 +16,23 @@
...
         super(uiBaseUrl, dateFormatterService, hearingBookingService);
         this.localAuthorityNameLookupConfiguration = localAuthorityNameLookupConfiguration;
+        this.mapper = mapper;
     }
 
     public Map<String, Object> buildGatekeeperNotification(CaseDetails caseDetails, String localAuthorityCode) {
-        return super.getCasePersonalisationBuilder(caseDetails)
+        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+
+        return super.getCasePersonalisationBuilder(caseDetails, caseData)
```
Gogetter: same comment as above

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
joez-f: Don't think it is necessary - it's fine to add these to getCasePersonalisationBuilder, this method is only used in places where we generate notifications for submitting application (and in those places we want these new values). So don't need to separate into another method

service/src/test/java/uk/gov/hmcts/reform/fpl/service/GatekeeperEmailContentProviderTest.java
```diff
@@ -22,16 +26,29 @@
...
+    @MockBean
     private LocalAuthorityNameLookupConfiguration localAuthorityNameLookupConfiguration;
 
-    @InjectMocks
+    @Autowired
+    private ObjectMapper mapper;
+
+    private final DateFormatterService dateFormatterService = new DateFormatterService();
+    private final HearingBookingService hearingBookingService = new HearingBookingService();
+
```
Gogetter: given your adding both _HearingBookingService_ and _DateFormatterService_ to _ContextConfiguration_ why are we instantiating instead of _@Autowired_ ?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CafcassEmailContentProviderTest.java
```diff
@@ -24,21 +28,35 @@
...
+    @MockBean
     private LocalAuthorityNameLookupConfiguration localAuthorityNameLookupConfiguration;
 
-    @InjectMocks
+    @Autowired
+    private ObjectMapper mapper;
+
+    private final DateFormatterService dateFormatterService = new DateFormatterService();
+    private final HearingBookingService hearingBookingService = new HearingBookingService();
+
```
Gogetter: given your adding both _HearingBookingService_ and _DateFormatterService_ to _ContextConfiguration_ why are we instantiating instead of _@Autowired_ ?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -48,13 +48,7 @@ public static String buildSubjectLineWithHearingBookingDateSuffix(final String s
...
     }
 
-    private static String buildHearingDateText(final List<Element<HearingBooking>> hearingBookings) {
-        return " hearing " + dateFormatterService.formatLocalDateToString(
-            hearingBookingService.getMostUrgentHearingBooking(
-                hearingBookings).getStartDate().toLocalDate(), FormatStyle.MEDIUM);
-    }
-
-    private static String getFirstRespondentLastName(final CaseData caseData) {
+    public static String getFirstRespondentLastName(final CaseData caseData) {
```
Gogetter: given this is now _public_ do we need a test scenario?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CafcassEmailContentProviderTest.java
joez-f: I'll use @Autowired instead 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
joez-f: yes

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
```diff
@@ -17,26 +17,29 @@
...
                                            HearingBookingService hearingBookingService) {
         this.uiBaseUrl = uiBaseUrl;
         this.dateFormatterService = dateFormatterService;
         this.hearingBookingService = hearingBookingService;
     }
 
     @SuppressWarnings("unchecked")
-    ImmutableMap.Builder<String, Object> getCasePersonalisationBuilder(CaseDetails caseDetails) {
-        List<String> ordersAndDirections = buildOrdersAndDirections((Map<String, Object>) caseDetails.getData().get("orders"));
+    ImmutableMap.Builder<String, Object> getCasePersonalisationBuilder(CaseDetails caseDetails, CaseData caseData) {
```
geek-soft: Passing same thing as two different arguments (in slightly different format - Map vs CaseData) does not make much sense. Somebody who will have to use `getCasePersonalisationBuilder` in the future will have bad headache.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
geek-soft: It does not make much sense to have `getFirstRespondentLastName` method in `EmailNotificationHelper` class. It was bad when method was private. Opening on API by making it public makes it even worse.

First getting last name of first responded does not lie within responsibilities of email notification helper and it adds another unnecessary reason to change helper class. Second as a developer I wound never think of looking such method in email notification helper class which makes it harder to reuse. Finally method should not take whole case data while only one small getter is used (method utilises 1/50 part of the class it gets) - method should take list of respondents and nothing more.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
joez-f: I think I could stick with just using caseDetails, I think I would have to do something like:
`        List<Map<String, Object>> respondents = (List<Map<String, Object>>) data.get("respondents1");
        Map<String, Object> value = (Map<String, Object>) respondents.get(0).get("value");
        Map<String, Object> party = (Map<String, Object>) value.get("party");
        Map<String, Object> lastName = (Map<String, Object>) value.get("lastName");`
to get the respondent last name. Bit of a headache, I don't like this way of extracting information; using caseData is a lot cleaner, but did not want to change existing code too much so added caseData as argument for simplicity. If preferred I could try retrieving lastName using Map format

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
joez-f: I agree, I did not know this method was even here until @Gogetter mentioned it to me. Where should this method lie? In a different helper class? And yes we should pass in list of respondents rather than all of caseData.

geek-soft: Could be separate helper, named right way so that it is easier to find / reuse.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProvider.java
geek-soft: I'd personally, as long as time allows, rewrite method to use CaseData. It will be, as you said, way more readable than current implementation. These operations on maps are tech debt ash should be killed sooner than later.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
joez-f: Since ticket is already in PO review stage I won't make further changes to this PR, but will make small refactor PR after merge to move this to helper method and also refactor getCasePersonalisationBuilder to use caseData. It looks like we can use caseData for most of these and then just pass in the Id from caseDetails instead of all of caseDetails

joez-f:
https://tools.hmcts.net/jira/browse/FPLA-1177 have made techDebt ticket, will work on it now and then make PR to address @geek-soft comments

PR #764, opened 2020-01-06
Checks / balances / logs / hacks

Gogetter:
**TEST**


service/src/main/resources/application.yaml
```diff
@@ -20,10 +19,27 @@ idam:
 fpl:
   local_authority_codes_forbidden_case_submission:
 
+logging:
+  level:
+    org:
+      springframework:
+        mail:
+          javamail: error
```
akomisarek: Probably we could even try debug! So we would know more.

Gogetter:
closing in favor of #759

PR #774, opened 2020-01-09
feat: court admin comply on behalf of parties not represented (FPLA-971)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-971

### Change description ###

New method within direcitonHelperService "getDirectionsForUnrepresentedParties".

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -160,8 +167,10 @@ public void addDirectionsToCaseDetails(CaseDetails caseDetails,
                 case PARENTS_AND_RESPONDENTS:
                     directions.addAll(clone);
 
-                    if (eventId == COMPLY_ON_BEHALF_SDO) {
+                    if (eventId == COMPLY_ON_BEHALF_COURT) {
                         filterResponsesNotCompliedOnBehalfOfByTheCourt("RESPONDENT", directions);
+
+                        directions = getDirectionsForUnrepresentedParties(caseDetails, directions);
```
akomisarek: We need to run similar logic for others as well, don't we?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -193,6 +202,85 @@ public void addDirectionsToCaseDetails(CaseDetails caseDetails,
...
+        return directions.stream()
+            .filter(element -> {
+                    List<Element<UUID>> ids = new ArrayList<>();
+
+                    List<Element<Representative>> representatives = defaultIfNull(
+                        (List<Element<Representative>>) caseDetails.getData().get("representatives"), emptyList());
+
+                    // if assigned to specific person and representatives exist -> find representative ids
+                    // if not -> return direction.
+                    if (directionIsValidForFiltering(element, representatives)) {
```
akomisarek: I think we could add another filter step to just filter directionsValidForFiltering. We would have one less if in that lambda.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -193,6 +202,85 @@ public void addDirectionsToCaseDetails(CaseDetails caseDetails,
...
+                            // if respondent exist -> set ids.
+                            // else return direction
+                            if (assigneeValue < respondents.size()) {
+                                ids = respondents.get(assigneeValue).getValue().getRepresentedBy();
+                            } else {
+                                return direction;
+                            }
+                        }
+
+                        if (element.getValue().getOtherPartiesAssignee() != null) {
```
akomisarek: Wouldn't else if be better in that scenario? 🤔 we can't have both other and respondent in single direction, can we?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -193,6 +202,85 @@ public void addDirectionsToCaseDetails(CaseDetails caseDetails,
...
+                            int assigneeValue = element.getValue().getOtherPartiesAssignee().ordinal();
+
+                            Others others = (Others) caseDetails.getData().get("others");
+
+                            List<Element<Other>> listOthers =
+                                defaultIfNull(others.getAdditionalOthers(), new ArrayList<>());
```
akomisarek: is there scenario others can be null? What if I never add someone and then assign the direction to other 1? It's going to throw NPE wouldn't it?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -193,6 +202,85 @@ public void addDirectionsToCaseDetails(CaseDetails caseDetails,
...
+
+                            listOthers.add(0, ElementUtils.element(others.getFirstOther()));
```
akomisarek: I think building this listOthers should be abstracted inside the The others class.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -193,6 +202,85 @@ public void addDirectionsToCaseDetails(CaseDetails caseDetails,
...
+                    }
+
+                    List<Element<Representative>> filteredRepresentatives = representatives.stream()
+                        .filter(x -> unwrapElements(finalIds).contains(UUID.fromString(x.getValue().getIdamId())))
+                        .filter(x -> x.getValue().getServingPreferences() != DIGITAL_SERVICE)
+                        .collect(toList());
+
+                    // if representative left in list, keep direction.
+                    // could be improved with a call to idam? return directions that belong to representative
+                    return !filteredRepresentatives.isEmpty();
```
akomisarek: Don't like collect, maybe .findAny().isPresent() would be better?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -193,6 +202,85 @@ public void addDirectionsToCaseDetails(CaseDetails caseDetails,
...
+                    } else {
+                        // include direction in filter
+                        return direction;
+                    }
+
+                    List<Element<UUID>> finalIds = ids;
+
+                    if (ids.isEmpty()) {
+                        //
+                        return direction;
```
akomisarek: Direction seems as something tangible, but this is boolean value, would you consider to rename?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
swalker125: yep! Just not implemented yet

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
swalker125: I believe this is now solved where I am using object mapper

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
swalker125: solved using object mapper

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
swalker125: I will do return -> filter(find the ids that match).anyMatch(DIGITAL_SERVICE).

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
swalker125: so I wanted this to make a bit more sense rather than returning true false. Can you think of a better name, I think we are getting at the same point. My thinking was, when true -> it adds the direction to the filtered list, hence return direction...

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
swalker125: extracted

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperServiceTest.java
```diff
@@ -1465,6 +1481,209 @@ void shouldReturnCaseManagementOrderDirectionsWhenServedCaseManagementOrdersIsNo
...
+
+            filteredDirections = service.getDirectionsForUnrepresentedParties(caseDetails, directions);
+
+            assertThat(filteredDirections).isEmpty();
+        }
+
+        // this is a case that can currently happen due to no validation when assigning directions.
+        // checks boundary and extreme.
+        @ParameterizedTest
+        @EnumSource(value = ParentsAndRespondentsDirectionAssignee.class, names = {"RESPONDENT_2", "RESPONDENT_10"})
```
akomisarek: 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
akomisarek: So personally how I would implement is:
1. I would extract the lambda to separate method, calling it something like `isDirectionForUnrepresentedParty` -  that reduces cognitive load on the main method which becomes simple: get direction, stream, filter collect and removes direction field from that level.
2. If someone wants to see how filtering is done then can jump to this method. I think in the method itself returning true wouldn't hurt as we already are in the mindset of checking if the direction is for appropriate party. 

If you want to keep it as one bigger methood, probably I would suggest something: `directionForUnrepresentedParty`? Would it be better?

PR #782, opened 2020-01-13
fix: only validate edited and new hearing bookings (FPLA-1178)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1178

### Change description ###

Compare caseDetails before to see which hearing bookings are new or have been updated.

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
@@ -1,6 +1,6 @@
 package uk.gov.hmcts.reform.fpl.controllers;
 
-import com.fasterxml.jackson.core.type.TypeReference;
+import org.assertj.core.util.Lists;
 import org.junit.jupiter.api.Test;
 import org.springframework.boot.test.autoconfigure.OverrideAutoConfiguration;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
@@ -25,9 +25,8 @@ import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.wrapElements;
 @WebMvcTest(HearingBookingDetailsController.class)
 @OverrideAutoConfiguration(enabled = true)
 class HearingBookingDetailsControllerTest extends AbstractControllerTest {
-
     private static final String ERROR_MESSAGE = "Enter a start date in the future";
-    private static final HearingBooking EMPTY_BOOKING = HearingBooking.builder().build();
+    private static final HearingBooking EMPTY_HEARING_BOOKING = HearingBooking.builder().build();
 
     HearingBookingDetailsControllerTest() {
         super("add-hearing-bookings");
@@ -38,7 +37,7 @@ class HearingBookingDetailsControllerTest extends AbstractControllerTest {
         LocalDateTime yesterday = LocalDateTime.now().minusDays(1);
 
         AboutToStartOrSubmitCallbackResponse callbackResponse = makeRequest(createHearing(
-            yesterday, yesterday.plusDays(1)), EMPTY_BOOKING);
+            yesterday, yesterday.plusDays(1)));
 
         assertThat(callbackResponse.getErrors()).contains(ERROR_MESSAGE);
     }
@@ -48,7 +47,7 @@ class HearingBookingDetailsControllerTest extends AbstractControllerTest {
         LocalDateTime tomorrow = LocalDateTime.now().plusDays(1);
 
         AboutToStartOrSubmitCallbackResponse callbackResponse = makeRequest(createHearing(
-            tomorrow, tomorrow.plusDays(1)), EMPTY_BOOKING);
+            tomorrow, tomorrow.plusDays(1)));
 
         assertThat(callbackResponse.getErrors()).doesNotContain(ERROR_MESSAGE);
     }
@@ -58,7 +57,7 @@ class HearingBookingDetailsControllerTest extends AbstractControllerTest {
         LocalDateTime today = LocalDateTime.now();
 
         AboutToStartOrSubmitCallbackResponse callbackResponse = makeRequest(createHearing(
-            today, today.plusDays(1)), EMPTY_BOOKING);
+            today, today.plusDays(1)));
 
         assertThat(callbackResponse.getErrors()).contains(ERROR_MESSAGE);
     }
@@ -68,7 +67,7 @@ class HearingBookingDetailsControllerTest extends AbstractControllerTest {
         LocalDateTime distantPast = LocalDateTime.now().minusYears(10000);
 
         AboutToStartOrSubmitCallbackResponse callbackResponse = makeRequest(createHearing(
-            distantPast, distantPast.plusDays(1)), EMPTY_BOOKING);
+            distantPast, distantPast.plusDays(1)));
 
         assertThat(callbackResponse.getErrors()).contains(ERROR_MESSAGE);
     }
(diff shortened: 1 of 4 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
```diff
@@ -59,4 +59,8 @@ public HearingBooking getHearingBookingByUUID(List<Element<HearingBooking>> hear
             .findFirst()
             .orElse(null);
     }
+
+    public void filterPreviousHearings(List<Element<HearingBooking>> before, List<Element<HearingBooking>> after) {
```
lewisbirks: Change the variable names to `previous` and `all`, bit clearer than before and after

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
```diff
@@ -70,28 +78,76 @@ void shouldNotReturnAnErrorWhenHearingDateIsSetInDistantFuture() throws Exceptio
...
+        assertThat(callbackResponse.getErrors()).doesNotContain(ERROR_MESSAGE);
+    }
+
+    @Test
+    void shouldOnlyValidateNewBookingsWhenBookingAlreadyExists() throws Exception {
+        LocalDateTime distantFuture = LocalDateTime.now().plusYears(1000);
+        LocalDateTime distantPast = LocalDateTime.now().minusYears(10000);
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = makeRequest(createHearing(
+            distantFuture, distantFuture.plusDays(1)), createHearing(distantPast, distantPast.plusDays(1)));
```
akomisarek: That test is not really testing if the old hearing is skipped does it?

Correct me if I am wrong, but that request is essentially removing the distantHearing as the "after" portion contains only one hearing.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
```diff
@@ -70,28 +78,76 @@ void shouldNotReturnAnErrorWhenHearingDateIsSetInDistantFuture() throws Exceptio
...
 
         assertThat(callbackResponse.getErrors()).doesNotContain(ERROR_MESSAGE);
     }
 
+    @Test
+    void shouldErrorWhenExistingBookingIsUpdatedToPastDate() {
+        LocalDateTime futureDate = LocalDateTime.now().plusDays(5);
+        UUID hearingId = randomUUID();
+
+        CallbackRequest request = callbackRequestWithEditedBooking(futureDate, hearingId);
```
akomisarek: Not ideal test, as we don't see on a glance what is previous date, but it's minor for me as the method is directly below so easy to find out.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
lewisbirks: TEST ME PLS

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
```diff
@@ -59,4 +59,8 @@ public HearingBooking getHearingBookingByUUID(List<Element<HearingBooking>> hear
...
+        after.removeIf(before::contains);
+    }
```
akomisarek: So a thought - it looks for me when I edit the hearing in past, it's not going to be removed from list, thus validation will happen for it, i.e. I could modify Magistrate name as there was typo in the name. 

Validation would fail for that entry. 

It essentially makes past hearing not editable does it? Maybe that's what the business would expect? Should we ask them?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
```diff
@@ -70,28 +78,76 @@ void shouldNotReturnAnErrorWhenHearingDateIsSetInDistantFuture() throws Exceptio
...
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postMidEvent(request);
+
+        assertThat(callbackResponse.getErrors()).containsOnlyOnce(ERROR_MESSAGE);
+    }
+
+    private CallbackRequest callbackRequestWithEditedBooking(LocalDateTime futureDate, UUID hearingId) {
+        return CallbackRequest.builder()
+                .caseDetails(CaseDetails.builder()
+                    .data(Map.of("hearingDetails", listBookingWithStartDate(hearingId, futureDate.minusDays(10))))
```
tomaszpowroznik: this is a bit dodgy, what is feature date is defined outside this method in line 99, but this code assumes what is value set there. If line 99 is changed to plusDays(15) then 112 will produce date in future :) This method should encapsulate logic about what is feature and past date

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
```diff
@@ -70,28 +78,76 @@ void shouldNotReturnAnErrorWhenHearingDateIsSetInDistantFuture() throws Exceptio
...
+        LocalDateTime distantPast = LocalDateTime.now().minusYears(10000);
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = makeRequest(createHearing(
+            distantFuture, distantFuture.plusDays(1)), createHearing(distantPast, distantPast.plusDays(1)));
 
         assertThat(callbackResponse.getErrors()).doesNotContain(ERROR_MESSAGE);
     }
 
+    @Test
+    void shouldErrorWhenExistingBookingIsUpdatedToPastDate() {
```
tomaszpowroznik: inconsistency, other tests are named  shouldReturnAnError..

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
```diff
@@ -31,7 +38,7 @@ void shouldReturnAnErrorWhenHearingDateIsSetToYesterday() throws Exception {
         LocalDateTime yesterday = LocalDateTime.now().minusDays(1);
 
         AboutToStartOrSubmitCallbackResponse callbackResponse = makeRequest(createHearing(
-            yesterday, yesterday.plusDays(1)));
+            yesterday, yesterday.plusDays(1)), EMPTY_BOOKING);
```
tomaszpowroznik: inconsistent naming, first param execute method with name about hearing, second param is about booking, but essentially these are same type of things, maybe EMPTY_HEARING_BOOKING.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
```diff
@@ -70,28 +78,76 @@ void shouldNotReturnAnErrorWhenHearingDateIsSetInDistantFuture() throws Exceptio
...
         return HearingBooking.builder()
             .startDate(startDate)
             .endDate(endDate)
             .build();
     }
 
-    private AboutToStartOrSubmitCallbackResponse makeRequest(HearingBooking hearingDetail) throws Exception {
-        Map<String, Object> map = mapper.readValue(mapper.writeValueAsString(hearingDetail),
+    private AboutToStartOrSubmitCallbackResponse makeRequest(HearingBooking after, HearingBooking before) throws Exception {
+        Map<String, Object> mapAfter = mapper.readValue(mapper.writeValueAsString(after),
```
tomaszpowroznik: really minor, naming variable after type is not best option, it does not give any context what is is

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
tomaszpowroznik: personally i would change signature to List<Element<HearingBooking>> getChangedHearing(before, after){
return after.stream().filter(hearing -> !before.contains(hearing)).collect(toList());
}
I think it is a bit better as it does not  mutate original collection and name gives a bit more information what is happening

swalker125: happy to change to more verbose. As midevent it was not persisting changes to data across the lifetime of the event, but does make sense for readability

swalker125: @lewisbirks I think before and after make sense in terms of the method, previous and all, I think they're a bit amiguous. Any further ideas?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerTest.java
swalker125: nope you're right. Different ids because of wrap elements, shall change

PR #788, opened 2020-01-13
feat: emergency protection order (FPLA-830)

Jamesnelson117:
**Before creating a pull request make sure that:**

- [x] commit messages are meaningful and follow good commit message guidelines
- [x] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)

### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/FPLA-830

### Change description ###
Added the 'Emergency protection order' as an option for generation on the 'create an order' event.
Added template to support EPO generation
CCD definition updates to support EPO pages

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #802, opened 2020-01-16
refactor: group enums that are used to map to a case field key (FPLA-NA)

lewisbirks:

N/A

### Change description ###

Made an interface for enums that are used to map to a case field.
Whilst not immediately required, it will be useful for #788, however as it alters the location of CMO keys enum it has been decided to make the change as a separate PR.
Future work could include making a `FixedList` interface and corresponding package with a `getLabel` method to be implemented in a similar way.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/enums/ccd/casefields/GeneratedEPOKey.java
```diff
@@ -1,16 +1,23 @@
...
+public enum GeneratedEPOKey implements CaseField {
     EPO_REMOVAL_ADDRESS("epoRemovalAddress"),
     EPO_CHILDREN("epoChildren"),
     EPO_END_DATE("epoEndDate"),
     EPO_PHRASE("epoPhrase"),
     EPO_TYPE("epoType");
 
     private final String key;
+
+    public static Stream<GeneratedEPOKey> asStream() {
```
lewisbirks: If anyone can think of a nice way to introduce this method as a contract that would be good, the issue is that it has to be static and interfaces can't allow static methods to be overridden and enums can't extend other classes. Any ideas??

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -270,11 +271,8 @@ private void aboutToSubmitAssertions(Map<String, Object> data, GeneratedOrder ex
             List<Element<GeneratedOrder>> orders = mapper.convertValue(data.get("orderCollection"),
                 new TypeReference<>() {});
 
-            Arrays.stream(GeneratedOrderKey.values())
-                .forEach(ccdField -> assertThat(data).doesNotContainKey(ccdField.getKey()));
-
-            Arrays.stream(GeneratedEPOKey.values())
-                .forEach(ccdField -> assertThat(data).doesNotContainKey(ccdField.getKey()));
+            Streams.concat(GeneratedOrderKey.asStream(), GeneratedEPOKey.asStream()).map(CaseField::getKey)
+                .forEach(ccdField -> assertThat(data).doesNotContainKey(ccdField));
```
Jamesnelson117: 👍

PR #809, opened 2020-01-20
fix: added replace method to deal with curly apostrophe (FPLA-1206)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1206

### Change description ###

Problem should not occur with new cases. This fix deals with old cases using a replace call when comparing and adding direction text within the persistHiddenDirectionValues method.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/DirectionHelperService.java
```diff
@@ -110,13 +110,15 @@ public void persistHiddenDirectionValues(List<Element<Direction>> directionWithV
...
-                    directionToAddValue.getValue().setDirectionText(direction.getValue().getDirectionText());
+                    directionToAddValue.getValue().setDirectionText(
+                        direction.getValue().getDirectionText().replace("’", "'"));
                 }
             }));
     }
 
     private boolean hasSameDirectionType(Element<Direction> directionToAddValue, Element<Direction> direction) {
-        return direction.getValue().getDirectionType().equals(directionToAddValue.getValue().getDirectionType());
+        return direction.getValue().getDirectionType()
```
lewisbirks: Does direction need the same treatment?

lewisbirks: If so it might be useful extracting the replacement to a private method

swalker125: no, so this is only actually a problem for the one case that Jason is drafting at the moment. The fix we pushed for christmas trees would not have corrected any data which has curly apostrophes / bullet points for any cases which are currently drafting an sdo.

lewisbirks: 🎄

PR #810, opened 2020-01-20
fix: draft CMO / compliance SDO populating same ccd fields (FPLA-1196)

pd-dudkiewicz:

https://tools.hmcts.net/jira/browse/FPLA-1196

### Change description ###

Draft CMO and compliance SDO populating same CCD fields due to reuse of custom direction fields. Fix includes adding separate fields for CMO.

```
[ ] Yes
[x] No
```


swalker125:
@akomisarek so we were going to always populate a CMO object at the end of the draft event. And then pull that info back out when we go back into draft... Doing so would also fix the problem. What do you think is cleaner? A third set of directions fields, or logic to place and pull a CMO in aboutToStart / aboutToSubmit

PR #824, opened 2020-01-24
feat: Child placement (FPLA-820)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-820

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/PlacementController.java
```diff
@@ -0,0 +1,110 @@
...
+    private final PlacementService placementService;
+
+    @PostMapping("/about-to-start")
+    public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody CallbackRequest callbackRequest) {
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+
+        boolean singleChild = placementService.hasSingleChild(caseData);
+
+        caseDetails.getData().put("singleChild", YesNo.from(singleChild));
```
pd-dudkiewicz: Should we put `caseDetails.getData()` in separate variable? Minor thing, so feel free to resolve.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/PlacementController.java
```diff
@@ -0,0 +1,110 @@
...
+        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+
+        UUID childId = getSelectedChildId(caseDetails, caseData);
+        Element<Child> child = placementService.getChild(caseData, childId);
+
+        Placement placement = mapper.convertValue(caseDetails.getData().get("placement"), Placement.class)
+            .setChild(child);
+
+        caseDetails.getData().put("placements", placementService.setPlacement(caseData, placement));
+        removeFields(caseDetails, "placement", "placementChildName", "singleChild");
```
pd-dudkiewicz: How about `caseDetails.getData().keySet().removeAll(Set.of(...))`? We wouldn't need new method.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/PlacementController.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @PostMapping("/mid-event")
+    public AboutToStartOrSubmitCallbackResponse handleMidEvent(@RequestBody CallbackRequest callbackRequest) {
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+
+        UUID childId = getSelectedChildId(caseDetails, caseData);
+        Element<Child> child = placementService.getChild(caseData, childId);
+
+        caseDetails.getData().put("childrenList", placementService.getChildrenList(caseData, child));
```
pd-dudkiewicz: Other possibility is to use `caseDetails.getData().putAll(Map.of(...))`. It creates unnecessary `Map` instance, but makes it more readable. Thoughts?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Placement.java
```diff
@@ -0,0 +1,40 @@
...
+    public DocumentReference application;
+
+    @JsonProperty("placementSupportingDocuments")
+    private List<Element<PlacementSupportingDocument>> supportingDocuments;
+
+    @JsonProperty("placementOrderAndNotices")
+    private List<Element<PlacementOrderAndNotices>> orderAndNotices;
+
+    @JsonIgnore
+    public Placement setChild(Element<Child> child) {
```
pd-dudkiewicz: That might be an overkill here, but how about going for:
```
public class PlacementBuilder {
    public PlacementBuilder child(Element<Child> child) {
        this.childId = child.getId();
        this.childName = child.getValue().getParty().getFullName();
        return this;
    }
}
```
That way we wouldn't have to change state of `Placement` instance and will return new one instead with `.toBuilder().child().build()`. Lombok just fills the rest of the builder. Thoughts?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/PlacementService.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Optional;
+import java.util.UUID;
+import java.util.function.Function;
+
+import static java.util.Optional.ofNullable;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.findElement;
+
+@Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
```
pd-dudkiewicz: That could probably be removed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/PlacementService.java
```diff
@@ -0,0 +1,73 @@
...
+                .childName(child.getValue().getParty().getFullName())
+                .build());
+    }
+
+    public List<Element<Placement>> setPlacement(CaseData caseData, Placement placement) {
+        List<Element<Placement>> placements = new ArrayList<>(caseData.getPlacements());
+
+        Optional<Element<Placement>> existingPlacement = findPlacement(placements, placement.getChildId());
+
+        if (existingPlacement.isPresent()) {
```
pd-dudkiewicz: How about `ifPresentOrElse`?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/PlacementController.java
tomaszpowroznik: done

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/PlacementController.java
tomaszpowroznik: TBH, i dont see big benefits, still need to add 3 entries to map, and then map to map. Personally i think that creating another nesting does not improve readability.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Placement.java
tomaszpowroznik: i think that it is a bit overkill here, to create new class to just set one value.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/PlacementService.java
tomaszpowroznik: of course:) done

service/src/main/java/uk/gov/hmcts/reform/fpl/service/PlacementService.java
tomaszpowroznik: done,

but  I dont think that having 2 lambdas inside method invocation is most readable solution. What i like about this is that optional is already unboxed and ready to use

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/PlacementController.java
tomaszpowroznik: the benefit of new method is that it gives you context what is happening and why (renamed method a bit)

e2e/tests/localAuthorityMaintainsCaseAfterSubmission_test.js
```diff
@@ -63,3 +63,35 @@ Scenario('local authority provides a statements of service', async (I, caseViewP
   I.seeAnswerInTab(7, 'Recipients 2', 'How were they sent?', recipients[1].sentBy);
   I.seeAnswerInTab(8, 'Recipients 2', 'Recipient\'s email address', recipients[1].email);
 });
+
+Scenario('local authority upload placement application', async (I, caseViewPage, placementEventPage) => {
```
joez-f: maybe we should also test scenario with only 1 children, as it changes which pages are displayed; currently we only cover scenario with 2 children in our test - but if this story is urgent then I wouldn't worry for now

service/src/test/java/uk/gov/hmcts/reform/fpl/config/utils/ElementUtilsTest.java
```diff
@@ -0,0 +1,233 @@
...
+import static java.util.Collections.emptyList;
+import static java.util.UUID.randomUUID;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.asDynamicList;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.findElement;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.unwrapElements;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.wrapElements;
+
+public class ElementUtilsTest {
```
joez-f: Good that we are added tests for ElementUtils now but it is a bit out of scope of PR, perhaps should have been in separate PR (should try to avoid inflating PR with other code changes - but I agree this is valuable). Just note for future

tomaszpowroznik:
> PR looks good to me and functionality works as intended
> 
> Note: there is an edge situation where admin can delete children from case, which then breaks this event. We have no validation to check if children exist or not, so if case has 0 children for some reason then event will fail. We should ideally have validation to stop this. This point should be raised to the business if we are not fixing it within this PR.
> Screenshot 2020-01-24 at 15 06 42

Confirmed with Jay and Theo that it is not real life scenario, and system should rather prevents removal of all children from case after submission instead of preventing placement
Anyway with current situation user wont be able to create event and mutate case into wrong state. Business is happy with how it is now

service/src/test/java/uk/gov/hmcts/reform/fpl/config/utils/ElementUtilsTest.java
tomaszpowroznik: Most of added tests are about functionality which has been added to ElementUtils due to placement needs so in this PR

e2e/tests/localAuthorityMaintainsCaseAfterSubmission_test.js
tomaszpowroznik: i did it on purpose like that, of course we can have another test but it will required new case to be created and submitted as we should not remove children after submission (possible in system but invalid from business pov). My concern was about performance of such solutions as our e2e test are already slow. Most common scenario is with 1 child but due to CCD bug for which we have workaround (only for multi children cases) i wonted to have this scenario covered as there is risk that functionality will broke when CCD fix their issue

PR #827, opened 2020-01-24
fix:owas suppressions for CVE-2019-17563

tomaszpowroznik:
Owasp suppressions

```
[ ] Yes
[X] No
```


lewisbirks:
Would this be resolved if we sorted out #812?

tomaszpowroznik:
> Would this be resolved if we sorted out #812?

yes, 812 would solve this issue

PR #836, opened 2020-01-27
fix: for roboticsdata applicant.organisationName for and include address postcode [FPLA-720]

Gogetter:
### Change description ###
update robotics data to use **_Applicant.organisationName_** as **_getFullName_** is never populated

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
```diff
@@ -86,7 +85,7 @@ private Applicant populateApplicant(final List<Element<uk.gov.hmcts.reform.fpl.m
         if (isNotEmpty(allApplicants)) {
             uk.gov.hmcts.reform.fpl.model.ApplicantParty applicantParty = allApplicants.get(0).getValue().getParty();
```
lewisbirks: Why do we not just import ApplicantParty?

PR #866, opened 2020-02-07
impr:  hearing details mandatory before sealed SDO can be submitted

Gogetter:
[FPLA-939](https://tools.hmcts.net/jira/browse/FPLA-939)

### Change description ###

- When SDO is sealed with no hearing details, give appropriate error
- Draft SDO can still be submitted without hearing details

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -226,4 +211,15 @@ private Document getDocument(String authorization, String userId, Map<String, Ob
 
         return uploadDocumentService.uploadPDF(userId, authorization, document.getBytes(), docTitle);
     }
+
+    private List<String> validate(CaseData caseData) {
```
swalker125: this probably does not belong in the controller. It also means you have to test in integration tests which isn't great. I think I would rather some sort of error message validation service. We can then extensively unit test and then cut down on the integration tests. What are your thoughts?

Gogetter: Absolutely agree it does not belong in here. Went with this option as its a minimal update and tried to reduce the PR size. Also, I think we will still need the integration test for the new scenario?

swalker125: hmm, so yes, a integration test is probably required. But with extensinve units test, the four we currently have would not be required

Gogetter: updated

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrderValidationServiceTest.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private CaseDetails buildCaseDetails(final OrderStatus orderStatus) {
+        return CaseDetails.builder()
+            .data(ImmutableMap.<String, Object>builder()
+                .put("standardDirectionOrder", buildOrderWithStatus(orderStatus))
+                .put("judgeAndLegalAdvisor", JudgeAndLegalAdvisor.builder().build())
+                .build())
+            .build();
+    }
```
lewisbirks: Little bugbear but could these be below the method below, makes more sense to me as you call `buildCaseDetailsWithHearingDetails` which calls `buildCaseDetails` which calls `buildOrderWithStatus`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrderValidationService.java
```diff
@@ -0,0 +1,25 @@
...
+import static java.util.Collections.singletonList;
+import static org.springframework.util.CollectionUtils.isEmpty;
+import static uk.gov.hmcts.reform.fpl.enums.OrderStatus.SEALED;
+
+@Service
+public class OrderValidationService {
+    public List<String> validate(final CaseData caseData) {
+        if (SEALED == caseData.getStandardDirectionOrder().getOrderStatus()
+            && isEmpty(caseData.getHearingDetails())) {
+            return singletonList(
```
lewisbirks: I think it is more of a convention to use `List.of` instead of `Collections.singletonList`

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -179,6 +162,7 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(
 
         return AboutToStartOrSubmitCallbackResponse.builder()
             .data(caseDetails.getData())
+            .errors(orderValidationService.validate(updated))
```
lewisbirks: I would wrap the above code in an if that means that we don't have to do any additional computation if there are errors

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrderValidationService.java
Gogetter: went with `Collections.singletonList` as we are just returning one string and `List.of` will be preferred if we have more than one string/object.

PR #870, opened 2020-02-10
feat: notify admin when any order issued (FPLA-25 / FPLA-1208)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-25
https://tools.hmcts.net/jira/browse/FPLA-1208

https://www.notifications.service.gov.uk/services/12f756df-f01d-4a32-a405-e1ea8a494fbb/templates/c652061a-31db-4ec5-bcf7-4acec5943658

### Change description ###
Updated notification handler to send an email to admin when any of the following orders are issued:

- 'Create an order' order (25)
- Notice of placement order (25)
- CMO (1208)

If any representatives need serving by post, then include their addresses in the notification

If it's a CMO, then include grey callout

### Test cases for QA: ###

Case with no representatives: verify notification sent to admin after

- Issuing CMO as judge (should have grey callout)
- Creating an order from 'Create an order'
- Notice of placement order uploaded/changed in 'Placement' event

Case with representatives who need serving by post: verify notification sent to admin after

- Issuing CMO as judge (should have grey callout)
- Creating an order from 'Create an order'
- Notice of placement order uploaded/changed in 'Placement' event. Should send notification for each new/changed notice of placement order uploaded during event

 (should include representatives address in notification)

```
[ ] Yes
[X] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerTest.java
```diff
@@ -466,6 +481,12 @@ private void verifyCMOTriggerEventAndNotificationSentToLocalAuthorityOnApprovedC
             eq(getExpectedCMOIssuedCaseLinkNotificationParameters(LOCAL_AUTHORITY_NAME)), eq(CASE_ID));
     }
 
+    private void verifyNotificationSentToAdminWhenCMOIssuedWithNoServingNeeded() throws NotificationClientException {
+        verify(notificationClient).sendEmail(
+            eq(ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN), eq("admin@family-court.com"),
+            eq(getExpectedParametersForAdminWhenNoRepresentativesServedByPost()), eq(CASE_ID));
```
swalker125: I don't like this method name, but assume it already exists, so no drama

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -125,20 +143,56 @@ void shouldReturnErrorsWhenFamilymanNumberIsNotProvided() {
         assertThat(callbackResponse.getErrors()).containsExactly("Enter Familyman case number");
     }
 
-    @Test
-    void shouldTriggerOrderEventWhenSubmitted() throws Exception {
-        String expectedCaseReference = "19898989";
-        postSubmittedEvent(buildCallbackRequest());
+    @TestInstance(PER_CLASS)
```
swalker125: I don't think this annotation is needed

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -125,20 +143,56 @@ void shouldReturnErrorsWhenFamilymanNumberIsNotProvided() {
...
+        @AfterEach
+        void resetInvocations() {
+            reset(notificationClient);
+        }
+
+        @Test
+        void submittedShouldNotifyAdminAndLAWhenNoRepresentativesNeedServing() throws Exception {
+            postSubmittedEvent(buildCallbackRequest());
+
+            verify(notificationClient).sendEmail(
```
swalker125: I like this verify eq assertions each on a new line. What do you think?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -125,20 +143,56 @@ void shouldReturnErrorsWhenFamilymanNumberIsNotProvided() {
...
-            eq(expectedOrderLocalAuthorityParameters()), eq(expectedCaseReference));
+            verifyZeroInteractions(notificationClient);
+        }
     }
 
     private CallbackRequest buildCallbackRequest() {
         return CallbackRequest.builder()
             .caseDetails(CaseDetails.builder()
-                .id(19898989L)
+                .id(12345L)
```
swalker125: why change?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -158,8 +227,8 @@ private CallbackRequest buildCallbackRequest() {
             .put("linkToDocument", documentUrl)
             .put("hearingDetailsCallout", subjectLine + ", hearing " + dateFormatterService.formatLocalDateToString(
                 dateIn3Months.toLocalDate(), FormatStyle.MEDIUM))
-            .put("reference", "19898989")
-            .put("caseUrl", "http://fake-url/case/" + JURISDICTION + "/" + CASE_TYPE + "/19898989")
+            .put("reference", CASE_ID)
+            .put("caseUrl", "http://fake-url/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345")
```
swalker125: why not use CASE_ID again?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
```diff
@@ -142,15 +146,18 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(
...
-                                     @RequestHeader(value = "user-id") String userId,
-                                     @RequestBody CallbackRequest callbackRequest) {
+    public void handleSubmittedEvent(@RequestBody CallbackRequest callbackRequest) {
         CaseData caseData = mapper.convertValue(callbackRequest.getCaseDetails().getData(), CaseData.class);
         String mostRecentUploadedDocumentUrl = service.getMostRecentUploadedOrderDocumentUrl(
             caseData.getOrderCollection());
 
-        applicationEventPublisher.publishEvent(new GeneratedOrderEvent(callbackRequest, authorization, userId,
-            concatGatewayConfigurationUrlAndMostRecentUploadedOrderDocumentPath(mostRecentUploadedDocumentUrl)));
+        byte[] documentContents = documentDownloadService.downloadDocument(requestData.authorisation(),
```
swalker125: so my understanding of requestData, you can just declare it in DocumentDownloadService. Using it like this doesn't actually provide any benefits over declaring it in the method signature. I would recommend moving and only sending mostRecentUploadedDocumentUrl

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -107,11 +110,14 @@ public void sendNotificationForC2Upload(final C2UploadedEvent event) {
     }
 
     @EventListener
-    public void sendNotificationForOrder(final GeneratedOrderEvent event) {
+    public void sendNotificationsForGeneratedOrder(final GeneratedOrderEvent event) {
```
swalker125: why change?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -107,11 +110,14 @@ public void sendNotificationForC2Upload(final C2UploadedEvent event) {
...
         EventData eventData = new EventData(event);
 
-        sendOrderNotificationForLocalAuthority(eventData.getCaseDetails(), eventData.getLocalAuthorityCode(),
+        sendGeneratedOrderNotificationToLocalAuthority(eventData.getCaseDetails(), eventData.getLocalAuthorityCode(),
             event.getMostRecentUploadedDocumentUrl());
+
+        sendOrderIssuedNotificationToHmctsAdmin(eventData.getCaseDetails(), eventData.getLocalAuthorityCode(),
```
swalker125: *court admin? or do we use hmcts admin in other places?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -285,15 +293,29 @@ private void sendNotification(String templateId, String email, Map<String, Objec
...
     }
 
-    private void sendOrderNotificationForLocalAuthority(final CaseDetails caseDetails, final String localAuthorityCode,
-                                                        final String mostRecentUploadedDocumentUrl) {
+    private void sendGeneratedOrderNotificationToLocalAuthority(final CaseDetails caseDetails,
+                                                                final String localAuthorityCode,
+                                                                final String mostRecentUploadedDocumentUrl) {
         Map<String, Object> localAuthorityParameters =
-            orderEmailContentProvider.buildOrderNotificationParametersForLocalAuthority(
+            generatedOrderEmailContentProvider.buildOrderNotificationParametersForLocalAuthority(
```
swalker125: why change? I don't see what benefit adding generated gives

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -285,15 +293,29 @@ private void sendNotification(String templateId, String email, Map<String, Objec
...
-        sendNotification(ORDER_NOTIFICATION_TEMPLATE, recipientEmail, localAuthorityParameters,
+        sendNotification(ORDER_NOTIFICATION_TEMPLATE_FOR_LA, recipientEmail, localAuthorityParameters,
+            Long.toString(caseDetails.getId()));
+    }
+
+    private void sendOrderIssuedNotificationToHmctsAdmin(final CaseDetails caseDetails,
+                                                         final String localAuthorityCode,
+                                                         final byte[] documentContents) {
+        Map<String, Object> hmctsParameters =
+            orderIssuedEmailContentProvider.buildOrderNotificationParametersForHmctsAdmin(
```
swalker125: what's the difference between this and generatedOrderEmailContentProvider. Should they be separate?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -285,15 +293,29 @@ private void sendNotification(String templateId, String email, Map<String, Objec
...
+                caseDetails, localAuthorityCode, documentContents);
+
+        String hmctsEmail = hmctsCourtLookupConfiguration.getCourt(localAuthorityCode).getEmail();
+
+        sendNotification(ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN, hmctsEmail, hmctsParameters,
```
swalker125: I would just use, parameters, and email. Don't think hmcts is needed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,79 @@
...
+        CaseData caseData = objectMapper.convertValue(caseDetails.getData(), CaseData.class);
+        List<Representative> representativesServedByPost = representativeService.getRepresentativesByServedPreference(
+            caseData.getRepresentatives(), POST);
+        List<String> formattedRepresentatives = formatRepresentativesForPostNotification(representativesServedByPost);
+
+        return ImmutableMap.<String, Object>builder()
+            .put("needsPosting", !representativesServedByPost.isEmpty() ? "Yes" : "No")
+            .put("doesNotNeedPosting", representativesServedByPost.isEmpty() ? "Yes" : "No")
+            .put("courtName", localAuthorityNameLookupConfiguration.getLocalAuthorityName(localAuthorityCode))
+            .putAll(caseUrlOrDocumentLink(!representativesServedByPost.isEmpty(), documentContents,
```
swalker125: why !representativesServedByPost.isEmpty()? Seems weird

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,79 @@
...
+    public Map<String, Object> buildOrderNotificationParametersForHmctsAdmin(final CaseDetails caseDetails,
+                                                                             final String localAuthorityCode,
+                                                                             final byte[] documentContents) {
+        CaseData caseData = objectMapper.convertValue(caseDetails.getData(), CaseData.class);
+        List<Representative> representativesServedByPost = representativeService.getRepresentativesByServedPreference(
+            caseData.getRepresentatives(), POST);
+        List<String> formattedRepresentatives = formatRepresentativesForPostNotification(representativesServedByPost);
+
+        return ImmutableMap.<String, Object>builder()
+            .put("needsPosting", !representativesServedByPost.isEmpty() ? "Yes" : "No")
```
swalker125: why ! I think there is a isNotEmpty() method? logic seems confusing

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,79 @@
...
+                caseDetails.getId()))
+            .put("respondentLastName", getFirstRespondentLastName(caseData.getRespondents1()))
+            .put("representatives", formattedRepresentatives.isEmpty() ? "" : formattedRepresentatives)
+            .build();
+    }
+
+    private Map<String, Object> caseUrlOrDocumentLink(boolean needsServing,
+                                                      final byte[] documentContents,
+                                                      Long caseId) {
+        ImmutableMap.Builder<String, Object> url = ImmutableMap.builder();
```
swalker125: why is this a map? variable name seems wrong

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+        try {
+            url.put("caseUrlOrDocumentLink", needsServing ? prepareUpload(documentContents)
```
swalker125: you could just return the string url here no?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,79 @@
...
+                : formatCaseUrl(uiBaseUrl, caseId));
+        } catch (NotificationClientException e) {
+            log.error("Unable to send notification due to ", e);
+        }
```
swalker125: is this right?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
```diff
@@ -21,4 +23,11 @@ public static String getFirstRespondentLastName(List<Element<Respondent>> respon
             .map(RespondentParty::getLastName)
             .orElse("");
     }
+
+    //To be deleted when print and post is available to use (before March 3rd)
+    public static List<String> formatRepresentativesForPostNotification(List<Representative> representatives) {
+        return representatives.stream()
+            .map(rep -> rep.getFullName() + "\n" + rep.getAddress().getAddressAsString(", "))
```
swalker125: use full words please

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -236,12 +242,16 @@ void shouldNotifyNonHmctsAdminOnC2Upload() throws IOException, NotificationClien
...
+                AUTH_TOKEN, USER_ID, mostRecentUploadedDocumentUrl, documentContents));
 
             verify(notificationClient, times(1)).sendEmail(
-                eq(ORDER_NOTIFICATION_TEMPLATE), eq(LOCAL_AUTHORITY_EMAIL_ADDRESS),
+                eq(ORDER_NOTIFICATION_TEMPLATE_FOR_LA), eq(LOCAL_AUTHORITY_EMAIL_ADDRESS),
                 eq(orderLocalAuthorityParameters), eq("12345"));
+
+            verify(notificationClient).sendEmail(
+                eq(ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN), eq(COURT_EMAIL_ADDRESS),
+                anyMap(), eq("12345"));
```
swalker125: anyMap() 👀 

why aren't we asserting parameters are correct here. I would say this is the most important part

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -295,6 +308,10 @@ void shouldNotifyLocalAuthorityOfCMOIssued() throws Exception {
             verify(notificationClient).sendEmail(
                 eq(CMO_ORDER_ISSUED_CASE_LINK_NOTIFICATION_TEMPLATE), eq(LOCAL_AUTHORITY_EMAIL_ADDRESS),
                 eq(expectedCMOIssuedNotificationParameters), eq("12345"));
+
+            verify(notificationClient).sendEmail(
+                eq(ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN), eq(COURT_EMAIL_ADDRESS),
+                anyMap(), eq("12345"));
```
swalker125: anyMap() 👀

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -347,7 +367,8 @@ void shouldNotifyAdminOfCMOReadyForJudgeReview() throws Exception {
             given(hmctsCourtLookupConfiguration.getCourt(LOCAL_AUTHORITY_CODE))
                 .willReturn(new Court(COURT_NAME, COURT_EMAIL_ADDRESS, COURT_CODE));
 
-            given(caseManagementOrderEmailContentProvider.buildCMOReadyForJudgeReviewNotificationParameters(caseDetails))
+            given(
```
swalker125: why is this changed? intentional?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,86 @@
+package uk.gov.hmcts.reform.fpl.utils;
```
swalker125: I am a firm believer that private methods for tests should belong in the test class they belong to

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
joez-f: Will check and remove if not needed

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
joez-f: Sure - just saw same thing from 1193 👍

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
joez-f: good point

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
joez-f: for my tests I'm expecting case ID to be 12345 so I need to change the ID here otherwise my tests will fail, hope it makes sense

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
joez-f: I agree, happy to add RequestData to DocumentDownloadService although it might inflate the PR a little? thoughts?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: I've added 'sendOrderIssuedNotificationToHmctsAdmin' so for clarity I thought I'd specify that this is for the GeneratedOrder event so there is a clear differentiation between a 'Generated Order' and any 'Order'. if you don't like, can revert

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: We use HmctsAdmin everywhere

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: see above comment about differentiating between orders that come from 'Create an order' event (Generated orders) and other orders

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: yes - generatedOrderEmailContentProvider is only for orders that come from 'Create an order' event. but I want to send a notification from 3 different triggers (CMO, notice of placement and create order event). So I need a new service that can provide to these 3 controllers.
Initially I did actually have them together but then realised they should be separate

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: no isNotEmpty() method for lists

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: see above

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: think name is fine - soon to be changed to just be caseURL when we have print&post

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: will leave as is, confusing gov notify stuff

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: discussed

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
joez-f: will have a quick try 👍

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
joez-f: line was too long

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: Multiple controller tests make use of these private methods, this seems better than duplicating for each controller test

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: used ObjectUtils isNotEmpty() as discussed

service/src/main/java/uk/gov/hmcts/reform/fpl/NotifyTemplates.java
```diff
@@ -11,7 +11,8 @@ private NotifyTemplates() {
     public static final String GATEKEEPER_SUBMISSION_TEMPLATE = "a79cffc8-7aec-445c-b310-7bed5fb479bc";
     public static final String STANDARD_DIRECTION_ORDER_ISSUED_TEMPLATE = "c06968ea-048b-49d6-a4b5-1f13ee995ba1";
     public static final String C2_UPLOAD_NOTIFICATION_TEMPLATE = "6b961e81-c5ff-4f6f-8c56-f90d932a2f9b";
-    public static final String ORDER_NOTIFICATION_TEMPLATE = "1f7c134e-f9c0-44ba-aa50-fce53eb208f7";
+    public static final String ORDER_NOTIFICATION_TEMPLATE_FOR_LA = "1f7c134e-f9c0-44ba-aa50-fce53eb208f7";
+    public static final String ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN = "c652061a-31db-4ec5-bcf7-4acec5943658";
     public static final String CMO_ORDER_ISSUED_CASE_LINK_NOTIFICATION_TEMPLATE =
```
Gogetter: please could you confirm the `cmoCallout` actually displays as a callout once the email is sent?
Screenshot 2020-02-14 at 10 25 10

service/src/main/java/uk/gov/hmcts/reform/fpl/events/CallbackEvent.java
```diff
@@ -4,6 +4,7 @@
 
 public class CallbackEvent {
 
+    //TODO use RequestData here
```
Gogetter: do we still need this todo?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
```diff
@@ -318,18 +332,31 @@ private void sendNotification(String templateId, String email, Map<String, Objec
...
 
-        sendNotification(ORDER_NOTIFICATION_TEMPLATE, recipientEmail, localAuthorityParameters,
+        sendNotification(ORDER_NOTIFICATION_TEMPLATE_FOR_LA, recipientEmail, localAuthorityParameters,
             Long.toString(caseDetails.getId()));
     }
 
+    private void sendOrderIssuedNotificationToHmctsAdmin(final CaseDetails caseDetails,
+                                                         final String localAuthorityCode,
+                                                         final byte[] documentContents,
+                                                         final IssuedOrderType issuedOrderType) {
```
Gogetter: any value just passing in only `EventData` and `IssuedOrderType` here?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocumentDownloadService.java
```diff
@@ -20,17 +21,18 @@
 
 @Slf4j
 @Service
-@RequiredArgsConstructor(onConstructor = @ __(@Autowired))
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class DocumentDownloadService {
     private final AuthTokenGenerator authTokenGenerator;
     private final DocumentDownloadClientApi documentDownloadClient;
     private final IdamApi idamApi;
+    private final RequestData requestData;
```
Gogetter: splendid 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,88 @@
...
+                                                                             final byte[] documentContents,
+                                                                             final IssuedOrderType issuedOrderType) {
+        CaseData caseData = objectMapper.convertValue(caseDetails.getData(), CaseData.class);
+        final String cmoCallout = buildSubjectLine(caseData);
+        List<Representative> representativesServedByPost = representativeService.getRepresentativesByServedPreference(
+            caseData.getRepresentatives(), POST);
+        List<String> formattedRepresentatives = formatRepresentativesForPostNotification(representativesServedByPost);
+
+        return ImmutableMap.<String, Object>builder()
+            .put("cmoCallout", issuedOrderType == CMO ? "^" + buildSubjectLineWithHearingBookingDateSuffix(
```
Gogetter: why prefer this option (implementing the callout from code)?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
```diff
@@ -21,4 +23,12 @@ public static String getFirstRespondentLastName(List<Element<Respondent>> respon
             .map(RespondentParty::getLastName)
             .orElse("");
     }
+
+    //To be deleted when print and post is available to use (before March 3rd)
```
Gogetter: do we have a jira to track this or do we need one?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -125,20 +144,63 @@ void shouldReturnErrorsWhenFamilymanNumberIsNotProvided() {
...
+            verify(notificationClient).sendEmail(
+                eq(ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN),
+                eq("admin@family-court.com"),
+                dataCaptor.capture(),
+                eq(CASE_ID));
+
+            MapDifference<String, Object> difference = verifyNotificationSentToAdminWhenOrderIssued(dataCaptor,
+                IssuedOrderType.GENERATED_ORDER);
+
+            assertThat(difference.areEqual()).isTrue();
```
swalker125: why do we need to do this? Will assertThat(dataCaptor).isEqualTo(GENERATED_ORDER) not work? I cannot see the verifyNotificationSentToAdminWhenOrderIssued method so I am not sure of implementation. Maybe just explain to me what this is doing

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
swalker125: you use CASE_ID in some places, but then not here. Please be consistent

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -149,6 +211,21 @@ private CallbackRequest buildCallbackRequest() {
             .build();
     }
 
+    private CallbackRequest buildCallbackRequestWithRepresentatives() {
+        return CallbackRequest.builder()
+            .caseDetails(CaseDetails.builder()
+                .id(12345L)
```
swalker125: parseLong(CASE_ID) I think

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementSubmittedEventControllerTest.java
```diff
@@ -184,26 +206,26 @@ void shouldNotSendEmailNotificationWhenNoChangesToOrder() throws NotificationCli
...
-                eq("1"));
+                eq(CASE_ID));
         }
 
         private Map<String, Object> expectedParameters() {
             return Map.of(
-                "respondentLastName", "Nelson",
-                "caseUrl", String.format("%s/case/%s/%s/%s", "http://fake-url", JURISDICTION, CASE_TYPE, 1L));
+                "respondentLastName", "Jones",
+                "caseUrl", String.format("%s/case/%s/%s/%s", "http://fake-url", JURISDICTION, CASE_TYPE, 12345L));
```
swalker125: let's use CASE_ID

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementSubmittedEventControllerTest.java
```diff
@@ -232,7 +254,7 @@ private CallbackRequest callbackRequestWithMatchingCaseDetailsBefore() {
 
         private CaseDetails populatedCaseDetails(UUID representativeId, Respondent respondent) {
             return CaseDetails.builder()
-                .id(1L)
+                .id(12345L)
```
swalker125: lets use CASE_ID

service/src/main/java/uk/gov/hmcts/reform/fpl/NotifyTemplates.java
```diff
@@ -11,7 +11,8 @@ private NotifyTemplates() {
     public static final String GATEKEEPER_SUBMISSION_TEMPLATE = "a79cffc8-7aec-445c-b310-7bed5fb479bc";
     public static final String STANDARD_DIRECTION_ORDER_ISSUED_TEMPLATE = "c06968ea-048b-49d6-a4b5-1f13ee995ba1";
     public static final String C2_UPLOAD_NOTIFICATION_TEMPLATE = "6b961e81-c5ff-4f6f-8c56-f90d932a2f9b";
-    public static final String ORDER_NOTIFICATION_TEMPLATE = "1f7c134e-f9c0-44ba-aa50-fce53eb208f7";
+    public static final String ORDER_NOTIFICATION_TEMPLATE_FOR_LA = "1f7c134e-f9c0-44ba-aa50-fce53eb208f7";
```
swalker125: I would describe what triggers the notification. Is an order uploaded?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
```diff
@@ -142,15 +146,17 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(
...
         String mostRecentUploadedDocumentUrl = service.getMostRecentUploadedOrderDocumentUrl(
             caseData.getOrderCollection());
 
-        applicationEventPublisher.publishEvent(new GeneratedOrderEvent(callbackRequest, authorization, userId,
-            concatGatewayConfigurationUrlAndMostRecentUploadedOrderDocumentPath(mostRecentUploadedDocumentUrl)));
+        byte[] documentContents = documentDownloadService.downloadDocument(mostRecentUploadedDocumentUrl);
+
+        applicationEventPublisher.publishEvent(
+            new GeneratedOrderEvent(callbackRequest, requestData.authorisation(), requestData.userId(),
+                concatGatewayConfigurationUrlAndMostRecentUploadedOrderDocumentPath(mostRecentUploadedDocumentUrl),
```
swalker125: this method name seems incredibly verbose... I mean does it just concat a string? why is this extracted?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
swalker125: please don't resolve comments before they are resolved! I would revert, it doesn't make it any clearer imo

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
swalker125: again, I don't see the benefit

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
swalker125:
```suggestion
            .put("callout", issuedOrderType == CMO ? "^" + buildSubjectLineWithHearingBookingDateSuffix(
```

swalker125: not fussed but cmoCallout adds no benefit

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,88 @@
...
+        this.representativeService = representativeService;
+        this.localAuthorityNameLookupConfiguration = localAuthorityNameLookupConfiguration;
+    }
+
+    public Map<String, Object> buildOrderNotificationParametersForHmctsAdmin(final CaseDetails caseDetails,
+                                                                             final String localAuthorityCode,
+                                                                             final byte[] documentContents,
+                                                                             final IssuedOrderType issuedOrderType) {
+        CaseData caseData = objectMapper.convertValue(caseDetails.getData(), CaseData.class);
+        final String cmoCallout = buildSubjectLine(caseData);
```
swalker125:
```suggestion
        final String callout = buildSubjectLine(caseData);
```

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -0,0 +1,88 @@
...
+        List<Representative> representativesServedByPost = representativeService.getRepresentativesByServedPreference(
+            caseData.getRepresentatives(), POST);
+        List<String> formattedRepresentatives = formatRepresentativesForPostNotification(representativesServedByPost);
```
swalker125: big wall of text here

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
```diff
@@ -21,4 +23,12 @@ public static String getFirstRespondentLastName(List<Element<Respondent>> respon
...
+    public static List<String> formatRepresentativesForPostNotification(List<Representative> representatives) {
+        return representatives.stream()
+            .map(representative -> representative.getFullName() + "\n" + representative.getAddress().getAddressAsString(
+                ", "))
+            .collect(Collectors.toList());
```
swalker125:
```suggestion
            .collect(toList());
```

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
```diff
@@ -21,4 +23,12 @@ public static String getFirstRespondentLastName(List<Element<Respondent>> respon
             .map(RespondentParty::getLastName)
             .orElse("");
     }
+
+    //To be deleted when print and post is available to use (before March 3rd)
+    public static List<String> formatRepresentativesForPostNotification(List<Representative> representatives) {
+        return representatives.stream()
+            .map(representative -> representative.getFullName() + "\n" + representative.getAddress().getAddressAsString(
```
swalker125: could extract the logic to build the address string to separate private method

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -273,30 +289,41 @@ void setup() {
...
-                localAuthorityNameLookupConfiguration, objectMapper);
+                gatekeeperEmailContentProvider, c2UploadedEmailContentProvider, generatedOrderEmailContentProvider,
+                orderIssuedEmailContentProvider, localAuthorityEmailContentProvider, notificationClient, idamApi,
+                inboxLookupService, caseManagementOrderEmailContentProvider, placementApplicationContentProvider,
+                representativeService, localAuthorityNameLookupConfiguration, objectMapper);
         }
 
         @Test
-        void shouldNotifyLocalAuthorityOfCMOIssued() throws Exception {
+        void shouldNotifyAdminAndLAOfCMOIssued() throws Exception {
```
swalker125: I don't like abbriviations

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -273,30 +289,41 @@ void setup() {
...
             cmoNotificationHandler.sendNotificationsForIssuedCaseManagementOrder(
                 new CaseManagementOrderIssuedEvent(callbackRequest, AUTH_TOKEN, USER_ID, documentContents));
 
             verify(notificationClient).sendEmail(
                 eq(CMO_ORDER_ISSUED_CASE_LINK_NOTIFICATION_TEMPLATE), eq(LOCAL_AUTHORITY_EMAIL_ADDRESS),
                 eq(expectedCMOIssuedNotificationParameters), eq("12345"));
+
+            verify(notificationClient).sendEmail(
+                eq(ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN), eq(COURT_EMAIL_ADDRESS),
+                eq(getExpectedParametersForAdminWhenNoRepresentativesServedByPost()), eq("12345"));
```
swalker125: eq on new lines?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/GeneratedOrderEmailContentProviderTest.java
```diff
@@ -79,7 +79,7 @@ void shouldReturnExactOrderLocalAuthorityNotificationParametersWithUploadedDocum
         CaseDetails caseDetails = createCaseDetailsWithSingleOrderElement();
 
         Map<String, Object> returnedLocalAuthorityParameters =
-            orderEmailContentProvider.buildOrderNotificationParametersForLocalAuthority(
+            generatedOrderEmailContentProvider.buildOrderNotificationParametersForLocalAuthority(
```
swalker125: if we revert this is file change less

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.hmcts.reform.fpl.enums.IssuedOrderType.CMO;
+import static uk.gov.hmcts.reform.fpl.enums.RepresentativeServingPreferences.POST;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.wrapElements;
+import static uk.gov.hmcts.reform.fpl.utils.EmailNotificationHelper.formatCaseUrl;
+
+public class NotifyAdminOrderIssuedTestHelper {
+
+    private NotifyAdminOrderIssuedTestHelper() {
+    }
```
swalker125: is constructor required?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final byte[] PDF = {1, 2, 3, 4, 5};
+    private static final String LOCAL_AUTHORITY_NAME = "Example Local Authority";
+
+    public static MapDifference<String, Object> verifyNotificationSentToAdminWhenOrderIssued(
+        ArgumentCaptor<Map<String, Object>> dataCaptor, IssuedOrderType issuedOrderType) {
+
+        Map<String, Object> resultData = new HashMap<>(dataCaptor.getValue());
+        Map<String, Object> expectedParameters = issuedOrderType == CMO
+            ? getExpectedCMOParametersForAdminWhenRepresentativesNeedServingByPost() :
+            getExpectedParametersForAdminWhenRepresentativesNeedServingByPost();
```
swalker125: I wouldn't do this here.

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+
+        assertThat(((JSONObject) resultData.get("caseUrlOrDocumentLink")).get("file")).isEqualTo(
+            ((JSONObject) expectedParameters.get("caseUrlOrDocumentLink")).get("file"));
+
+        resultData.remove("caseUrlOrDocumentLink");
+        expectedParameters.remove("caseUrlOrDocumentLink");
```
swalker125: why?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+    public static MapDifference<String, Object> verifyNotificationSentToAdminWhenOrderIssued(
+        ArgumentCaptor<Map<String, Object>> dataCaptor, IssuedOrderType issuedOrderType) {
+
+        Map<String, Object> resultData = new HashMap<>(dataCaptor.getValue());
+        Map<String, Object> expectedParameters = issuedOrderType == CMO
+            ? getExpectedCMOParametersForAdminWhenRepresentativesNeedServingByPost() :
+            getExpectedParametersForAdminWhenRepresentativesNeedServingByPost();
+
+        assertThat(((JSONObject) resultData.get("caseUrlOrDocumentLink")).get("file")).isEqualTo(
+            ((JSONObject) expectedParameters.get("caseUrlOrDocumentLink")).get("file"));
```
swalker125: why JSONObject? Does isEqualToComparingFieldByField not solve any problems here?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+
+        return Maps.difference(expectedParameters, resultData);
+    }
+
+    public static Map<String, Object> getExpectedParametersForAdminWhenNoRepresentativesServedByPost() {
+        return ImmutableMap.<String, Object>builder()
+            .put("cmoCallout", "")
+            .putAll(commonParametersNoPostingNeeded())
+            .build();
+    }
```
swalker125: Map.of()

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public static Map<String, Object> getExpectedCMOParametersForAdminWhenNoRepresentativesServedByPost() {
+        return ImmutableMap.<String, Object>builder()
+            .put("cmoCallout", "^Jones, SACCCCCCCC5676576567")
+            .putAll(commonParametersNoPostingNeeded())
+            .build();
```
swalker125: Map.of()

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+                .addressLine1("Street")
+                .postTown("Town")
+                .postcode("Postcode")
+                .build())
+            .servingPreferences(POST)
+            .build());
+    }
+
+    private static Map<String, Object> commonParametersNoPostingNeeded() {
+        return ImmutableMap.<String, Object>builder()
```
swalker125: Map.of()

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+        String fileContent = new String(Base64.encodeBase64(PDF), ISO_8859_1);
+        JSONObject jsonFileObject = new JSONObject().put("file", fileContent);
+
+        Map<String, Object> expectedMap = new HashMap<>();
+        expectedMap.put("needsPosting", "Yes");
+        expectedMap.put("doesNotNeedPosting", "No");
+        expectedMap.put("courtName", LOCAL_AUTHORITY_NAME);
+        expectedMap.put("respondentLastName", "Jones");
+        expectedMap.put("representatives", List.of("Paul Blart\nStreet, Town, Postcode"));
+        expectedMap.put("caseUrlOrDocumentLink", jsonFileObject);
```
swalker125: might be too many lines, but Map.of() ?

service/src/main/java/uk/gov/hmcts/reform/fpl/NotifyTemplates.java
joez-f: image

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
swalker125: or does it need to be mutable?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+
+        return expectedMap;
+    }
+
+    private static Map<String, Object> getExpectedParametersForAdminWhenRepresentativesNeedServingByPost() {
```
swalker125: needs to be mutable?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
```diff
@@ -0,0 +1,119 @@
...
+        Map<String, Object> expectedMap = new HashMap<>();
+        expectedMap.put("cmoCallout", "");
+        expectedMap.putAll(commonParametersPostingNeeded());
+
+        return expectedMap;
+    }
+
+    private static Map<String, Object> getExpectedCMOParametersForAdminWhenRepresentativesNeedServingByPost() {
```
swalker125: needs to be mutable?

service/src/main/java/uk/gov/hmcts/reform/fpl/events/CallbackEvent.java
joez-f: It's techdebt, maybe I should make it clearer in the comment 👍

service/src/test/resources/core-case-data-store-api/callback-request.json
```diff
@@ -67,7 +67,7 @@
           "childInformation": "Cousin",
           "litigationIssues": "NO",
           "genderIdentification": "Unknown",
-          "representedBy" : []
+          "representedBy": []
```
swalker125: err, did I not remove this in my PR? Know @Jamesnelson117 commented. Can't remember outcome

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: Notification is same for CMO/notice of placement/create an order except for the callout which is only on CMO, so I used same gov notify template and then add callout here if it is CMO (cannot do conditional logic + placeholder variable in gov notify otherwise I would do it there)

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
joez-f: https://tools.hmcts.net/jira/browse/FPLA-1287 comment isn't really necessary so I will delete it

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: I think this is fine, there is a similar method for LA, also documentContents is tied to the event rather than EventData so that would still need passing in separately

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerTest.java
```diff
@@ -306,6 +314,13 @@ void submittedShouldTriggerCMOProgressionAndSendNotificationsToOnlyLocalAuthorit
             eq(CMO_ORDER_ISSUED_DOCUMENT_LINK_NOTIFICATION_TEMPLATE), eq(CAFCASS_EMAIL_ADDRESS),
             anyMap(), eq(CASE_ID));
 
+        verify(notificationClient).sendEmail(
+            eq(ORDER_NOTIFICATION_TEMPLATE_FOR_ADMIN), eq("admin@family-court.com"),
+            dataCaptor.capture(), eq(CASE_ID));
```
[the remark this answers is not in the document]
joez-f: Might look inconsistent with above verify in same method that I haven't touched, happy to update whole method though

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
```diff
@@ -125,20 +144,63 @@ void shouldReturnErrorsWhenFamilymanNumberIsNotProvided() {
         assertThat(callbackResponse.getErrors()).containsExactly("Enter Familyman case number");
     }
 
-    @Test
-    void shouldTriggerOrderEventWhenSubmitted() throws Exception {
-        String expectedCaseReference = "19898989";
-        postSubmittedEvent(buildCallbackRequest());
+    @Nested
+    class Submitted {
```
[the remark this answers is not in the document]
joez-f: I'd want to split the whole class out at once rather than just take one type of callback out now and leave the rest - I can either add techdebt note or go ahead and split whole class (and I should do same for ActionCaseManagementOrderControllerTest in that case)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
joez-f: verifyNotificationSentToAdminWhenOrderIssued returns MapDifference between the actual data and the expected parameters (all found in NotifyAdminOrderIssuedTestHelper), so here we check that there's no difference (i.e we got what we expected) - nothing to do with GENERATED_ORDER, this is an enum I pass in so that I know if it's a CMO or not (cmo needs callout)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
joez-f: 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/NotifyTemplates.java
joez-f: order is issued - will update name 👍

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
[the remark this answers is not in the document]
swalker125: no please don't follow wrong practice. Separate new code out, write comment for rest of code.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerTest.java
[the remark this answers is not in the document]
swalker125: new code - make right.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
joez-f: Not sure, I think @Gogetter made this one a while back, I'm guessing it's extracted to catch the URISyntaxException, happy to change it and do it within handleSubmittedEvent without extracting

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: I didn't resolve this one, I left it open as I was waiting for response 🙂

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
joez-f: will revert 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: thought it might help people looking at gov notify template in future to know that the callout is only shown for CMO (since same template handles 3 types of order). think there is benefit

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
joez-f: I'm happy keeping it here given it is not very complex

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: extracted to private method

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: compared separately in line above, Maps.difference will return false if not removed as these are JSON objects that will be compared by reference not value so will never be true on .equals

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: tried, didnt work, I think based on my above comment to do with comparing JSONObjects

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: why is this preferred? do you mean like this:
```
        Map<String, Object> expectedMap = new HashMap<>();
        expectedMap.put("cmoCallout", "");
        expectedMap.putAll(commonParametersNoPostingNeeded());
        return expectedMap;
```

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: nope 👍 you're right

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
joez-f: inlined callout

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerTest.java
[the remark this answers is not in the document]
joez-f: okay, will split out 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
Gogetter: @joez-f the method prefixes the gateway url (does not require user to login I think) with the path from `mostRecentUploadedDocumentUrl` (which is doc store and requires login). This given we are sending notification to people with no case access.

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f:
```
[ant:checkstyle] [WARN] reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java:25:1: Utility classes should not have a public or default constructor. [HideUtilityClassConstructor]
```
nevermind, it's required

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: No nicer way of doing this as far as I'm aware, @lewisbirks agrees

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: needs to be mutable as we will remove caseUrlOrDocumentLink before we assert the other fields

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerSubmittedTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Captor
+    private ArgumentCaptor<Map<String, Object>> dataCaptor;
+
+    GeneratedOrderControllerSubmittedTest() {
+        super("create-order");
+    }
+
+    @Nested
+    class Submitted {
```
joez-f: will remove

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
joez-f: lombok annotation makes public constructors, need private one here

service/src/test/resources/core-case-data-store-api/callback-request.json
joez-f: 🤷‍♂

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
lewisbirks: Could use `String.format(...)`, might make it look a bit nicer

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/NotifyAdminOrderIssuedTestHelper.java
lewisbirks: Here `Map.of()` could be used, in the other [situation](https://github.com/hmcts/fpl-ccd-configuration/pull/870/files/c17124d9faaa9df50e8820757420c7d7fbe14f3e#r379378335) where you are adding the contents of another map then `Map.of()` cannot be used

joez-f: yep I just realised too, changed 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/PeopleInCaseHelper.java
joez-f: changed, although personally don't think it looks much nicer

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementSubmittedEventControllerTest.java
```diff
@@ -184,26 +208,27 @@ void shouldNotSendEmailNotificationWhenNoChangesToOrder() throws NotificationCli
...
                 eq(expectedParameters()),
-                eq("1"));
+                eq(CASE_ID));
         }
 
         private Map<String, Object> expectedParameters() {
             return Map.of(
-                "respondentLastName", "Nelson",
-                "caseUrl", String.format("%s/case/%s/%s/%s", "http://fake-url", JURISDICTION, CASE_TYPE, 1L));
+                "respondentLastName", "Jones",
```
swalker125: err why Jones -> Nelson?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandler.java
swalker125: hmm I think this makes sense though, you then have less params in your method

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementSubmittedEventControllerTest.java
joez-f: NotifyAdminOrderIssuedTestHelper expects respondentLastName to be Jones, and since it is used in 3 controller tests I've changed test data in controllers where it wasn't consistent (so that I can cleanly utilise my test helper)

joez-f: justification is that 3 controllers trigger the same notification so it made sense to centralise the code used for testing it, so that I can call it from 3 different places rather than duplicate it 3 times

PR #881, opened 2020-02-14
fix/feat: add hearing date to draft SDO screens (FPLA-1002)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1002

### Change description ###

New text fields for each page in SDO journey. Populate in DraftOrdersController AboutToStart

```
[ ] Yes
[x] No
```


swalker125:
Screen Shot 2020-02-14 at 16 54 58

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -64,6 +69,14 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
         CaseDetails caseDetails = callbackrequest.getCaseDetails();
         CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
 
+        List<Element<HearingBooking>> element = defaultIfNull(
+            caseData.getHearingDetails(), wrapElements(HearingBooking.builder().build()));
+
+        String hearingDate = getHearingDate(element);
+
+        Stream.of(DirectionAssignee.values()).forEach(assignee ->
```
lewisbirks: Could you use the same field across multiple pages (think not but can't remember)?

ccd-definition/CaseField.json
```diff
@@ -426,6 +426,55 @@
...
+    "ID": "allPartiesHearingDate",
+    "Label": "The next hearing: ",
+    "FieldType": "Text",
+    "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "localAuthorityDirectionsHearingDate",
+    "Label": "The next hearing: ",
```
lewisbirks: Not sure about the `: `

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -78,6 +91,17 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
             .build();
     }
 
+    private String getHearingDate(List<Element<HearingBooking>> element) {
+        String date;
+
+        if (element.get(0).getValue().getStartDate() != null) {
+            date = formatLocalDateTimeBaseUsingFormat(element.get(0).getValue().getStartDate(), "d MMMM yyyy, h:mma");
```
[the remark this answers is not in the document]
joez-f: support 👍 ⬆️

ccd-definition/CaseField.json
```diff
@@ -426,6 +426,55 @@
     "ID": "hearingDate",
     "Label": "The next hearing is on ${hearingDetails.startDate}.",
     "FieldType": "Label",
+    "SecurityClassification": "Public",
+    "Comment": "safe to be removed?"
```
joez-f: from what I understand this isn't shown anywhere at the moment? but would old cases with this case field still break?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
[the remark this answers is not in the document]
joez-f: also is setting the 'date' variable better than having two return points? I'd prefer  if ... return ... else return ... but just my preference

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -64,6 +69,14 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
         CaseDetails caseDetails = callbackrequest.getCaseDetails();
         CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
 
+        List<Element<HearingBooking>> element = defaultIfNull(
```
tomaszpowroznik: hearings instead of element

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -64,6 +69,14 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
...
+            caseData.getHearingDetails(), wrapElements(HearingBooking.builder().build()));
+
+        String hearingDate = getHearingDate(element);
```
tomaszpowroznik: String nextHearingDate = getNextHearingDate(hearings)

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -78,6 +91,17 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
             .build();
     }
 
+    private String getHearingDate(List<Element<HearingBooking>> element) {
+        String date;
+
+        if (element.get(0).getValue().getStartDate() != null) {
```
tomaszpowroznik: can we be sure that collection is properly ordered, I mean that first element is before second? Is it expected behaviour when hearing already happened (first element is in past)? 
Maybe existing HearingBookingService is better place for this method.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerTest.java
```diff
@@ -117,6 +106,39 @@ void aboutToStartCallbackShouldSplitDirectionsIntoSeparateCollections() {
...
+            .build();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails);
+
+        Stream.of(DirectionAssignee.values()).forEach(assignee ->
+            assertThat(callbackResponse.getData().get(assignee.toHearingDateField()))
+                .isEqualTo("1 January 2020, 12:00am"));
+    }
+
+    private List<Direction> createDirections() {
```
tomaszpowroznik: usually we put private methods at the bottom of class, can be static

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DateFormatterService.java
```diff
@@ -18,7 +18,7 @@ public String formatLocalDateToString(LocalDate date, String format) {
         return date.format(DateTimeFormatter.ofPattern(format, Locale.UK));
     }
 
-    public String formatLocalDateTimeBaseUsingFormat(LocalDateTime dateTime, String format) {
+    public static String formatLocalDateTimeBaseUsingFormat(LocalDateTime dateTime, String format) {
```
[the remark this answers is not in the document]
swalker125: 🤞

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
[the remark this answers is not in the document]
swalker125: hmm, @lewisbirks happy to make this a constant accessible from the service. I will not be changing other places it used however. As for return @joez-f, I see no real difference either way

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
swalker125: no, you get a unique constraint violation exception 😕

ccd-definition/CaseField.json
swalker125: let me loop in Collette

ccd-definition/CaseField.json
swalker125: Yeah this is my question. If this gets populated in anyway, then when I remove from the definition, it COULD error. I will ask CCD

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
swalker125: hmm I am not sure getNextHearing is right either. The method gets the first hearing in the collection

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
swalker125: hmm, so collection ordering I cannot guarantee. I guess this method should be refactored to get the earliest hearing. The first hearing is ALWAYS for the SDO. But you're right, just saying the first element could break things if the user was to use the collection screen differently to what I expect. I shall refactor.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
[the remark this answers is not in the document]
lewisbirks: Yeah, I understand not changing it else where, that is a rather large task

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
lewisbirks: Thought so 😢

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
swalker125: getMostUrgentHearingBooking seems to be doing this 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
swalker125: I have called getFirstHearingStartDate

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -78,6 +89,19 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
...
 
+    private String getFirstHearingStartDate(List<Element<HearingBooking>> hearings) {
+        String hearingDate;
+
+        try {
+            LocalDateTime startDate = hearingBookingService.getMostUrgentHearingBooking(hearings).getStartDate();
+            hearingDate = formatLocalDateTimeBaseUsingFormat(startDate, FULL_DATE);
+        } catch (IllegalStateException e) {
+            hearingDate = "Please enter a hearing date";
+        }
```
tomaszpowroznik: ideally service should return Optional and client should decide how to handle this

PR #883, opened 2020-02-17
feat: display amount to pay (FPLA-1301)

lewisbirks:

[FPLA-1301](https://tools.hmcts.net/jira/browse/FPLA-1301)

### Change description ###

Added logic for interacting with the fee register, filter fees based on rules, and display the appropriate fee.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/config/payment/FeesConfig.java
```diff
@@ -0,0 +1,80 @@
...
+                      @Value("${fee.types.interim_supervision_order}") String interimSupervisionOrder,
+                      @Value("${fee.types.other}") String other,
+                      @Value("${fee.types.placement}") String placement,
+                      @Value("${fee.types.supervision_order}") String superVisionOrder) {
+
+        this.orderFeeParameters = new HashMap<>();
+        this.url = url;
+        this.api = api;
+
+        addFeeParameters(C2_WITH_NOTICE, parseStringValue(c2WithNotice));
```
akomisarek: Personally I think it's necessary too complex, couldn't we have got complex type in Yaml without all of this parsing? would read a lot nicer, and afaik you can have map automatically populated by spring https://mkyong.com/spring-boot/spring-boot-configurationproperties-example/

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -49,39 +56,31 @@
...
         @RequestBody CallbackRequest callbackRequest) {
+
         CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        Map<String, Object> data = caseDetails.getData();
+        CaseData caseData = mapper.convertValue(data, CaseData.class);
+
+        List<FeeResponse> fees = feeService.getFees(FeeType.fromOrderType(caseData.getOrders().getOrderType()));
+        BigDecimal amount = feeService.extractFeeToUse(fees)
+            .map(FeeResponse::getAmount)
+            .orElse(BigDecimal.ZERO);
```
akomisarek: Is it correct handling? might be worth checking with the business, what IF fees is broken and we can't query? how we should behave?

service/src/main/java/uk/gov/hmcts/reform/fpl/config/payment/FeesConfig.java
lewisbirks: Will look into it, hopefully it will be a lot nicer if it works

service/src/main/java/uk/gov/hmcts/reform/fpl/model/payment/fee/FeeParameters.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts.reform.fpl.model.payment.fee;
```
akomisarek: Small comment, I don't like it leaving in fpl.model it's not really FPL model is it?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/payment/fee/FeeResponse.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.reform.fpl.model.payment.fee;
```
akomisarek: Same not fpl model

service/src/main/java/uk/gov/hmcts/reform/fpl/service/payment/FeeService.java
```diff
@@ -0,0 +1,88 @@
...
+
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+import static org.springframework.web.util.UriComponentsBuilder.fromHttpUrl;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class FeeService {
+
+    private final RestTemplate restTemplate;
```
akomisarek: Couldn't you use @Feign here? the url building seems a bit ugly, please check OrganisationApi. Don't mind if it works but I think Feign is a bit nicer (but there are ea bit issues with conf)

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/BigDecimalHelper.java
```diff
@@ -0,0 +1,34 @@
...
+        BigDecimal pence = amount.multiply(HUNDRED);
+        return String.valueOf(pence.longValue());
+    }
+
+    public static BigDecimal fromCCDMoneyGBP(String amount) {
+        try {
+            int integer = Integer.parseInt(amount);
+            return BigDecimal.valueOf(integer, 2);
+        } catch (NumberFormatException ex) {
+            log.error("couldn't convert \"{}\" to int", amount);
```
akomisarek: That potentially should return optional instead of null. Always better practice

service/src/main/resources/application.yaml
```diff
@@ -209,6 +209,22 @@ fpl:
     username: 'fpl-system-update@mailnesia.com'
     password: 'Password12'
 
+fee:
+  url: 'http://fees-register-api-aat.service.core-compute-aat.internal'
+  api: '/fees-register/fees/lookup'
+  types:
+    # dummy data, to be updated
```
akomisarek: As advised somewhere else, probably that would be much better to be strongly typed data.

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/payment/FeeType.java
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        return orderTypes.stream()
+            .map(FeeType::fromOrderType)
+            .collect(toUnmodifiableList());
+    }
+
+    public static FeeType fromOrderType(OrderType orderType) {
+        return FeeType.valueOf(orderType.name());
+    }
```
akomisarek: Seems not ideal to tight so much, what if we have got not matching fields? we are not handling the errors properly, is it a problem? Maybe not?

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
```diff
@@ -0,0 +1,52 @@
...
+            return ImmutableList.of();
+        }
+
+        return orderTypes.stream()
+            .map(FeeType::fromOrderType)
+            .collect(toUnmodifiableList());
+    }
+
+    public static FeeType fromOrderType(OrderType orderType) {
+        switch (orderType) {
```
pd-dudkiewicz: I think it'd be cleaner to add `Map<OrderType, FeeType>` and access it here instead of using switch

service/src/main/java/uk/gov/hmcts/reform/fpl/service/payment/FeeService.java
```diff
@@ -0,0 +1,76 @@
...
+            .filter(Objects::nonNull)
+            .max(Comparator.comparing(FeeResponse::getAmount));
+    }
+
+    public List<FeeResponse> getFees(List<FeeType> feeTypes) {
+        if (feeTypes == null) {
+            return ImmutableList.of();
+        }
+
+        return feeTypes.stream()
```
tomaszpowroznik: maybe parallel stream can be used here to make remote calls in parallel

service/src/main/java/uk/gov/hmcts/reform/fpl/config/payment/FeesConfig.java
```diff
@@ -0,0 +1,80 @@
...
+                      @Value("${fee.types.c2_with_notice}") String c2WithNotice,
+                      @Value("${fee.types.c2_without_notice}") String c2WithoutNotice,
+                      @Value("${fee.types.care_order}") String careOrder,
+                      @Value("${fee.types.education_supervision_order}") String educationSupervisionOrder,
+                      @Value("${fee.types.emergency_protection_order}") String emergencyProtectionOrder,
+                      @Value("${fee.types.interim_care_order}") String interimCareOrder,
+                      @Value("${fee.types.interim_supervision_order}") String interimSupervisionOrder,
+                      @Value("${fee.types.other}") String other,
+                      @Value("${fee.types.placement}") String placement,
+                      @Value("${fee.types.supervision_order}") String superVisionOrder) {
```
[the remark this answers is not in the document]
lewisbirks: No longer needed as we are directly importing the map from the application.yaml

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/FeeServiceTest.java
```diff
@@ -0,0 +1,99 @@
...
+        @ParameterizedTest
+        @NullAndEmptySource
+        void shouldReturnAnEmptyListWhenNullOrEmptyListIsPassed(List<FeeType> list) {
+            assertThat(feeService.getFees(list)).isEmpty();
+        }
+
+        @ParameterizedTest
+        @EnumSource(FeeType.class)
+        void shouldGetAFeeResponseWhenGivenAPopulated(FeeType feeType) {
+            // TODO: 17/02/2020 mock something
```
akomisarek: Let's complete test:D

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
```diff
@@ -0,0 +1,52 @@
...
+        if (orderTypes == null) {
+            return ImmutableList.of();
+        }
+
+        return orderTypes.stream()
+            .map(FeeType::fromOrderType)
+            .collect(toUnmodifiableList());
+    }
+
+    public static FeeType fromOrderType(OrderType orderType) {
```
akomisarek: Could we have unit test iterating on all values of OrderType to ensure we added mapping him? I prefer failure if someone forgets about something than IllegalStateException later in the runtime.

docker/wiremock/mappings/fee-register/c2-without-notice.json
```diff
@@ -0,0 +1,33 @@
...
+        "equalTo": "general application"
+      },
+      "jurisdiction1": {
+        "equalTo": "family"
+      },
+      "jurisdiction2": {
+        "equalTo": "family court"
+      },
+      "keyword": {
+        "equalTo": "without-notice"
```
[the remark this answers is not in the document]
lewisbirks: nah, for c2 they are that long

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static Map<OrderType, FeeType> buildOrderFeeMap() {
+        return Map.of(
+            OrderType.CARE_ORDER, CARE_ORDER,
+            OrderType.EDUCATION_SUPERVISION_ORDER, EDUCATION_SUPERVISION_ORDER,
+            OrderType.EMERGENCY_PROTECTION_ORDER, EMERGENCY_PROTECTION_ORDER,
+            OrderType.INTERIM_CARE_ORDER, INTERIM_CARE_ORDER,
+            OrderType.INTERIM_SUPERVISION_ORDER, INTERIM_SUPERVISION_ORDER,
+            OrderType.SUPERVISION_ORDER, SUPERVISION_ORDER,
+            OrderType.OTHER, OTHER
```
akomisarek: Hmm actually if the names are the same couldn't we loop here?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -49,51 +58,59 @@
...
-        String label = String.format(CONSENT_TEMPLATE, userDetailsService.getUserName(authorization));
+        List<String> errors = validate(caseData);
+        BigDecimal amount = null;
 
-        Map<String, Object> data = caseDetails.getData();
-        data.put("submissionConsentLabel", label);
+        try {
+            amount = getAmountToPay(caseData.getOrders());
+        } catch (FeignException ex) {
+            // TODO: 21/02/2020 Replace me in a new story
```
akomisarek: Is this story created?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -49,51 +58,59 @@
...
+        CaseData caseData = mapper.convertValue(data, CaseData.class);
 
-        String label = String.format(CONSENT_TEMPLATE, userDetailsService.getUserName(authorization));
+        List<String> errors = validate(caseData);
+        BigDecimal amount = null;
 
-        Map<String, Object> data = caseDetails.getData();
-        data.put("submissionConsentLabel", label);
+        try {
+            amount = getAmountToPay(caseData.getOrders());
```
akomisarek: Shouldnn't we short circuit logic if the validation errors are on case? I don't see why we query Fee register in case of error, we are not adding it to the result anyway. Please consider changing it.

service/src/test/java/uk/gov/hmcts/reform/fpl/testbeans/TestFeeConfig.java
```diff
@@ -0,0 +1,38 @@
...
+            "family",
+            "family court",
+            "KLM",
+            "private law"
+        );
+
+        FeesConfig config = new FeesConfig();
+
+        config.setParameters(Arrays.stream(FeeType.values())
+            .collect(collectingAndThen(toMap(feeType -> feeType, ignored -> parameters), ImmutableMap::copyOf)));
```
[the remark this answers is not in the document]
akomisarek: Not anymore, that works today.

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
lewisbirks: Would that then essentially be a replication of https://github.com/hmcts/fpl-ccd-configuration/pull/883#discussion_r380146843?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
lewisbirks: https://tools.hmcts.net/jira/browse/FPLA-1353 this is at the bottom of the backlog

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/FeeServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+        void shouldGetAFeeResponseWhenGivenAPopulated() {
+            FeeResponse expectedFeeResponse = expectedFeeResponse();
+
+            when(feesRegisterApi.findFee("default", "miscellaneous", "family", "family court", "KLM", "private law"))
+                .thenReturn(expectedFeeResponse);
+
+            List<FeeResponse> fees = feeService.getFees(List.of(CARE_ORDER, SUPERVISION_ORDER, PLACEMENT));
+
+            assertThat(fees).hasSize(3);
+            assertThat(fees).containsOnly(expectedFeeResponse);
```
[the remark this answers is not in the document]
lewisbirks: Due to how it is being mocked, the same object is being returned each time, containsOnly accounts for duplications in the collection so you only have to pass the parameter once

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/BigDecimalHelper.java
```diff
@@ -0,0 +1,35 @@
...
+public class BigDecimalHelper {
+
+    private static final BigDecimal HUNDRED = BigDecimal.valueOf(100);
+
+    private BigDecimalHelper() {
+
+    }
+
+    public static String toCCDMoneyGBP(BigDecimal amount) {
+        if (amount == null) {
```
pd-dudkiewicz:
```
return Optional.of(amount)
    .map(a -> a.multiply(HUNDRED))
    .map(BigDecimal::longValue)
    .map(String::valueOf)
    .orElse(StringUtils.EMPTY);
```

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/BigDecimalHelper.java
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        BigDecimal pence = amount.multiply(HUNDRED);
+        return String.valueOf(pence.longValue());
+    }
+
+    public static Optional<BigDecimal> fromCCDMoneyGBP(String amount) {
+        try {
+            int integer = Integer.parseInt(amount);
+            return Optional.of(BigDecimal.valueOf(integer, 2));
```
pd-dudkiewicz: What's the reason we don't use `BigDecimal` constructor?

lewisbirks: `valueOf` and the constructor do slightly different things, `valueOf` uses the string representation of the passed value where as the constructor uses the passed primitive as accurately as possible.
See [javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/math/BigDecimal.html#%3Cinit%3E(double)) and [stack overflow post](https://stackoverflow.com/a/7186298)

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeResponse.java
```diff
@@ -0,0 +1,17 @@
...
+import java.math.BigDecimal;
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class FeeResponse {
+    private String code;
+    private String description;
+    private Integer version;
+    @JsonProperty(value = "fee_amount")
+    private BigDecimal amount;
```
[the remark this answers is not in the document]
lewisbirks: Can do, was thinking that that name was a bit repetitive as you are already in the fee object so it will obviously be the amount for the fee

service/src/main/java/uk/gov/hmcts/reform/fnp/client/FeesRegisterApi.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import uk.gov.hmcts.reform.fnp.model.fee.FeeResponse;
+import uk.gov.hmcts.reform.fpl.config.FeignConfiguration;
+
+@FeignClient(name = "fees-register-api", url = "${fee.api.url}", configuration = FeignConfiguration.class)
+public interface FeesRegisterApi {
+    @GetMapping("/fees-register/fees/lookup")
+    FeeResponse findFee(
+        @RequestParam(name = "channel") final String channel,
```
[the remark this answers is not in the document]
lewisbirks: Not that I can think of, can change if you want

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -49,51 +58,59 @@
...
 
         return AboutToStartOrSubmitCallbackResponse.builder()
             .data(data)
-            .errors(validate(mapper.convertValue(data, CaseData.class)))
+            .errors(errors)
             .build();
     }
 
+    private BigDecimal getAmountToPay(Orders orders) throws FeignException {
+        List<FeeResponse> fees = feeService.getFees(FeeType.fromOrderType(
```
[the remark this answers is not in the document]
lewisbirks: Yeah, was wondering if there was too much logic in the controller

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeResponse.java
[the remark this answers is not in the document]
pd-dudkiewicz: Good point, my idea was to make it less noisy without the annotation and having it represent response one-to-one everywhere it's used. Happy to leave it as it is though, so feel free to resolve. :+1:

service/src/main/java/uk/gov/hmcts/reform/fpl/service/payment/FeeService.java
```diff
@@ -0,0 +1,74 @@
...
+            .max(Comparator.comparing(FeeResponse::getAmount));
+    }
+
+    public List<FeeResponse> getFees(List<FeeType> feeTypes) throws FeignException {
+        if (feeTypes == null) {
+            return ImmutableList.of();
+        }
+
+        return feeTypes.stream()
+            .map(this::makeRequest)
```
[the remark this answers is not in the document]
lewisbirks: Don't think so, would be more of an issue that the controller should work around I think. Not sure exactly how it would be handled here rather than the controller handling it to know that it should populate an error

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/BigDecimalHelper.java
lewisbirks: merged the first 2 maps as well with a private method

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
```diff
@@ -0,0 +1,53 @@
...
+    INTERIM_CARE_ORDER,
+    INTERIM_SUPERVISION_ORDER,
+    OTHER,
+    PLACEMENT,
+    SUPERVISION_ORDER;
+
+    private static final Map<OrderType, FeeType> orderToFeeMap = buildOrderFeeMap();
+
+    public static FeeType fromOrderType(OrderType orderType) {
+        return orderType != null ? orderToFeeMap.get(orderType) : null;
```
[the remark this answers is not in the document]
lewisbirks: Shouldn't be, currently it is only called from the overloaded method, and that list that is passed to it shouldn't have any nulls in it. Do you think I should remove the check? Could add `@Nonnull` to make the requirement more specific.

e2e/tests/localAuthoritySubmitsApplication_test.js
```diff
@@ -475,6 +475,7 @@ Scenario('local authority tries to submit without giving consent', async (I, cas
 
 Scenario('local authority submits after giving consent @create-case-with-mandatory-sections-only', async (I, caseViewPage, submitApplicationEventPage) => {
   await caseViewPage.goToNewActions(config.applicationActions.submitCase);
+  I.see('£2,055.00');
```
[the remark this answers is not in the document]
lewisbirks: Don't really think it is necessary

service/src/main/java/uk/gov/hmcts/reform/fnp/client/FeesRegisterApi.java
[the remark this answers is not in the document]
lewisbirks: Changed from final

service/src/main/java/uk/gov/hmcts/reform/fpl/service/payment/FeeService.java
```diff
@@ -0,0 +1,74 @@
...
+                parameters.getService()
+            );
+
+            log.debug("Fee response: {} ", fee);
+
+            return fee;
+        } catch (FeignException ex) {
+            log.error("Fee response error: {} => body: \"{}\"",
+                ex.status(), ex.getMessage());
+            throw ex;
```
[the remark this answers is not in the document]
lewisbirks: Name isn't very imaginative but done (could possibly be extended for different error types like the `FeignException` which might give better names)

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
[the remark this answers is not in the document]
lewisbirks: moved to `getFeeAmountForOrders` in the service

service/src/main/java/uk/gov/hmcts/reform/fnp/client/FeesRegisterApi.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.hmcts.reform.fnp.client;
+
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import uk.gov.hmcts.reform.fnp.model.fee.FeeResponse;
+import uk.gov.hmcts.reform.fpl.config.FeignConfiguration;
+
+@FeignClient(name = "fees-register-api", url = "${fee.api.url}", configuration = FeignConfiguration.class)
```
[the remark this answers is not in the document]
akomisarek: Is this one being worked on then?

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
akomisarek: Hah sorry :D, yeah I am happy with that it's quite explicit. Let's leave it.

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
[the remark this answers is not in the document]
pd-dudkiewicz: If we cannot come up with valid scenario when `orderType` is null, I'd suggest to remove the check. We want our app to blow up with NPE, because it means that something is seriously wrong.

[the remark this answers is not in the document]
pd-dudkiewicz: Also, just noticed that this method is not used anywhere (apart from unit test), so we should probably remove it.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/BigDecimalHelper.java
```diff
@@ -15,21 +16,27 @@ private BigDecimalHelper() {
...
         }
     }
+
+    private static Long toPence(BigDecimal amount) {
+        return amount.multiply(HUNDRED).longValue();
+    }
+
+    private static BigDecimal fromPence(String amount) {
+        long pence = Long.parseLong(amount);
+        return BigDecimal.valueOf(pence, 2);
```
pd-dudkiewicz: I'd suggest to extract `2` to const and give it some descriptive name.

lewisbirks: Extracted to `SCALE`

service/src/main/java/uk/gov/hmcts/reform/fnp/model/fee/FeeType.java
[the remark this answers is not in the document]
lewisbirks: Good point, removed method and altered tests

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -74,7 +70,9 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
 
         if (errors.isEmpty()) {
             try {
-                BigDecimal amount = getAmountToPay(caseData.getOrders());
+                List<OrderType> orderTypes = caseData.getOrders() == null ? null : caseData.getOrders().getOrderType();
```
pd-dudkiewicz: We could do `Optional.ofNullable(caseData.getOrders()).map(Orders::getOrderType)` here and pass it to `getFeeAmountForOrders` method, where `ofNullable` could be removed.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/FeeServiceTest.java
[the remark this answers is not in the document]
akomisarek: ok I see so what if we add couple more elements to config? I am afraid now that you can essentially hardcode and the unit tests will pass or not? :P

service/src/main/java/uk/gov/hmcts/reform/fpl/service/payment/FeeService.java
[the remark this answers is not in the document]
pd-dudkiewicz: Fair enough :+1: resolving

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/BigDecimalHelper.java
pd-dudkiewicz: Thanks for explanation

service/src/main/java/uk/gov/hmcts/reform/fnp/exception/FeeRegisterException.java
```diff
@@ -0,0 +1,31 @@
...
+        super();
+        this.status = status;
+    }
+
+    public FeeRegisterException(int status, String message) {
+        super(message);
+        this.status = status;
+    }
+
+    public FeeRegisterException(int status, String message, Throwable cause) {
```
pd-dudkiewicz: This is the only constructor used, do we need others?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/FeeServiceTest.java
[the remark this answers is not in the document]
lewisbirks: Was my worry as well, will expand the config for new values (probably just change the keyword)

service/src/main/java/uk/gov/hmcts/reform/fnp/exception/FeeRegisterException.java
lewisbirks: Was just overloading the default exception methods, can remove the others if you want

pd-dudkiewicz:
Remember to add `FEES_REGISTER_API_URL` in chart and bump its version. And `cnp-flux-config` will need to be updated once we got this PR merged.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
pd-dudkiewicz: nevermind, did my homework, it's not a good practice to use Optionals as method params 😄 resolving

service/src/main/java/uk/gov/hmcts/reform/fnp/exception/FeeRegisterException.java
pd-dudkiewicz: I'd remove that if it's not used.

pd-dudkiewicz:
was `submitApplication.json ` move intentional?

lewisbirks:
> was `submitApplication.json ` move intentional?

Yeah, wanted to keep the base and nonprod files grouped for easy access

PR #910, opened 2020-02-26
feat: auto populate applicant details (FPLA-1103)

toireasam:

### Change description ###

```
[ ] Yes
[ ] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -1,21 +1,51 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.when;
 
 @ActiveProfiles("integration-test")
 @WebMvcTest(ApplicantController.class)
 @OverrideAutoConfiguration(enabled = true)
 class ApplicantAboutToStartControllerTest extends AbstractControllerTest {
 
+    private final String serviceAuthToken = RandomStringUtils.randomAlphanumeric(10);
```
swalker125: this seems like overkill, but if you're fixed on it, let's make it static and lets have a static import of randomAlphanumeric

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -1,21 +1,51 @@
...
+    final String userAuthToken = "Bearer token";
```
swalker125: this can be private and static also

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +56,94 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+        CaseData data = mapper.convertValue(callbackResponse.getData(), CaseData.class);
+
+        assertThat(data.getAllApplicants()).contains(buildApplicant());
+    }
+
+    @Test
+    void shouldFindOrganisation() {
+        Organisation organisation = buildOrganisation();
+        given(authTokenGenerator.generate()).willReturn(serviceAuthToken);
+        given(organisationApi.findOrganisationById(userAuthToken, serviceAuthToken)).willReturn(organisation);
```
swalker125: this is defined in the before each

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +56,94 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+
+        Organisation actualOrganisation = organisationApi.findOrganisationById(userAuthToken, serviceAuthToken);
+
+        assertThat(actualOrganisation).isEqualTo(organisation);
+    }
+
+    @Test
+    public void shouldThrowExceptionWhenOrganisationNotFound() {
+        Exception exception = new FeignException.NotFound("", REQUEST, new byte[]{});
```
swalker125: hmm this test seems odd. You're throwing a specific type of feign exception but asserting a generic one, is this correct? 

I see no point extracting exception to a variable.

REQUEST isn't a very helpful variable name

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +56,94 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+
+        assertThatThrownBy(() -> organisationApi.findOrganisationById(userAuthToken, serviceAuthToken))
+            .isInstanceOf(FeignException.class);
+
+    }
+
+    private Organisation buildOrganisation() {
+        return Organisation.builder().name("Organisation")
+        .contactInformation(buildOrganisationContactInformation())
+            .build();
```
swalker125: is the spacing correct here? Maybe we should space down .name()

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +56,94 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+        ArrayList<ContactInformation> contactInformation = new ArrayList<>();
+        contactInformation.add(ContactInformation.builder()
+            .addressLine1("Flat 12, Pinnacle Apartments")
+            .addressLine1("Saffron Central")
+            .county("London")
+            .country("United Kingdom")
+            .postCode("CR0 2GE")
+            .build());
+
+        return  contactInformation;
```
swalker125: return newArrayList.of() can be used here. It's a static method in Lists :)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +56,94 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+
+    private Element<Applicant> buildApplicant() {
+        return Element.<Applicant>builder()
+            .value(Applicant.builder()
+                .party(ApplicantParty.builder()
+            .organisationName("Organisation")
+                    .address(buildApplicantContactInformation())
+                    .build())
+                .build())
+            .build();
```
swalker125: you can use ElementUtils.element(Applicant.builder()...) static method here. Static import

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantController.java
```diff
@@ -1,43 +1,68 @@
...
+                               ObjectMapper mapper,
+                               OrganisationApi organisationApi,
+                               AuthTokenGenerator authTokenGenerator,
+                               RequestData requestData) {
         this.applicantService = applicantService;
         this.updateAndValidatePbaService = updateAndValidatePbaService;
         this.mapper = mapper;
+        this.organisationApi = organisationApi;
+        this.authTokenGenerator = authTokenGenerator;
+        this.requestData = requestData;
```
swalker125: could use @RequiredArgsConstructor(onConstructor = @__(@ Autowired)) to remove the need for the constructor here

swalker125: remove space between @ and A, just didn't want to tag someone haha

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantController.java
```diff
@@ -1,43 +1,68 @@
...
     }
 
     @PostMapping("/about-to-start")
     public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody CallbackRequest callbackrequest) {
         CaseDetails caseDetails = callbackrequest.getCaseDetails();
         CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
 
-        caseDetails.getData().put("applicants", applicantService.expandApplicantCollection(caseData));
+        Organisation organisation = Organisation.builder().build();
```
swalker125: could you add the builder into that catch block? Makes it clear where and why this is being done

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
```diff
@@ -3,37 +3,73 @@
...
-    public List<Element<Applicant>> expandApplicantCollection(CaseData caseData) {
+    public List<Element<Applicant>> expandApplicantCollection(CaseData caseData, Organisation organisation) {
         if (isEmpty(caseData.getApplicants())) {
-            return ImmutableList.of(Element.<Applicant>builder()
-                .value(Applicant.builder()
-                    .party(ApplicantParty.builder()
-                        // A value within applicant party needs to be set in order to expand UI view.
-                        .partyId(UUID.randomUUID().toString())
+            if (isEmpty(organisation)) {
+                return ImmutableList.of(Element.<Applicant>builder()
```
swalker125: this can be return List.of(element(Applicant.builder()..... .build()))

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
```diff
@@ -3,37 +3,73 @@
...
+                    .value(Applicant.builder()
+                        .party(ApplicantParty.builder()
+                            // A value within applicant party needs to be set in order to expand UI view.
+                            .partyId(UUID.randomUUID().toString())
```
swalker125: static import

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
```diff
@@ -3,37 +3,73 @@
...
 
-    public List<Element<Applicant>> expandApplicantCollection(CaseData caseData) {
+    public List<Element<Applicant>> expandApplicantCollection(CaseData caseData, Organisation organisation) {
         if (isEmpty(caseData.getApplicants())) {
-            return ImmutableList.of(Element.<Applicant>builder()
-                .value(Applicant.builder()
-                    .party(ApplicantParty.builder()
-                        // A value within applicant party needs to be set in order to expand UI view.
-                        .partyId(UUID.randomUUID().toString())
+            if (isEmpty(organisation)) {
```
swalker125: is this going to catch what you expect? We are always passing an organisation here due to the builder

swalker125: could we actually be clever here and always add organisation to applicant, wrap the values of organisation in optionals sort of thing? Removes the need for buildApplicantWithOrganisationDetails method, and the need for this isEmpty method

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
```diff
@@ -3,37 +3,73 @@
...
+        return Address.builder()
+            .addressLine1(contactInformation.getAddressLine1())
+            .addressLine2(contactInformation.getAddressLine2())
+            .addressLine3(contactInformation.getAddressLine3())
+            .county(contactInformation.getCounty())
+            .country(contactInformation.getCountry())
+            .postcode(contactInformation.getPostCode())
+            .postTown(contactInformation.getTownCity())
+            .build();
+    }
```
swalker125: can we not create method in ContactInformation called toAddress or buildAddress, which returns Address object?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
```diff
@@ -47,8 +51,9 @@ void shouldNotExpandApplicantCollectionWhenApplicantsAlreadyExists() {
                 .build()))
             .build();
 
-        assertThat(service.expandApplicantCollection(caseData)).hasSize(1);
-        assertThat(service.expandApplicantCollection(caseData).get(0).getValue().getParty().partyId).isEqualTo(uuid);
+        assertThat(service.expandApplicantCollection(caseData, organisation)).hasSize(1);
+        assertThat(service.expandApplicantCollection(caseData, organisation)
```
swalker125: we could probably use ElementUtils.unwrapElements method here to make this a little simpler

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
```diff
@@ -130,4 +135,26 @@ void shouldNotAddNewPartyIdWhenApplicantsAlreadyHasPartyIdValue() {
...
+    void shouldReturnApplicantCollectionWithoutOrganisationDetailsWhenNoOrganisationExists() {
+        CaseData caseData = CaseData.builder().build();
+
+        List<Element<Applicant>> applicants = service.expandApplicantCollection(caseData,
+            Organisation.builder().build());
+        assertThat(applicants.get(0).getValue().getParty().getOrganisationName()).isNull();
+    }
+
+    private Organisation buildOrganisation() {
+        return Organisation.builder().name("Organisation").build();
```
swalker125: could we not set this as a constant and use everywhere. No difference between using this and using Organisation.builder().build()

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
```diff
@@ -130,4 +135,26 @@ void shouldNotAddNewPartyIdWhenApplicantsAlreadyHasPartyIdValue() {
 
         assertThat(applicant.getParty().partyId).isEqualTo(uuid);
     }
+
+    @Test
+    void shouldReturnApplicantCollectionWithOrganisationDetailsWhenOrganisationExists() {
+        CaseData caseData = CaseData.builder().build();
+
+        List<Element<Applicant>> applicants = service.expandApplicantCollection(caseData, buildOrganisation());
+        assertThat(applicants.get(0).getValue().getParty().getOrganisationName())
```
swalker125: unwrapElements here will make this a bit nicer

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
```diff
@@ -130,4 +135,26 @@ void shouldNotAddNewPartyIdWhenApplicantsAlreadyHasPartyIdValue() {
...
+            .isEqualTo(buildOrganisation().getName());
+    }
+
+    @Test
+    void shouldReturnApplicantCollectionWithoutOrganisationDetailsWhenNoOrganisationExists() {
+        CaseData caseData = CaseData.builder().build();
+
+        List<Element<Applicant>> applicants = service.expandApplicantCollection(caseData,
+            Organisation.builder().build());
+        assertThat(applicants.get(0).getValue().getParty().getOrganisationName()).isNull();
```
swalker125: unwrapElements here will make this a bit nicer

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: agreed, changed

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: why would this need to be static in a test method?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: Removed exception variable and changed the assertion to specific feign exception.. 

not sure on changing REQUEST name, it's like this in other test classes. would you suggest something like REFERENCE_DATA_REQUEST instead? http request is to them..

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: removed :)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: changed :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
toireasam: good idea, done !

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
toireasam: done :)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: This method isn't needed anymore so i've removed

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantController.java
toireasam: done

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantController.java
toireasam: i've done this now :) makes sense

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
toireasam: I've added the organisation always here, it isn't using optionals but have a look and see what you think. :) i've removed need for buildApplicantWithOrganisationDetails method

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
toireasam: do you mean like this at top of class

 private final Organisation ORGANISATION = Organisation.builder().name("Organisation").build();

instead of a method?

toireasam: Actually i've changed this buildOrganisation method to include the address now also so i don't think this comment is relevant anymore

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -1,21 +1,43 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
 
 @ActiveProfiles("integration-test")
 @WebMvcTest(ApplicantController.class)
 @OverrideAutoConfiguration(enabled = true)
 class ApplicantAboutToStartControllerTest extends AbstractControllerTest {
 
+    private String serviceAuthToken = "Bearer service token";
+    private final String userAuthToken = "Bearer token";
```
swalker125: this should be defined in AbstractControllerTest already so you can use that import. Maybe move serviceAuthToken to there as well

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +48,53 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+        given(authTokenGenerator.generate()).willReturn(serviceAuthToken);
+        given(organisationApi.findOrganisationById(userAuthToken, serviceAuthToken))
+            .willReturn(Organisation.builder().build());
+
         AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails);
 
         assertThat(callbackResponse.getData()).containsKey("applicants");
     }
+
+    @BeforeEach
```
swalker125: let's put the beforeEach above all the tests, then you can remove the duplicated mocks from the above test

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +48,53 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+    void shouldAddOrganisationDetailsToApplicantWhenOrganisationExists() {
+        CaseDetails caseDetails = CaseDetails.builder()
+            .data(Map.of("data", "some data"))
+            .build();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails);
+
+        CaseData data = mapper.convertValue(callbackResponse.getData(), CaseData.class);
+
+        String applicantOrganisationName = data.getAllApplicants().get(0).getValue().getParty().getOrganisationName();
```
swalker125: you can use ElementUtils.unwrapElements here to operate on List < Applicant >

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
```diff
@@ -26,8 +48,53 @@ void shouldPrepopulateApplicantDataWhenNoApplicantExists() {
...
+            .willReturn(Organisation.builder().build());
+
         AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails);
 
         assertThat(callbackResponse.getData()).containsKey("applicants");
     }
+
+    @BeforeEach
+    void setup() {
+        Organisation organisation = buildOrganisation();
```
swalker125: I think I commented before, you could set organisation globally and then cut down the amount of times you call buildOrganisation(), i.e String organisationName below, would just be organisation.getName()

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
```diff
@@ -3,37 +3,50 @@
...
-            return ImmutableList.of(Element.<Applicant>builder()
-                .value(Applicant.builder()
-                    .party(ApplicantParty.builder()
-                        // A value within applicant party needs to be set in order to expand UI view.
-                        .partyId(UUID.randomUUID().toString())
+            return ImmutableList.of(element(Applicant.builder()
+                        .party(ApplicantParty.builder()
+                            // A value within applicant party needs to be set in order to expand UI view.
+                            .partyId(randomUUID().toString())
+                            .organisationName(organisation.getName())
```
swalker125: I assume this will never be null?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
```diff
@@ -3,37 +3,50 @@
...
-                    .build())
-                .build());
-
+                    .build()));
         } else {
             return caseData.getApplicants();
         }
     }
 
+    private Address getOrganisationContactDetails(Organisation organisation) {
```
swalker125: I would call this getOrganisationAddress

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
```diff
@@ -3,37 +3,50 @@
...
+        ContactInformation contactInformation = ContactInformation.builder().build();
+
+        if (!isNull(organisation.getContactInformation())) {
```
swalker125: can use Objects.nonNull here I think

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
```diff
@@ -130,4 +137,45 @@ void shouldNotAddNewPartyIdWhenApplicantsAlreadyHasPartyIdValue() {
...
+        CaseData caseData = CaseData.builder().build();
+
+        List<Applicant> applicants = unwrapElements(service.expandApplicantCollection(
+            caseData, buildOrganisation()));
+
+        assertThat(applicants.get(0).getParty().getOrganisationName())
+            .isEqualTo(buildOrganisation().getName());
+        assertThat(applicants.get(0).getParty().getAddress()).isEqualTo(buildOrganisation()
+            .getContactInformation().get(0).toAddress());
+    }
```
swalker125: I would extract buildOrganisation() to a variable in this test rather than call it 3 times

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
```diff
@@ -130,4 +137,45 @@ void shouldNotAddNewPartyIdWhenApplicantsAlreadyHasPartyIdValue() {
...
+
+    @Test
+    void shouldReturnApplicantCollectionWithoutOrganisationDetailsWhenNoOrganisationExists() {
+        CaseData caseData = CaseData.builder().build();
+
+        List<Applicant> applicants = unwrapElements(service.expandApplicantCollection(
+            caseData, Organisation.builder().build()));
+        assertThat(applicants.get(0).getParty().getOrganisationName()).isNull();
```
swalker125: lets have a space inbetween applicants and the assertion

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrganisationServiceTest.java
```diff
@@ -138,4 +141,50 @@ void shouldRethrowExceptionOtherThanNotFound() {
...
+
+    private List<ContactInformation> buildOrganisationContactInformation() {
+        return List.of(ContactInformation.builder()
+            .addressLine1("Flat 12, Pinnacle Apartments")
+            .addressLine1("Saffron Central")
+            .county("London")
+            .country("United Kingdom")
+            .postCode("CR0 2GE")
+            .build());
+    }
```
swalker125: in this class I would do something like set Organisation.builder().build() and buildOrganisation() to global variables. EMPTY and POPULATED or something like that. What do you think?

e2e/fragments/addressPostcodeLookup.js
```diff
@@ -33,7 +33,10 @@ module.exports = {
   },
 
   enterAddressManually(address) {
-    I.click(this.cantEnterPostcodeLink);
+    if (this.cantEnterPostcodeLink.length > 0) {
```
swalker125: I can see some of the e2es are now failing around entering postcode, the only thing I can think would change is this. Maybe worth looking into? Easiest and quickest way to run e2e is to use an existing caseId in the correct state and hardcode it rather than creating a new case :)

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrganisationServiceTest.java
toireasam: happy with this! done

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
toireasam: cool :)

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
toireasam: makes sense, done :)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: done :)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: cool :)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: resolving as these are moved

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ApplicantServiceTest.java
toireasam: changed

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantAboutToStartControllerTest.java
toireasam: Resolving as don't think this is a big issue

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicantService.java
toireasam: doesn't throw null pointer, discussed

e2e/fragments/addressPostcodeLookup.js
toireasam: I spent some time trying to add a conditional into the applicant event page e2e yesterday, eventually discussed this with Adam and Tomeck and it's been decided that if reference data don't populate the address then we aren't going to populate it in the e2e. so resolving this. green build now.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrganisationServiceTest.java
```diff
@@ -48,6 +51,8 @@
     private static final String AUTH_TOKEN_ID = "Bearer authorisedBearer";
     private static final String SERVICE_AUTH_TOKEN_ID = "Bearer authorised service";
     private static final String USER_EMAIL = "test@test.com";
+    private static final Organisation EMPTY_ORGANISATION = Organisation.builder().build();
```
swalker125: there are other places you do this, I like this private static final declaration, the most. Let's implement it the same in other tests please :)

toireasam: done :D

PR #913, opened 2020-03-02
fix: don't block case submission on fee failure

lewisbirks:

mandatory work before payments goes live, related to https://tools.hmcts.net/jira/browse/FPLA-1425

### Change description ###

Quick fix to ensure that the case application process is not blocked by an issue with the fee register.
If there is an error it just won't display + populate the amount to pay field.

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerAboutToStartTest.java
```diff
@@ -92,6 +96,19 @@ void shouldNotAddAmountToPayFieldWhenFeatureToggleIsFalse() {
         assertThat(response.getData()).doesNotContainKey("amountToPay");
     }
 
+    @Test
+    void shouldNotDisplayAmountToPayFieldWhenErrorIsThrown() {
+        given(ldClient.boolVariation(eq("FNP"), any(), anyBoolean())).willReturn(true);
+        given(feeService.getFeeAmountForOrders(any())).willThrow(new FeeRegisterException(300, "duplicate", null));
```
akomisarek: Super piiiicky, I don't like status error 300. 
300 Multiple Choices
Indicates multiple options for the resource from which the client may choose (via agent-driven content negotiation). For example, this code could be used to present multiple video format options, to list files with different filename extensions, or to suggest word-sense disambiguation

I wouold suggest 409 it this instance as this would exactly mimic your "duplicate". 

Sorry but I am pedantic with the error codes!

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -150,7 +148,7 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmitEvent(
             .put("document_filename", document.originalDocumentName)
             .build());
 
-        data.remove("amountToPay");
+        removeTemporaryFields(caseDetails, "amountToPay", "displayAmountToPay");
```
akomisarek: Possibly could be extracted to the util class as it's noot specific to that controller.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerAboutToStartTest.java
lewisbirks: Chose 300 as that is the error I have most experience receiving from the fee-register but yeah can change to 409

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
lewisbirks: Was more thinking it could become part of some sort of base controller class. Although I don't think Lombok's constructor annotations allow for things to be passed to a super class 🤔, will have a look.

lewisbirks: Doesn't seem like it https://stackoverflow.com/a/29771875 (old answer but doesn't seem to have changed), would have been useful to have a base class which contained stuff like the object mapper ect, ah well, will move to a controller helper class

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerAboutToStartTest.java
akomisarek: Oh did they return this as 300?

lewisbirks: yes

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/ControllerHelper.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.hmcts.reform.fpl.utils;
+
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+public class ControllerHelper {
```
akomisarek: I am not sure if the ControllerHelper is good name I mean we process something on caseDetails, the reference to controller seems confusing.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerAboutToStartTest.java
```diff
@@ -91,6 +95,19 @@ void shouldNotAddAmountToPayFieldWhenFeatureToggleIsFalse() {
         assertThat(response.getData()).doesNotContainKey("amountToPay");
     }
 
+    @Test
+    void shouldNotDisplayAmountToPayFieldWhenErrorIsThrown() {
+        given(ldClient.boolVariation(eq("FNP"), any(), anyBoolean())).willReturn(true);
```
tomaszpowroznik: as repeated couple times, can be extracted to method
givenPaymentEnabled(boolean enabled){
given(ldClient.boolVariation(eq("FNP"), any(), anyBoolean())).willReturn(enabled);
}

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -75,18 +76,16 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
...
                     FeesData feesData = feeService.getFeesDataForOrders(caseData.getOrders());
                     data.put("amountToPay", BigDecimalHelper.toCCDMoneyGBP(feesData.getTotalAmount()));
+                    data.put("displayAmountToPay", YES.getValue());
                 }
-
             } catch (FeeRegisterException ignore) {
-                // TODO: 21/02/2020 Replace me in FPLA-1353
-                //  this is an error message for when the Fee Register is unavailable
-                errors.add("XXX");
+                data.put("displayAmountToPay", NO.getValue());
```
tomaszpowroznik: I would log error here as well

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsHelperTest.java
```diff
@@ -41,4 +49,36 @@ void shouldThrowAnExceptionIfCaseNumberIsNull() {
...
+        private CaseDetails caseDetails = CaseDetails.builder().data(data).build();
+
+        @BeforeEach
+        void populateMap() {
+            data.put("key1", "some value 1");
+            data.put("key2", "some value 2");
+            data.put("key3", 3);
+        }
+
+        @AfterEach
```
tomaszpowroznik: should not be needed, as each test is executed in new instance of test class

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
lewisbirks: Logged elsewhere so it isn't needed twice

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsHelperTest.java
lewisbirks: Wasn't aware of that, removed.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
tomaszpowroznik: not sure if we have to remove amoutToPay, it can be useful audit information, but anyway sits in payment service

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -75,18 +76,16 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(
...
+            } finally {
+                String label = String.format(CONSENT_TEMPLATE, userDetailsService.getUserName(authorization));
```
tomaszpowroznik: personally i would move this before try-catch instead of using finally, but it is ok

PR #916, opened 2020-03-04
feat: payment history tab (FPLA-1298)

lewisbirks:

[FPLA 1298](https://tools.hmcts.net/jira/browse/FPLA-1298)

### Change description ###

Added access to the Payment History tab and case field to HMCTS staff.
Mocked the calls that is made to the payments to populate the tab.

Latest version of ccd-docker is also required for the local mocking and the latest version of case-web-management is required for the UI.

```
[ ] Yes
[x] No
```


e2e/tests/localAuthoritySubmitsApplication_test.js
```diff
@@ -482,3 +482,9 @@ Scenario('local authority submits after giving consent @create-case-with-mandato
   caseViewPage.selectTab(caseViewPage.tabs.documents);
   I.see('council_v_Smith.pdf');
 });
+
+Scenario('local authority confirms payment after submission', async  (I, caseViewPage) => {
+  caseViewPage.selectTab(caseViewPage.tabs.paymentHistory);
+  I.refreshPage(); // Required for some versions of the local build
```
pd-dudkiewicz: Are we leaving it like this?

lewisbirks: Not sure, doesn't work for me without it and when people update their images surely they will run into the same issue. Strange thing was though I wasn't getting it on preview 🤔

PR #922, opened 2020-03-09
fix:update ccd web url (changed after ccd web migration to aks in aat)

tomaszpowroznik:
Update ccd web url

```
[ ] Yes
[X] No
```


charts/fpl-case-service/values.aat.template.yaml
```diff
@@ -7,7 +7,7 @@ java:
     IDAM_S2S_AUTH_URL: http://rpe-service-auth-provider-aat.service.core-compute-aat.internal
     CORE_CASE_DATA_API_URL: http://ccd-data-store-api-aat.service.core-compute-aat.internal
     DOCUMENT_MANAGEMENT_URL: http://dm-store-aat.service.core-compute-aat.internal
-    CCD_UI_BASE_URL: https://ccd-case-management-web-aat.service.core-compute-aat.internal
+    CCD_UI_BASE_URL: https://www-ccd.aat.platform.hmcts.net
```
lewisbirks: Do we need to set this here? Is it not just overriding the value from `values.yaml` with the same value?

tomaszpowroznik: you are right, i would like to do chart files cleanup in separate commit

PR #929, opened 2020-03-10
refactor: charts cleanup

pd-dudkiewicz:
### Change description ###

Removed redundant env variables and secrets definitions from charts.

```
[ ] Yes
[x] No
```


charts/fpl-case-service/values.yaml
```diff
@@ -5,20 +5,49 @@ tags:
 java:
   image: hmctspublic.azurecr.io/fpl/case-service:latest
   environment:
+    AUTH_IDAM_CLIENT_BASEURL: https://idam-api.aat.platform.hmcts.net
```
Gogetter: we can also use `{{ .Values.global.environment }}` instead of hardcoding the environment name

charts/fpl-case-service/values.aat.template.yaml
```diff
@@ -3,39 +3,4 @@ java:
   ingressHost: ${SERVICE_FQDN}
   environment:
```
swalker125: I guess whole environment block can be removed

PR #932, opened 2020-03-12
impr: EPO child description not mandatory after selecting 'yes' on description needed

joez-f:
### Change description ###

EPO child description made mandatory after selecting 'Yes' on description needed. Currently users can select 'Yes' and then not enter anything in description (this results in failed about-to-submit callback)

```
[ ] Yes
[X] No
```


joez-f:
https://tools.hmcts.net/jira/browse/FPLA-830 original ticket does not state anything about mandatory/optional (but logically it should be mandatory otherwise there is a callback error).

PR #936, opened 2020-03-13
Make chart env variables dynamic

mounikahmcts:
Made chart env variable dynamic by setting environment name from {{ .Values.global.environment }}

### DIFF
--- a/charts/fpl-case-service/Chart.yaml
+++ b/charts/fpl-case-service/Chart.yaml
@@ -1,7 +1,7 @@
 name: fpl-case-service
 apiVersion: v1
 home: https://github.com/hmcts/fpl-ccd-configuration
-version: 1.6.0
+version: 1.7.0
 description: FPL Case Service
 maintainers:
   - name: HMCTS Family Public Law team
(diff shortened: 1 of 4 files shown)


charts/fpl-case-service/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: fpl-case-service
 apiVersion: v1
 home: https://github.com/hmcts/fpl-ccd-configuration
-version: 1.6.0
+version: 1.7.0
```
pd-dudkiewicz: Remember to create PR for https://github.com/hmcts/cnp-flux-config repo as well

charts/fpl-case-service/values.yaml
```diff
@@ -5,20 +5,22 @@ tags:
...
-    IDAM_S2S_AUTH_URL: http://rpe-service-auth-provider-aat.service.core-compute-aat.internal
-    RD_PROFESSIONAL_API_URL: http://rd-professional-api-aat.service.core-compute-aat.internal
-    CORE_CASE_DATA_API_URL: http://ccd-data-store-api-aat.service.core-compute-aat.internal
-    SEND_LETTER_URL: http://rpe-send-letter-service-aat.service.core-compute-aat.internal
-    PAYMENT_API_URL: http://payment-api-aat.service.core-compute-aat.internal
-    FEES_REGISTER_API_URL: http://fees-register-api-aat.service.core-compute-aat.internal
-    DOCUMENT_MANAGEMENT_URL: http://dm-store-aat.service.core-compute-aat.internal
-    CCD_UI_BASE_URL: http://www-ccd.aat.platform.hmcts.net
+    IDAM_CLIENT_ID: fpl_case_service
+    IDAM_CLIENT_REDIRECT_URI: https://fpl-case-service-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/oauth2/callback
```
pd-dudkiewicz: nice :+1: can we also put keyVaults section here, and remove it from preview and aat files?

charts/fpl-case-service/Chart.yaml
mounikahmcts: sure will do it once it is merged to master

charts/fpl-case-service/values.yaml
mounikahmcts: @pd-dudkiewicz  Done. Moved key vaults as well

PR #941, opened 2020-03-17
fix: flaky e2e test fix

pd-dudkiewicz:

```
[ ] Yes
[x] No
```


e2e/steps_file.js
```diff
@@ -196,6 +197,7 @@ module.exports = function () {
 
     async addAnotherElementToCollection(collectionName) {
       const numberOfElements = await this.grabNumberOfVisibleElements('.collection-title');
+      this.waitForElement('.form-group .button');
```
tomaszpowroznik: maybe this one should be line above?

pd-dudkiewicz: removed it, turned out that `addAnotherElementToCollection` was not working for scanned documents

e2e/pages/events/attachScannedDocsEvent.page.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
   },
 
   enterScannedDocument(scannedDocument, file) {
+    I.waitForElement('.form-group .button');
```
tomaszpowroznik: i think this one is not really needed

PR #947, opened 2020-03-18
feat: (FPLA-1466) allow admin to add case number in all states from submitted onw…

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-1466

```
[ ] Yes
[X] No
```


tomaszpowroznik:
> we can open up the add familymanid event to all states and cut down a lot of this code. Admin cannot see open state so this will not have any impact

The event in submitted and post submitted are a bit different in a way that they dont have CallBackURLSubmittedEvent callback. Of course adding this callback to them wont change a lot, it was done this way just to save this one additional call which now is not needed but most likely will be needed in future, so happy to change

swalker125:
> > we can open up the add familymanid event to all states and cut down a lot of this code. Admin cannot see open state so this will not have any impact
> 
> The event in submitted and post submitted are a bit different in a way that they dont have CallBackURLSubmittedEvent callback (send to robotics). Of course adding this callback to them wont change a lot as we check anyway if case number is added or updated, it was done this way just to save this one additional call which now is not needed but most likely will be needed in future, so happy to change. In general relying on access control in such way is a bit slippery as changes to access control breaks such functionality. I think that it can be a bit confusing as well to someone who is not familiar with made assumption. Additionally it is not resilient, if new state will be introduced at some point and admin will have access to it this event will be visible there even if not intended.
> If user has both admin and LA (or any other role with access to open state) role (probably not valid from business pov but possible) then event will be available from open state
> Another thing is that tools we have like project-event-permissions.sh produces wrong output in such configuration

hmm so I would argue this IS how CCD should be used in this case. If you look at it from the point of view of a hmcts admin has acess to x states, and this event is available in those states... it makes sense. The duplication we cut down as well is a huge win imo. But I understand you point around future states. I think we can guarantee an lasolicitor will never ever be a hmcts admin, but again, I guess it _could_ be possible... I guess I will leave this with you what you want to do, the CCD config generator hides this duplication a little bit, essentially saying I want this event -> in x, y, z state. So the swap to that may be the solution we need

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -245,3 +245,10 @@ Scenario('HMCTS admin sends email to gatekeeper with a link to the case', async
   await I.completeEvent('Save and continue');
   I.seeEventSubmissionConfirmation(config.administrationActions.sendToGatekeeper);
 });
+
+Scenario('HMCTS admin update FamilyMan reference number after sending case to gatekeeper', async (I, caseViewPage, loginPage, enterFamilyManCaseNumberEventPage) => {
```
swalker125: I don't this test is necessary any longer

PR #975, opened 2020-03-23
hack: disable mutation build (FPLA-373)

akomisarek:
### Change description ###

Disable mutation tests as they are failing anyway. 

```
[ ] Yes
[X] No
```


Jenkinsfile_CNP
```diff
@@ -69,6 +69,7 @@ withPipeline(type, product, component) {
     loadVaultSecrets(serviceSecrets + otherSecrets)
   }
   onPR {
+    enableMutationTest()
```
lewisbirks: This seems to be doing the opposite of what the PR says it should be doing. Surely all you need to do is disable it in the nightly file?

akomisarek: Yes, I am trying to see if this is going to enable on PR and then I will remove from nightly pipeline :D, sorry it's sort of in the draft mode.

PR #985, opened 2020-03-25
fix: added aadIdentityName to base chart

swalker125:
 
N/A

### Change description ###

add aad identity name to base chart

```
[ ] Yes
[x] No
```


charts/fpl-case-service/values.yaml
```diff
@@ -3,6 +3,7 @@ tags:
   ccd-idam-pr: false
 
 java:
+  aadIdentityName: family-public-law
```
akomisarek: Do you need to bump versions now?

PR #990, opened 2020-03-26
fix: Pick AppInsightInstrumentation dynamically

mounikahmcts:
Observed that perfTest AppInsights is not working because of incorrect app insight key in the flux config file. So, modified this to pick it dynamically from the azure app insights.


charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -16,6 +16,27 @@ java:
...
+        - local-authority-code-to-hmcts-court-mapping
+        - local-authority-code-to-cafcass-mapping
+        - local-authority-code-to-shared-inbox-mapping
+        - local-authority-fallback-inbox
+        - system-update-user-username
+        - system-update-user-password
+        - robotics-notification-sender
+        - robotics-notification-recipient
+        - ld-sdk-key
+        - ctsc-inbox
```
mounikahmcts: AppInsights are not required for preview.  So, overriding key vaults here.

infrastructure/main.tf
```diff
@@ -40,6 +40,13 @@ resource "azurerm_application_insights" "appinsights" {
   tags = "${var.common_tags}"
 }
 
+#Copying appinsights key to the valut
+resource "azurerm_key_vault_secret" "AZURE_APPINSGHTS_KEY" {
+  name         = "AppInsightsInstrumentationKey"
+  value        = "${azurerm_application_insights.appinsights.instrumentation_key}"
+  key_vault_id = "${module.key-vault.key_vault_id}"
+}
```
tomaszpowroznik: nice

service/src/main/resources/bootstrap.yaml
```diff
@@ -21,3 +21,4 @@ spring:
         robotics-notification-recipient: robotics.notification.recipient
         ld-sdk-key: ld.sdk_key
         ctsc-inbox: fpl.ctsc_inbox
+        AppInsightsInstrumentationKey: azure.application-insights.instrumentation-key
```
tomaszpowroznik: needs to be removed from flux, as it will be overriden otherwise

PR #996, opened 2020-03-27
 fix: Clean up tf files as we completely migrated to aks

mounikahmcts:
https://tools.hmcts.net/jira/browse/FPLA-597

Clean up tf files as we completely migrated to aks


infrastructure/state.tf
```diff
@@ -1,25 +1,3 @@
...
-
-data "terraform_remote_state" "core_apps_compute" {
-  backend = "azurerm"
-
-  config {
-    resource_group_name  = "mgmt-state-store-${var.subscription}"
-    storage_account_name = "mgmtstatestore${var.subscription}"
-    container_name       = "mgmtstatestorecontainer${var.env}"
-    key                  = "core-compute/${var.env}/terraform.tfstate"
-  }
```
tomaszpowroznik: not sure of purpose of these, but lets try without it

PR #1003, opened 2020-03-30
fix: update customerReference to have a default value of 'Not provided' when no value entered [FPLA-1524]

Gogetter:
### Change description ###

1. PBA for C110a

    - If no value is entered for the Customer reference field by the user then send the value "Not provided" so that the payment can be processed

2. PBA for C2
    - Change the field label from File reference to Customer reference (for consistency)
    - If no value is entered for the Customer reference field by the user then send the value "Not provided" so that the payment can be processed

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/payment/PaymentService.java
```diff
@@ -79,13 +80,17 @@ public void makePaymentForC2(Long caseId, CaseData caseData) {
         CreditAccountPaymentRequest paymentRequest = getCreditAccountPaymentRequest(caseId,
             c2DocumentBundle.getPbaNumber(),
             c2DocumentBundle.getClientCode(),
-            c2DocumentBundle.getFileReference(),
+            defaultCustomerReference(c2DocumentBundle.getCustomerReference()),
```
akomisarek: Could we specify that's only default when blank? It suggest we always use default?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/payment/PaymentService.java
```diff
@@ -79,13 +80,17 @@ public void makePaymentForC2(Long caseId, CaseData caseData) {
...
             localAuthorityName,
             feesData);
 
         callPaymentsApi(paymentRequest);
     }
 
+    public String defaultCustomerReference(final String currentValue) {
+        return isNotBlank(currentValue) ? currentValue : "Not provided";
```
akomisarek: Maybe we could extract as const to top level of the file?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/PaymentServiceTest.java
```diff
@@ -93,18 +96,21 @@ void setup() {
...
                     .type(WITH_NOTICE)
                     .pbaNumber("PBA123")
                     .clientCode("clientCode")
-                    .fileReference("fileReference")
+                    .customerReference(customerReference)
                     .build())))
                 .build();
-            CreditAccountPaymentRequest expectedPaymentRequest = testCreditAccountPaymentRequestBuilder()
+            CreditAccountPaymentRequest expectedPaymentRequest = testCreditAccountPaymentRequestBuilder(
+                customerReference)
```
akomisarek: I don't see how this test works, if we pass nullandemptySource why we are not explicitly checking against "Not provided" is different label.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/PaymentServiceTest.java
```diff
@@ -229,13 +239,14 @@ void resetInvocations() {
...
-    private CreditAccountPaymentRequest.CreditAccountPaymentRequestBuilder testCreditAccountPaymentRequestBuilder() {
+    private CreditAccountPaymentRequest.CreditAccountPaymentRequestBuilder testCreditAccountPaymentRequestBuilder(
+        final String customerReference) {
         return CreditAccountPaymentRequest.builder()
             .accountNumber("PBA123")
             .caseReference("clientCode")
             .ccdCaseNumber(String.valueOf(CASE_ID))
             .currency(GBP)
-            .customerReference("fileReference")
+            .customerReference(paymentService.defaultCustomerReference(customerReference))
```
akomisarek: Ah I see now, ok I don't like this approach, it causes thee linee too be ultimately not tested - we expect exactly what's returned by the service, we need to test it and provide the value manually.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/C2DocumentBundle.java
```diff
@@ -11,7 +12,10 @@
     private final String nameOfRepresentative;
     private final String pbaNumber;
     private final String clientCode;
-    private final String fileReference;
+
+    @JsonProperty(value = "fileReference")
```
akomisarek: I guess we could check with devops if we have any entries like that (probably not, and then don't have JsonProprety), but I am noot massively concerned.

Gogetter:
Welcome onboard @nkoranne. Yes, that is a good suggestion for the initial implementation of a feature. In general we try (our best 😃), to add the jira number to the title of the pr.

akomisarek:
@nkoranne hi Nikhil, welcome on board, I am happy to have conversation about that, we try to include the ticket number in the tickets and ocmmits so you can navigate easily, but we haven't considered adding it to the test, maybe it something we could start doing!

nkoranne:
Thanks Adam and Etim.
Yeah sure, we can discuss it during the retro.

All other code changes looks fine to me

PR #1004, opened 2020-03-30
impr: use java models for cmo generation (FPLA-1488)

swalker125:
### Change description ###

```
[ ] Yes
[ ] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftCMOController.java
```diff
@@ -77,7 +77,8 @@ public AboutToStartOrSubmitCallbackResponse handleMidEvent(@RequestHeader("autho
         final Map<String, Object> data = caseDetails.getData();
         final CaseData caseData = mapper.convertValue(data, CaseData.class);
 
-        Map<String, Object> cmoTemplateData = docmosisTemplateDataGenerationService.getTemplateData(caseData, true);
+        Map<String, Object> cmoTemplateData = docmosisTemplateDataGenerationService.getCaseManagementOrderData(caseData)
+            .toMap(mapper);
 
         Document document = getDocument(authorization, userId, cmoTemplateData);
```
lewisbirks: Not really sure where the best place to put this comment is but thoughts on having `AbstractDocmosisOrder` extend a base class called something like `AbstractDocmosisData` or something like that which has the default implementation of the toMap method
```java
public abstract class AbstractDocmosisData {
    public Map<String, Object> toMap(ObjectMapper mapper) {
        return mapper.convertValue(this, new TypeReference<>() {});
    }
}
```
and the the subsequent classes can call this method and perform additional logic
```java
public class AbstractDocmosisOrder extends AbstractDocmosisData {
    // fields...

    public Map<String, Object> toMap(ObjectMapper mapper) {
        Map<String, Object> map = super.toMap(mapper);
        
       // additional logic...

        return map;
    }
}
```
With this we can then have an overloaded the `DocmosisDocumentGeneratorService.generateDocmosisDocument`
to have a method that looks like
``` java
public class DocmosisDocumentGeneratorService {
    // fields and other methods...

    public DocmosisDocument generateDocmosisDocument(AbstractDocmosisData templateData, DocmosisTemplates template) {
        return generateDocmosisDocument(templateData.toMap(mapper), template);
    }
}
```
Whether we would also pass the mapper as a parameter in the method or autowire it into the class is another question as well.

I feel like this would at least from a readability point of view be an improvement as we would no longer be passing around a map that doesn't have that much meaning but instead an object which is more clear as to what it will be doing, it's use, and it's contents.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocmosisDocumentGeneratorService.java
```diff
@@ -15,48 +15,46 @@
...
-
-    @Autowired
-    DocmosisDocumentGeneratorService(RestTemplate restTemplate,
-                                     DocmosisConfiguration docmosisDocumentGenerationConfiguration) {
-        this.restTemplate = restTemplate;
-        this.docmosisDocumentGenerationConfiguration = docmosisDocumentGenerationConfiguration;
+    private final DocmosisConfiguration configuration;
+    private final ObjectMapper mapper;
+
+    public DocmosisDocument generatedDocmosisDocument(AbstractDocmosisData templateData, DocmosisTemplates template) {
```
lewisbirks, line 30: generate or generated? I think that generate makes more sense as that is what is actually happening.

swalker125: definitely generate, typo by me

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisData.java
```diff
@@ -5,8 +5,10 @@
 
 import java.util.Map;
 
-public abstract class AbstractDocmosisData {
+public class AbstractDocmosisData {
+
     public Map<String, Object> toMap(ObjectMapper mapper) {
-        return mapper.convertValue(this, new TypeReference<>() {});
+        return mapper.convertValue(this, new TypeReference<>() {
+        });
```
lewisbirks, line 12: Seems like unnecessary change

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseManagementOrder.java
```diff
@@ -33,7 +34,7 @@
 
     @JsonIgnore
     public boolean isDraft() {
-        return action == null || !action.getType().equals(SEND_TO_ALL_PARTIES);
+        return ofNullable(action).map(x -> !SEND_TO_ALL_PARTIES.equals(action.getType())).orElse(true);
```
lewisbirks, line 37: Not sure then conversion to the optional improves anything although I would agree that the flipping of the equals is better

lewisbirks: If you want to be insistent on the optional then it would be better to use x in the map function rather than action

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerMidEventTest.java
```diff
@@ -55,24 +56,24 @@
...
+            .willReturn(docmosisDocument);
 
-    @Test
-    void midEventShouldGenerateDraftStandardDirectionDocument() {
         given(uploadDocumentService.uploadPDF(PDF, DRAFT_ORDER_FILE_NAME))
             .willReturn(document);
+    }
 
+    @Test
+    void midEventShouldGenerateDraftStandardDirectionDocument() {
```
lewisbirks, line 76: Don't think that the test needs to start with `midEventShould`, just `should` should be enough

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisData.java
swalker125: ahh damn, reformat code does this for me by default, shall revert. Good spot

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseManagementOrder.java
swalker125: fine to revert and then swap the equals around

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerMidEventTest.java
swalker125: I totally agree, however this was done in a refactoring of draft orders controller and not in this PR. I am not going address within this.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisCaseManagementOrder.java
```diff
@@ -0,0 +1,52 @@
...
+    private final Schedule schedule;
+    private final boolean scheduleProvided;
+
+    private final List<DocmosisRecital> recitals;
+
+    //not needed can manage in doc
+    private final boolean recitalsProvided;
+
+    //not needed can manage in doc
+    private final int numberOfChildren;
```
lewisbirks, line 21: is this going to be done in this PR or in a future date?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/NoticeOfProceedingsServiceTest.java
```diff
@@ -61,8 +61,7 @@
     @Autowired
     private HearingVenueLookUpService hearingVenueLookUpService;
 
-    private CommonCaseDataExtractionService commonCaseDataExtractionService = new CommonCaseDataExtractionService(
-        hearingVenueLookUpService);
+    private CommonCaseDataExtractionService commonCaseDataExtractionService = new CommonCaseDataExtractionService();
```
lewisbirks, line 64: Could just be autowired in

lewisbirks: Although I guess just creating a new one doesn't actually hurt and requires less from spring so may be slightly better in performance

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerAboutToSubmitTest.java
```diff
@@ -68,6 +72,7 @@
     private static final UUID ID = randomUUID();
 
     private final DocumentReference cmoDocument = buildFromDocument(document());
+    private CaseDetails populatedCaseDetails;
```
toireasam, line 75: this could be private final CaseDetails populatedCaseDetails  = populatedCaseDetails();

and removed from setup

swalker125: I update populatedCaseDetails in each test, so needs to be reset in the beforeEach

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisCaseManagementOrder.java
swalker125: doing now!

service/src/test/java/uk/gov/hmcts/reform/fpl/service/NoticeOfProceedingsServiceTest.java
swalker125: will leave as is

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerAboutToSubmitTest.java
swalker125: I update populatedCaseDetails directly so I need to reset before each test in the before each

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisOrder.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import static java.util.stream.Collectors.groupingBy;
+import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
+
+@Getter
+@AllArgsConstructor
+public class AbstractDocmosisOrder extends AbstractDocmosisData {
```
akomisarek, line 15: Are we sure all of those are true for any order? I.e. we don't have complianceDeadline for most of the order do we? It seems only for SDO / CMO, might be confusing if we wanted to reuse this class for other orders. Would you consider renaming?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisOrder.java
```diff
@@ -0,0 +1,41 @@
...
+    private final DocmosisJudgeAndLegalAdvisor judgeAndLegalAdvisor;
+    private final String courtName;
+    private final String familyManCaseNumber;
+    private final String dateOfIssue;
+    private final String complianceDeadline;
+    private final List<DocmosisRespondent> respondents;
+    private final List<DocmosisChild> children;
```
akomisarek, line 22: Probably not ideal with duplicating the Docmosis from the package name, but I can imagine it makes development easier due to naming clash, so it's fine.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisOrder.java
```diff
@@ -0,0 +1,41 @@
...
+    private final boolean respondentsProvided;
+    private final String applicantName;
+    private final DocmosisHearingBooking hearingBooking;
+    private final List<DocmosisDirection> directions;
```
akomisarek, line 26: Again not every order has direction, this is abstract, but not really modelling the reality :(

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisStandardDirectionOrder.java
```diff
@@ -1,42 +1,28 @@
...
-
-        if (isNotEmpty(this.directions)) {
-            map.putAll(this.directions.stream().collect(groupingBy(direction -> direction.assignee.getValue())));
-        }
-
-        map.remove("directions");
-
-        return map;
+@Getter
+public class DocmosisStandardDirectionOrder extends AbstractDocmosisOrder {
```
akomisarek, line 9: Maybe the order doesn't need to be abstract then? This class doesn't provide much value.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
```diff
@@ -1,7 +1,6 @@
 package uk.gov.hmcts.reform.fpl.utils;
```
akomisarek, line 1: Shouldn't this leave with tests?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderGenerationServiceTest.java
```diff
@@ -0,0 +1,287 @@
+package uk.gov.hmcts.reform.fpl.service;
```
akomisarek, line 1: Let's ask Raghu/Dough to review it :D

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisOrder.java
swalker125: No these are not all common for orders, I've already had this chat with Lewis and I going to keep working on as I work my way through other orders. This Order object will probably turn into something a little more specific to group is as CaseManagementOrders maybe? And AbstractDocmosisOrder will keep whatever is common.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisOrder.java
swalker125: Yes, so I don't really want Docmosis in the name at all, I just think it's confusing to be using two different Child types in the same lines of code. Any other ideas for this? Maybe being the same name is fine?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisOrder.java
swalker125: related to top comment

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
lewisbirks: I think so but I think it also used within both integration and unit tests so one breaks if it is placed in the package of the other

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisStandardDirectionOrder.java
swalker125: again, I believe related to my top comment.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
swalker125: YES. I despise this class. However, it is used in integration AND unit tests so isn't as simple as just moving. I am not addressing within this PR

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderGenerationServiceTest.java
swalker125: yep, I am going to move it to QA for a sanity check

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/AbstractDocmosisOrder.java
akomisarek: It's fine, but Java will force you to use fully typed name for one of them so you will have:
uk.gov.hmcts.reform.fpl.model.docmosis.Child which is a bit annoying as well. So you can import only one with clashing name and need to fully reference the other. 

I personally prefer fully referencing the less used one as it is anyway only one or two lines where the mapping happens, so maybe something you could consider.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataGeneratorHelper.java
akomisarek: Leave means live! :D but yes happy to not address it here.

PR #1022, opened 2020-04-02
feat: CMO review notifications for representatives (FPLA-926)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-926
GOV Notify template: https://www.notifications.service.gov.uk/services/12f756df-f01d-4a32-a405-e1ea8a494fbb/templates/7a513cfd-cf3c-4778-aca0-4bbd1a37863a

### Change description ###

Send notifications to representatives when a draft CMO needs reviewing by parties

```
[ ] Yes
[X] No
```


service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -607,6 +609,36 @@ void shouldNotifyCtscAdminOfCMOReadyForJudgeReviewWhenCtscIsEnabled() throws Exc
...
+                EMAIL))
+                .willReturn(getExpectedEmailRepresentativesForAddingPartiesToCase());
+
+            notificationHandler.sendEmailForCaseManagementOrderReadyForPartyReview(
+                new CaseManagementOrderReadyForPartyReviewEvent(callbackRequest, requestData, documentContents));
+            
+            verify(notificationService).sendEmail(
+                eq(CMO_READY_FOR_PARTY_REVIEW_NOTIFICATION_TEMPLATE),
+                eq("fred@flinstone.com"),
+                dataCaptor.capture(),
```
joez-f, line 632: These parameters are tested in controller test - is there any benefit to testing them again here? Or is using this OK? (verified that the email has been sent to right address + with right template)

joez-f: Cool, that makes sense - I will add tests in for these parameters

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseManagementOrderProgressionControllerTest.java
```diff
@@ -56,6 +75,7 @@
     private static final String FAMILY_MAN_CASE_NUMBER = "SACCCCCCCC5676576567";
     private static final String HMCTS_ADMIN_INBOX = "admin@family-court.com";
     private static final String CTSC_ADMIN_INBOX = "FamilyPublicLaw+ctsc@gmail.com";
+    private static final byte[] PDF = {1, 2, 3, 4, 5};
```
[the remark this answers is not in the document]
joez-f: For now I think let's keep it here but yes maybe can make a small tech debt to clean up some controller tests and move more fields to AbstractControllerTest - should make things more generic (use same email addresses/caseid etc for everything, would make writing tests easier)

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationHandlerTest.java
```diff
@@ -607,6 +609,36 @@ void shouldNotifyCtscAdminOfCMOReadyForJudgeReviewWhenCtscIsEnabled() throws Exc
                 "12345");
         }
 
+        @Test
+        void shouldNotifyRepresentativesOfCMOReadyForPartyReview() throws Exception {
```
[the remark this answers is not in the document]
joez-f: I think the class has a lot of Exceptions that are never thrown - I'll change this one, will ask @Gogetter to update them in his notification refactor

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderProgressionService.java
```diff
@@ -56,6 +58,7 @@ private void progressDraftCaseManagementOrder(CaseDetails caseDetails, CaseManag
                 break;
             case PARTIES_REVIEW:
                 caseDetails.getData().put(CASE_MANAGEMENT_ORDER_SHARED.getKey(), order.getOrderDoc());
+                publishReadyForPartyReviewEvent(caseDetails, requestData);
```
pd-dudkiewicz, line 61: Might be worth thinking about verifying `applicationEventPublisher` calls in unit tests. Happy to leave is as it is though, because I can see that we're not doing it for any route currently. But it feels like this class is doing too much to test it properly and some parts could potentially be extracted.

joez-f: Good to think about, we can add it to tech improvements perhaps (looks like we could verify applicationEventPublisher in several tests where we currently don't)

PR #1055, opened 2020-04-07
impr: update to 'create an order' notification to include LA and Digital Reps [FPLA-1519]

Gogetter:
[FPLA-1519](https://tools.hmcts.net/jira/browse/FPLA-1519)

### Change description ###
When an email is triggered by an order being created via "create an order" event, send the email to Local Authorities AND any digitall_service representatives of parties on the case.

### QA Note ###
Template mentioned in jira is currently not updated **as this is currently used in production**. New temp notification template is [here](https://www.notifications.service.gov.uk/services/12f756df-f01d-4a32-a405-e1ea8a494fbb/templates/a0f911cf-4010-4022-bf34-a183cd711b8e).

**Once all test is confirmed, I can update needed template.**

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/utils/NotifyAttachedDocumentLinkHelper.java
```diff
@@ -0,0 +1,30 @@
...
+import uk.gov.service.notify.NotificationClientException;
+
+import java.util.Optional;
+
+import static org.apache.commons.lang3.ArrayUtils.isNotEmpty;
+import static org.apache.commons.lang3.exception.ExceptionUtils.getStackTrace;
+import static uk.gov.service.notify.NotificationClient.prepareUpload;
+
+@Slf4j
+public class NotifyAttachedDocumentLinkHelper {
```
joez-f, line 14: if we're adding new helper can we also make use of it in other places where we attach a document? OrderIssuedEmailContentProvider + CaseManagementOrderEmailContentProvider can use this helper

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/OrderIssuedEmailContentProvider.java
```diff
@@ -84,6 +85,17 @@ public OrderIssuedEmailContentProvider(@Value("${ccd.ui.base.url}") String uiBas
             .build();
     }
 
+    public Map<String, Object> buildOrderNotificationParameters(final CaseDetails caseDetails,
```
joez-f, line 88: maybe this could be named `buildNotificationParametersWithCaseUrlAndAttachedDocument`

and the method above `buildNotificationParametersWithAttachedDocument`
thoughts? might be a bit clearer

Gogetter: we can leave as-is, as this can be reused generically. Also, renaming the method method above changes its existing function and leads to more changes thereby bloating the pr.

PR #1062, opened 2020-04-08
increase default pupeteer timeout (30 sec)

tomaszpowroznik:
current timeout of 10 sec is not enough, especially when log out/log in, and cause plenty of retries which do not help when application is slow


tomaszpowroznik:
> I thought that the default was 1s https://codecept.io/helpers/Puppeteer/#configuration

https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagewaitforselectorselector-options

PR #1072, opened 2020-04-14
impr:Use openID idam endpoints to get user details (FPLA-1637)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-1637


nkoranne:
looks good

PR #1073, opened 2020-04-14
fix: added updated order to caseData for next hearing date in cmo (FPLA-1624)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1624

### Change description ###

A different object was being updated with new values, but then not being sent to the template service for document generation. This change adds a toBuilder method on case data and adds the updated order object. 

This is not the cleanest fix, I would like to revist the cmo controllers in a wider refactoring. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderController.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderController.java
@@ -143,7 +143,7 @@ public class ActionCaseManagementOrderController {
             order = caseManagementOrderService.addNextHearingToCMO(caseData.getNextHearingDateList(), order);
         }
 
-        Document document = getDocument(caseData, order.isDraft());
+        Document document = getDocument(caseData.toBuilder().caseManagementOrder(order).build(), order.isDraft());
 
         order = caseManagementOrderService.addDocument(order, document);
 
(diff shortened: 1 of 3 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderController.java
```diff
@@ -143,7 +143,7 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(
             order = caseManagementOrderService.addNextHearingToCMO(caseData.getNextHearingDateList(), order);
         }
 
-        Document document = getDocument(caseData, order.isDraft());
+        Document document = getDocument(caseData.toBuilder().caseManagementOrder(order).build(), order.isDraft());
```
Gogetter, line 146: could we pass `caseData.toBuilder().caseManagementOrder(order).build()` into a different variable?

akomisarek, line 146: No tests related to this change? Seems fishy. 🤔

swalker125: no need to, it's not used again. I think we can definitely clean up how this works in a separate tech debt. I already have one open to remove all this logic from the controller

Gogetter: Cool. Looked at it and there was a lot happening. But I agree it can be cleaned up in a TD.

PR #1076, opened 2020-04-15
feat: Add option "not required" for SWET (FPLA-1184)

mounikahmcts:

[FPLA-1184](https://tools.hmcts.net/jira/browse/FPLA-1184)

### Change description ###
Add option "not required" for SWET

Screenshot 2020-04-15 at 09 04 18


e2e/fragments/caseDocuments.js
```diff
@@ -25,6 +25,30 @@ module.exports = {
     };
   },
 
+  uploadDocumentsWhenSWETNotRequired() {
```
tomaszpowroznik, line 28: do we need to upload all documents again? It will just slow test, can we just update SWET document?

tomaszpowroznik:
> what about AC no 2:
> If user has selected that any other document is "In SWET" it should not be possible to select this option
> I expected backend validation to happen in this PR

It was already implemented

e2e/fragments/caseDocuments.js
mounikahmcts: As discussed, the page has validation for all the other fields. So, uploading everything

PR #1078, opened 2020-04-15
feat: make "Send to Gatekeeper" available in gatekeeping (FPLA-1599)

lewisbirks:

[FPLA-1599](https://tools.hmcts.net/jira/browse/FPLA-1599)

### Change description ###

Provided the ability to notify other gatekeepers of a case in the gatekeeping state.
This new event is identical to the "Send to gatekeeper" event available in submitted state apart from the name of the event.
Also included is a quick modification to case data to actually match with the case field that it corresponds to however it is never used so probably wasn't a major issue and may even be removed as part of [FPLA-970](https://tools.hmcts.net/jira/browse/FPLA-970).

```
[ ] Yes
[x] No
```


ccd-definition/AuthorisationCaseEvent/AuthorisationCaseEvent.json
```diff
@@ -1328,5 +1328,26 @@
...
+    "CaseEventID": "notifyGatekeeper",
+    "UserRole": "caseworker-publiclaw-courtadmin",
+    "CRUD": "CRU"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseEventID": "notifyGatekeeper",
+    "UserRole": "caseworker-publiclaw-judiciary",
+    "CRUD": "R"
```
mounikahmcts, line 1351: are we not changing live from ?

lewisbirks: Don't believe so, that is on of the fields that is just a constant throughout all of the definitions. Although thinking about it may be able to be used for feature toggling purposes 🤔

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperControllerAboutToStartTest.java
```diff
@@ -0,0 +1,57 @@
...
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails("", SUBMITTED));
+
+        assertThat(callbackResponse.getErrors()).containsExactly("Enter Familyman case number");
+    }
+
+    @Test
+    void shouldNotValidateFamilymanNumberWhenInGatekeepingState() {
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails("", GATEKEEPING));
+
+        assertThat(callbackResponse.getErrors()).isEmpty();
```
mounikahmcts, line 37: As per method name, don't we need to check for method call details here "validateGroupService.validateGroup"?

ccd-definition/AuthorisationCaseEvent/AuthorisationCaseEvent.json
mounikahmcts: okay 👍 i will make a note of it

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperControllerAboutToStartTest.java
lewisbirks: That would probably be a better way of doing it, I just ensured that bad data was sent but I think maybe spying on the validator would be better.

PR #1079, opened 2020-04-16
fix: Correct spacing issues in the SDO [FPLA-1269]

Gogetter:
[FPLA-1269](https://tools.hmcts.net/jira/browse/FPLA-1269)

### Change description ###
Fix for spacing between sections (directions body text) of the SDO document

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -150,7 +151,7 @@ private DocmosisRespondent buildRespondent(RespondentParty respondent) {
                         formattedDirections.add(DocmosisDirection.builder()
                             .assignee(direction.getValue().getAssignee())
                             .title(formatTitle(directionNumber++, direction.getValue(), configOrder.getDirections()))
-                            .body(direction.getValue().getDirectionText())
+                            .body(chomp(direction.getValue().getDirectionText()))
```
tomaszpowroznik, line 154: chomp removes only singel new line, if there are multiple it wont be removed. Should not it be StringUtils.trim instead

Gogetter: thinking about it again `trim` will be better suited just in case [ordersConfig.json](https://github.com/hmcts/fpl-ccd-configuration/blob/5fec14b4325600d4c702cc59f014903f1dad88e7/service/src/main/resources/ordersConfig.json#L19) is updated in future as currently only one `\n` is added to the text.

PR #1082, opened 2020-04-16
fix: updated applicant object validation to accept empty for phone and mobile

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1647

### Change description ###

Validation was throwing an error for empty string values. The code itself was dealing with text characters and white space, replacing it with empty but the validation was failing. I have updated the object level validation to accept empty string and updated tests

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/robotics/Applicant.java
```diff
@@ -16,11 +16,11 @@
     private final String jobTitle;
     private final Address address;
 
-    @Pattern(regexp = "^\\+?\\d+$", message = "is invalid")
+    @Pattern(regexp = "^\\+?\\d*$", message = "is invalid")
```
akomisarek, line 19: As discussed I would remove it, but happy to create separate story for it tomorro.w

PR #1083, opened 2020-04-16
Fix flaky e2e (sign-out -> sign-in scenario)

tomaszpowroznik:


e2e/steps_file.js
```diff
@@ -121,7 +121,7 @@ module.exports = function () {
 
     signOut() {
       this.click('Sign Out');
-      this.wait(2); // in seconds
+      this.waitForText('Sign in', 10);
```
KiranVarma02, line 124: Just FMI, waitFor 10 secs seems okay but do we need to add a limit? should we add limit in other places too?

tomaszpowroznik: default is 1 sec which is fine in most cases, here it is about slow idam signout process. It fixes issue:

2020-04-16T16:00:48.455Z]   1) Case administration after submission

[2020-04-16T16:00:48.455Z]        HMCTS admin revoke case access from representative:

[2020-04-16T16:00:48.455Z]      Execution context was destroyed, most likely because of a navigation.

[2020-04-16T16:00:48.455Z]   

[2020-04-16T16:00:48.455Z]   Scenario Steps:

[2020-04-16T16:00:48.456Z]   

[2020-04-16T16:00:48.456Z]   - I.grabText("#user-name") at /opt/jenkins/workspace/PL_fpl-ccd-configuration_PR-1079/e2e/steps_file.js:38:33

which is specific to slow idam

PR #1087, opened 2020-04-17
feat: notify to more than one gatekeeper (FPLA-970)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-970

### Change description ###
Send To Judge event now accepts a collection of email addresses. Notifications will be sent to each distinct email address provided. 

Screenshot 2020-04-17 at 14 19 57

```
[ ] Yes
[x] No
```


ccd-definition/CaseField.json
```diff
@@ -354,6 +362,15 @@
     "HintText": "For example, joe.bloggs@la.gov.uk",
     "SecurityClassification": "Public"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "gateKeeperEmails",
+    "Label": "Gatekeeper",
+    "FieldType": "Collection",
+    "FieldTypeParameter": "GateKeeperEmail",
```
Jamesnelson117, line 371: Using Complex type as using

`  {
    "LiveFrom": "01/01/2017",
    "ID": "gateKeeperEmails",
    "Label": "Gatekeeper",
    "FieldType": "Collection",
    "FieldTypeParameter": "Email",
  },`

Uses Gatekeeper 1, 2, 3 etc as text input heading rather than email. Also fails to perform client side validation.

ccd-definition/CaseField.json
```diff
@@ -341,7 +341,15 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "ID": "gateKeeperLabel",
```
lewisbirks, line 343: if it isn't too much change could we change the names of all of these to have lower case 'k'.
I know we might not be able to for `gateKeeperEmail` but the others are new/labels

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperControllerAboutToStartTest.java
```diff
@@ -49,19 +55,25 @@ void shouldNotValidateFamilymanNumberWhenInGatekeepingState() {
...
     }
 
     private CaseDetails caseDetails(String familyManNumber, String state) {
         return CaseDetails.builder()
             .id(12345L)
             .data(Map.of(
                 "familyManCaseNumber", familyManNumber,
-                "gateKeeperEmail", "send@spam.com"
+                "gateKeeperEmails", wrapElements(
+                        element(UUID.randomUUID(), EmailAddress.builder().email("test1@gmail.com").build()),
```
lewisbirks, line 75: shouldn't need to do the UUID part, there is an overloaded method without it

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperControllerSubmittedTest.java
```diff
@@ -58,27 +60,40 @@ void shouldNotPublishPopulateStandardDirectionsEventWhenEventIsNotInSubmittedSta
     }
 
     @Test
-    void shouldBuildGatekeeperNotificationTemplateWithCompleteValues() throws Exception {
+    void shouldNotifyMultipleGatekeepersWithExpectedNotificationParameters() throws Exception {
+        postSubmittedEvent(callbackRequest());
+
+        verify(notificationClient, times(1)).sendEmail(
```
lewisbirks, line 66: times(1) is default behaviour and shouldn't be needed

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperController.java
```diff
@@ -60,6 +64,12 @@ public void handleSubmittedEvent(@RequestBody CallbackRequest callbackRequest) {
         if (SUBMITTED.getValue().equals(callbackRequest.getCaseDetails().getState())) {
             applicationEventPublisher.publishEvent(new PopulateStandardDirectionsEvent(callbackRequest, requestData));
         }
-        applicationEventPublisher.publishEvent(new NotifyGatekeeperEvent(callbackRequest, requestData));
+        applicationEventPublisher.publishEvent(new NotifyGatekeepersEvent(callbackRequest, requestData));
+    }
+
+    private List<Element<EmailAddress>> resetGateKeeperEmailCollection() {
+        return List.of(
+            element(UUID.randomUUID(), EmailAddress.builder().email("").build())
```
lewisbirks, line 72: Do you need to set the email to an empty string here? What would happen if it is just null which is the case when .email isn't called?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotifyGatekeeperEventHandler.java
```diff
@@ -1,32 +1,59 @@
...
+
+            parameters.put("gatekeeper_recipients",
+                gatekeeperEmailContentProvider.buildRecipientsLabel(emailList, recipientEmail));
+
+            notificationService.sendEmail(GATEKEEPER_SUBMISSION_TEMPLATE, recipientEmail, parameters,
+                eventData.getReference());
+        });
+    }
+
+    private List<String> getDistinctGatekeeperEmails(List<Element<EmailAddress>> emailCollection) {
```
lewisbirks, line 52: Could we just return the stream here instead of collecting to a list when it isn't really needed

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotifyGatekeeperEventHandlerTest.java
```diff
@@ -55,17 +58,24 @@ void shouldSendEmailToGatekeeper() {
...
+
         given(gatekeeperEmailContentProvider.buildGatekeeperNotification(callbackRequest().getCaseDetails(),
             LOCAL_AUTHORITY_CODE)).willReturn(expectedParameters);
 
         notifyGatekeeperEventHandler.sendEmailToGatekeeper(
-            new NotifyGatekeeperEvent(callbackRequest(), requestData));
+            new NotifyGatekeepersEvent(callbackRequest(), requestData));
 
-        verify(notificationService).sendEmail(
+        verify(notificationService, times(1)).sendEmail(
```
lewisbirks, line 73: times(1) shouldn't be needed again

ccd-definition/CaseField.json
Jamesnelson117: Agreed. Let's do this for email collection and the newly added hint label.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotifyGatekeeperEventHandler.java
lewisbirks: Ignore me, just seen that it is needed in the for each as well as for constructing the loop

Jamesnelson117: Discussed and we decided to keep as it is.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperController.java
Jamesnelson117: Setting to empty string expands the first item as in the collection on page load, as well as reseting the collection. It's not a requirement I guess to expand the collection page load but it's consistent with other events.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/GatekeeperEmailContentProvider.java
```diff
@@ -30,4 +32,16 @@ protected GatekeeperEmailContentProvider(@Value("${ccd.ui.base.url}") String uiB
             .put("localAuthority", config.getLocalAuthorityName(localAuthorityCode))
             .build();
     }
+
+    public String buildRecipientsLabel(List<String> emailList, String recipientEmail) {
+        List<String> filteredEmailList = emailList.stream()
+            .filter(email -> !recipientEmail.equals(email))
+            .collect(Collectors.toList());
```
lewisbirks, line 39: Instead of collecting to an intermediary list, thoughts on collecting to a string straight away? Instead of `Collectors.toList()` you can use `Collectors.joining(", ")` which will still return an empty string if nothing is present in the filtered list

PR #1093, opened 2020-04-20
impr: handle login issues when idam redirection destroy puppeteer context

tomaszpowroznik:
e2e


e2e/helpers/puppeter_helper.js
```diff
@@ -9,6 +9,11 @@ module.exports = class PuppeteerHelpers extends Helper {
     return page.reload();
   }
 
+  waitForNavigationTest() {
```
akomisarek, line 12: Is this used? I don't see any changes invoking this method.

tomaszpowroznik: the last of other options tried, removed

PR #1098, opened 2020-04-20
feat: updated date values for directions in standard direction order (FPLA-1643)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1643

### Change description ###

Updated orders.config with new direction deadlines.
Refactored date logic to always prepopulate direction dates.

```
[ ] Yes
[x] No
```


service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/PopulateStandardDirectionsHandlerTest.java
```diff
@@ -75,222 +99,292 @@
...
+
+    private OrderDefinition getOrderDefinitionContainingTwoDirectionsWithDifferentDeltas() {
+        return OrderDefinition.builder()
+            .directions(ImmutableList.of(
+                directionConfiguration(DIRECTION_TEXT, TWO_DAYS_BEFORE_HEARING),
+                directionConfiguration(DIRECTION_TEXT, THREE_DAYS_BEFORE_HEARING)))
+            .build();
+    }
+
+    private DirectionConfiguration directionConfiguration(String directionText, String deltaValue) {
```
toireasam, line 361: Do we need both of these methods? can we not just have one with time, directionText, delta value passed in?

swalker125: seems silly to be passing in all that stuff when it's not needed, overloading methods isn't a bad thing :)

toireasam: ok makes sense :)

PR #1103, opened 2020-04-21
impr: Improve time based assertions in tests (FPLA-870)

muyinatech:
### JIRA link ###

https://tools.hmcts.net/jira/browse/FPLA-870

### Change description ###
Time based assertions may fail in unit test
Change tests to use date/time from TimeConfiguration.

```
[ ] Yes
[x] No
```


muyinatech:
3 integration tests have failed in Jenkins, but passed locally. Investigating now and will add a comment when fixed.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerMidEventTest.java
```diff
@@ -84,10 +79,10 @@ void midEventShouldGenerateDraftStandardDirectionDocument() {
 
         CaseDetails caseDetails = CaseDetails.builder()
             .data(createCaseDataMap(directions)
-                .put("dateOfIssue", time.now().toLocalDate().toString())
+                .put("dateOfIssue", dateNow().toString())
                 .put("judgeAndLegalAdvisor", JudgeAndLegalAdvisor.builder().build())
                 .put("caseLocalAuthority", "example")
-                .put("dateSubmitted", time.now().toLocalDate().toString())
+                .put("dateSubmitted", dateNow().toString())
```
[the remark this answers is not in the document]
muyinatech: Yup, will remove.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AbstractControllerTest.java
```diff
@@ -195,12 +202,20 @@ AboutToStartOrSubmitCallbackResponse postSubmittedEvent(String filename) {
         return postSubmittedEvent(filename, SC_OK);
     }
 
+    LocalDateTime timeNow() {
```
lewisbirks, line 205: I would maybe call this one just `now` instead. Although we don't have a use for it right now we may in the future have a field that is LocalTime and then it might get confusing

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerAboutToStartTest.java
```diff
@@ -65,34 +57,34 @@ void shouldAddPreviousTimeAsDateOfIssuedWhenInCaseManagementOrder() {
...
-            NOW.plusDays(5).format(DATE_TIME_FORMATTER),
-            NOW.plusDays(2).format(DATE_TIME_FORMATTER));
 
         AboutToStartOrSubmitCallbackResponse response = postAboutToStartEvent(caseDetails);
 
         CaseData caseData = mapper.convertValue(response.getData(), CaseData.class);
 
-        assertThat(getHearingDates(response)).isEqualTo(expected);
-        assertThat(getHearingDates(response)).doesNotContain(NOW.format(DATE_TIME_FORMATTER));
+        assertThat(getHearingDates(caseData)).isEqualTo(List.of(
```
lewisbirks, line 84: This could instead be `containsOnly`

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseManagementOrderProgressionControllerTest.java
```diff
@@ -89,6 +90,11 @@
         super("cmo-progression");
     }
 
+    @BeforeEach
+    void setUp() {
+        futureDate = timeNow().plusDays(1);
```
lewisbirks, line 95: If you make this static you could move it to a `@BeforeClass` and then only need to do it once

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseManagementOrderProgressionControllerTest.java
```diff
@@ -68,16 +69,16 @@
 @WebMvcTest(CaseManagementOrderProgressionController.class)
 @OverrideAutoConfiguration(enabled = true)
 class CaseManagementOrderProgressionControllerTest extends AbstractControllerTest {
-    private static final UUID uuid = randomUUID();
+    private static final UUID ID = randomUUID();
```
lewisbirks, line 72: I would call this `HEARING_ID` just to give a bit more context as to what the uuid is being used for. More indicates an issue with the CMO class though in my opinion.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerMidEventTest.java
```diff
@@ -118,10 +113,10 @@ void midEventShouldMigrateJudgeAndLegalAdvisorWhenUsingAllocatedJudge() {
     void midEventShouldReturnEmptyDirectionsListInStandardDirectionOrder() {
         CaseDetails caseDetails = CaseDetails.builder()
             .data(createCaseDataMap(buildTestDirections())
-                .put("dateOfIssue", time.now().toLocalDate().toString())
+                .put("dateOfIssue", dateNow().toString())
                 .put("judgeAndLegalAdvisor", JudgeAndLegalAdvisor.builder().build())
                 .put("caseLocalAuthority", "example")
-                .put("dateSubmitted", time.now().toLocalDate().toString())
+                .put("dateSubmitted", dateNow().toString())
                 .build())
```
lewisbirks, line 120: Can probably remove the `toString()` like you have with the others

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerAboutToSubmitTest.java
```diff
@@ -55,7 +54,7 @@ void shouldReturnHearingsWhenNoHearingsExistInPast() {
 
     @Test
     void shouldReturnHearingsWhenNoHearingsExistInFuture() {
-        List<Element<HearingBooking>> hearingDetails = newArrayList(bookingWithStartDate(TODAY.plusDays(-5)));
+        List<Element<HearingBooking>> hearingDetails = newArrayList(bookingWithStartDate(timeNow().plusDays(-5)));
```
lewisbirks, line 57: Could we just change this to `timeNow().minusDays(5)`

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/SendDocumentControllerTest.java
```diff
@@ -56,7 +56,6 @@
 @OverrideAutoConfiguration(enabled = true)
 class SendDocumentControllerTest extends AbstractControllerTest {
```
lewisbirks, line 58: Just seen in this test that time is mocked to return a specifc time, don't think that should be necessary.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AbstractControllerTest.java
muyinatech: I can change it to dateTimeNow()

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ChildController.java
```diff
@@ -78,7 +79,7 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
                 .map(Child::getParty)
                 .map(Party::getDateOfBirth)
                 .filter(Objects::nonNull)
-                .filter(dateOfBirth -> dateOfBirth.isAfter(LocalDate.now()))
+                .filter(dateOfBirth -> dateOfBirth.isAfter(time.now().toLocalDate()))
```
lewisbirks, line 82: Do you think it would be worthwhile adding methods similar to what you did in the abstract controller class to the time interface? Could have a `dateNow()` that defaults to calling `now().toLocalDate()`

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -19,13 +22,13 @@
...
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class EmailNotificationHelper {
-    private static final HearingBookingService hearingBookingService = new HearingBookingService();
 
-    private EmailNotificationHelper() {
-    }
+    private final HearingBookingService hearingBookingService;
 
-    public static String buildSubjectLine(final CaseData caseData) {
+    public String buildSubjectLine(final CaseData caseData) {
```
lewisbirks, line 31: Maybe not in this PR so could be a new tech debt ticket but do you think this method and then subsequently the one below should be refactored to only take the required params, that is a lot of data being passed around in case data that is irrelevant to the method

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AbstractControllerTest.java
muyinatech: actually I will change it to now() as suggested, to match the Time interface.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ChildController.java
muyinatech: Yup I will add dateNow() to the time interface.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
```diff
@@ -20,9 +22,12 @@
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.unwrapElements;
 
 @Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class HearingBookingService {
     public static final String HEARING_DETAILS_KEY = "hearingDetails";
 
+    private final Time time;
```
lewisbirks, line 29: Missed a call in `isPastHearing()`

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
muyinatech: I agree and should be handled in a separate ticket.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
muyinatech: Good spot, thanks.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
```diff
@@ -73,6 +78,7 @@
     @BeforeEach
     void setup() {
         given(userDetailsService.getUserName()).willReturn("Emma Taylor");
+        today = time.now().toLocalDate();
```
lewisbirks, line 81: I would say that we could just remove this variable

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseNoteServiceTest.java
```diff
@@ -24,23 +27,32 @@
...
+        service = new CaseNoteService(idamClient, time);
+    }
+
     @Nested
     class BuildCaseNote {
 
         @BeforeEach
-        void setup() {
+        void setUp() {
+            service = new CaseNoteService(idamClient, time);
```
lewisbirks, line 55: This isn't necessary as the outer for each is called before this one

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ChildController.java
lewisbirks: If we do this then we don't need the methods in the abstract controller class.

lewisbirks: Also, if you don't mind increasing the file count even more, maybe the name of `Time` could be better. From that name you might assume it doesn't deal with dates so maybe it is worth while renaming it to DateTime? Then it can have 3 methods
```java
public interface DateTime {
    LocalDateTime now();

    default LocalDate dateNow() {
        return now().toLocalDate();
    }

    default LocalTime timeNow() {
        return now().toLocalTime();
    }
}
```

lewisbirks: Actually thinking about it maybe leave it for a PR straight after this one has merged if you want to do it

muyinatech: I would prefer to defer the change to another PR.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseNoteServiceTest.java
muyinatech: yup will remove

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseManagementOrderProgressionControllerTest.java
```diff
@@ -90,9 +93,9 @@
         super("cmo-progression");
     }
 
-    @BeforeEach
+    @BeforeAll
```
lewisbirks, line 96: Doesn't `@BeforeAll` require a static method?

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/UpcomingHearingsFoundNotificationHandlerTest.java
```diff
@@ -49,7 +52,7 @@
 
     private LocalDate hearingDate;
 
-    @BeforeEach
+    @BeforeAll
```
lewisbirks, line 55: Again, thought it required a static method

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseManagementOrderProgressionControllerTest.java
muyinatech: I couldn't make it a static method as now() is non-static. Added @TestInstance(PER_CLASS) which allows @BeforeAll on non-static.

lewisbirks: I would say it may be better to go with simplicity then and leave it as the `@BeforeEach` then, thoughts?

muyinatech: Ok I will change back to @BeforeEach.

PR #1107, opened 2020-04-22
impr: cmo generation to use java models (FPLA-1488)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1488

### Change description ###

CMO generation using java objects instead of maps

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderGenerationService.java
```diff
@@ -233,42 +235,53 @@ private DocmosisRepresentative getApplicantDetails(String applicantName, Solicit
...
+
+        int directionNumber = 2;
+        ParentsAndRespondentsDirectionAssignee respondentHeader = null;
+        OtherPartiesDirectionAssignee otherHeader = null;
+
+        for (Element<Direction> element : directions) {
+            Direction direction = element.getValue();
+
+            if (direction.getParentsAndRespondentsAssignee() != null) {
+                DocmosisDirection.Builder builder = buildBaseDirection(direction, directionNumber++);
```
lewisbirks, line 253: If you want this could be extracted to outside the `if`, and the add also could be extracted to the end of the `if`

swalker125: will update because that's a good little change

PR #1111, opened 2020-04-22
Replace deprecated idam authenticateUser method

tomaszpowroznik:

PR #1114, opened 2020-04-23
impr: stop use of IdamClient.authenticateUser as this is now deprecated. Now using IdamClient.getAccessToken

Gogetter:
### Change description ###

Stop use of `IdamClient.authenticateUser` as this is now deprecated. Now using `IdamClient.getAccessToken`

```
[ ] Yes
[x] No
```


Gogetter:
Same as #1111

PR #1088, opened 2020-04-17
impr: improve handling of images in the generated documents (FPLA-1640)

muyinatech:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/FPLA-1640

### Change description ###
Make the following improvements to the docmosis templates:-
- Remove hardcoded images from the templates and replace with placeholder images
- Remove base64 passed to the images

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1117, opened 2020-04-23
fix: persist hearing id in cmo object if it already exists (FPLA-1673)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1673

### Change description ###

update prepareCmo method in DraftCMOService with :
`  .id(oldCMO.map(CaseManagementOrder::getId).orElse(idFromDynamicList))`

Updated CMO tests

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerMidEventTest.java
```diff
@@ -34,31 +54,32 @@
...
-        caseDetails.getData().put(CASE_MANAGEMENT_ORDER_LOCAL_AUTHORITY.getKey(),
-            CaseManagementOrder.builder().build());
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postMidEvent(getCaseDetails());
 
-        AboutToStartOrSubmitCallbackResponse callbackResponse = postMidEvent(caseDetails);
+        verify(uploadService).uploadPDF(PDF, "draft-case-management-order.pdf");
 
-        verify(uploadDocumentService).uploadPDF(PDF, "draft-case-management-order.pdf");
+        //need to pass in captor for draft image string
+        assertThat(captor.getValue()).isEqualToComparingFieldByField(expectedTemplateData(captor.getValue()));
```
lewisbirks, line 82: I feel like there has to be nicer way of doing this, maybe in this assert we can ignore the draft image string and then have another assert to just check that it isn't empty

lewisbirks: Also either in this one or in #1088 whichever gets merged first this can then be changed as we will know the exact value

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftCMOControllerTest.java
```diff
@@ -77,83 +90,230 @@
...
+        given(uploadService.uploadPDF(any(), any())).willReturn(document());
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postMidEvent(getCaseDetails());
+
+        verify(uploadService).uploadPDF(pdf, "draft-case-management-order.pdf");
+
+        assertThat(callbackResponse.getData()).containsKey(CASE_MANAGEMENT_ORDER_LOCAL_AUTHORITY.getKey());
+        assertThat(getDocumentReference(callbackResponse)).isEqualTo(expectedDocument());
+        //need to pass in captor for draft image string
+        assertThat(captor.getValue()).isEqualToComparingFieldByField(expectedTemplateData(captor.getValue()));
```
lewisbirks, line 141: Same as before

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderGenerationService.java
```diff
@@ -132,7 +132,7 @@ private String getScheduleProvided(CaseManagementOrder caseManagementOrder) {
 
     private DocmosisJudgeAndLegalAdvisor getJudgeAndLegalAdvisorData(JudgeAndLegalAdvisor judgeAndLegalAdvisor) {
         return DocmosisJudgeAndLegalAdvisor.builder()
-            .judgeTitleAndName(defaultIfBlank(formatJudgeTitleAndName(judgeAndLegalAdvisor), DEFAULT))
+            .judgeTitleAndName(formatJudgeTitleAndName(judgeAndLegalAdvisor))
```
lewisbirks, line 135: Do we still not want to display the default string in the draft?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/robotics/SampleRoboticsTestDataHelper.java
```diff
@@ -61,7 +61,7 @@ public static RoboticsData invalidRoboticsDataWithZeroOwningCourt() {
                     .postcode("BT22 2345")
                     .country("UK")
                     .build())
-                .relationshipToChild(null)
+                .relationshipToChild("Cousin")
```
lewisbirks, line 64: Necessary change?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerMidEventTest.java
swalker125: this will go in first, I am happy to address separately, however I believe this is fine for now

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderGenerationService.java
swalker125: formatJudgeTitleAndName method handles

service/src/test/java/uk/gov/hmcts/reform/fpl/service/robotics/SampleRoboticsTestDataHelper.java
swalker125: yes, relationshipToChild is mandatory field and will never be null. So this is unrealistic scenario

lewisbirks: Ahh I see you have updated it in the populated case details

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseManagementOrderGenerationService.java
swalker125: `  public static String formatJudgeTitleAndName(JudgeAndLegalAdvisor judgeAndLegalAdvisor) {
        return Optional.ofNullable(judgeAndLegalAdvisor)
            .filter(judge -> judge.getJudgeTitle() != null)
            .map(JudgeAndLegalAdvisorHelper::mapJudgeOrAdvisor)
            .orElse("");
    }`

lewisbirks: Somehow in my head had the 2 constants the wrong way round, thinking it would be the `HEARING_EMPTY_PLACEHOLDER`. Need more sleep

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderControllerMidEventTest.java
lewisbirks: If this is being merged in first then it doesn't matter as we can just sort it out in that other story

PR #1119, opened 2020-04-23
impr:Refactor e2e sign in

tomaszpowroznik:


e2e/tests/hmctsCompliesOnBehalfOfOtherPartiesAfterSDOIsIssued_test.js
```diff
@@ -6,56 +6,42 @@ let caseId;
...
+  await draftStandardDirectionsEventPage.useAllocatedJudge('Bob Ross');
+  await draftStandardDirectionsEventPage.enterDatesForDirections(directions[0]);
+  draftStandardDirectionsEventPage.markAsFinal();
+  await I.completeEvent('Save and continue');
 });
 
-Scenario('hmcts admin complies with directions on behalf of other parties', async (I, caseViewPage, complyOnBehalfOfOthersEventPage) => {
-  await I.signIn(config.hmctsAdminEmail, config.hmctsAdminPassword);
-  await I.navigateToCaseDetails(caseId);
+Before(async I => await I.navigateToCaseDetailsAs(config.hmctsAdminUser, caseId));
```
akomisarek, line 42: Do we need to do it (changing user) before every test? I am afraid if smoeone adds new test this might be not required.

e2e/tests/smoke_test.js
```diff
@@ -1,11 +1,7 @@
 /* global process */
-// const config = require('../config.js');
 
 Feature('Smoke tests @smoke-tests');
```
akomisarek, line 3: Worth discussing maybe our smoke tests should at least create a case, we would ensure our pods are started properly and not only ccd, but out of scpoe here.

e2e/tests/bulkScanUserUploadsScannedDocuments_test.js
```diff
@@ -5,30 +5,19 @@ let caseId;
...
+  submitApplicationEventPage.giveConsent();
+  await I.completeEvent('Submit');
 
+  await I.navigateToCaseDetailsAs(config.hmctsAdminUser, caseId);
   await caseViewPage.goToNewActions(config.administrationActions.bulkScan);
   attachScannedDocsEventPage.enterScannedDocument(scannedDocument, config.testFile);
   await I.click('Continue');
   handleSupplementaryEvidenceEventPage.handleSupplementaryEvidence();
   await I.completeEvent('Submit');
-  I.seeEventSubmissionConfirmation(config.administrationActions.bulkScan);
```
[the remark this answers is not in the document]
tomaszpowroznik: I.completeEvent check if there is green banner, that is enough, if for some reason it is not test will fail.

e2e/tests/smoke_test.js
tomaszpowroznik: yep, but it is not this ticket

e2e/tests/hmctsCompliesOnBehalfOfOtherPartiesAfterSDOIsIssued_test.js
tomaszpowroznik: agree, done

PR #1125, opened 2020-04-24
fix: hide nested field in children event if initial question set to 'No' (FPLA-1095)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-1095

### Change description ###

Hide the 'Which court are you applying to question?' field if the initial question 'Are you considering adoption at this stage?' is set to 'No'. The field we are hiding is nested and currently shows regardless for both yes/no responses on the initial question.

```
[ ] Yes
[X] No
```


joez-f:
Screenshot 2020-04-24 at 15 23 22 with yes selected
Screenshot 2020-04-24 at 15 25 13 with no selected

PR #1126, opened 2020-04-24
refactor: casePersonalisedContent provider to use java objects (FPLA-1656)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1656

### Change description ###
casePersonalisedContentProvider class renamed to SharedNotifyContentProvider
SharedNotifyContentProvider now uses java objects

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -273,39 +270,42 @@ private CaseDetails enableSendToCtscOnCaseDetails(YesNo enableCtsc) {
...
-            .putAll(getCommonParameters())
-            .put("dataPresent", "No")
-            .put("fullStop", "Yes")
-            .put("ordersAndDirections", "")
-            .put("timeFramePresent", "No")
-            .put("timeFrameValue", "")
-            .put("urgentHearing", "No")
-            .put("nonUrgentHearing", "No")
-            .put("firstRespondentName", "");
+    private <T extends SharedNotifyTemplate> T getInCompleteParameters(T template) {
```
lewisbirks, line 289: Incomplete

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -273,39 +270,42 @@ private CaseDetails enableSendToCtscOnCaseDetails(YesNo enableCtsc) {
...
+        return updatedTemplate;
     }
 
-    private Map<String, Object> getCommonParameters() {
-        return Map.of(
-            "localAuthority", "Example Local Authority",
-            "reference", CASE_REFERENCE.toString(),
-            "caseUrl", String.format("http://fake-url/case/%s/%s/%s", JURISDICTION, CASE_TYPE, CASE_REFERENCE)
-        );
+    private <T extends SharedNotifyTemplate> T setCommonTemplateParameters(T template) {
```
lewisbirks, line 304: You don't need to return anything for this method if you want you are going to be effecting the same object so unless you call this method directly from the test you don't need to return the object as the prior method will already have the reference for it

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperControllerSubmittedTest.java
```diff
@@ -64,31 +65,32 @@ void shouldNotifyMultipleGatekeepersWithExpectedNotificationParameters() throws
 
         verify(notificationClient).sendEmail(
             GATEKEEPER_SUBMISSION_TEMPLATE, GATEKEEPER_EMAIL,
-            buildExpectedParameters(CAFCASS_EMAIL), "12345");
+            getTemplate(CAFCASS_EMAIL).toMap(mapper), "12345");
 
         verify(notificationClient).sendEmail(
             GATEKEEPER_SUBMISSION_TEMPLATE, CAFCASS_EMAIL,
-            buildExpectedParameters(GATEKEEPER_EMAIL), "12345");
+            getTemplate(GATEKEEPER_EMAIL).toMap(mapper), "12345");
```
lewisbirks, line 72: Why not just verify against the NotifyTemplate version of the method call? You can capture the data in an ArgumentCaptor and compare against it there. I did something similar in the handler test. Although thinking about it, does this test need to worry about the content of the email sent, could you just verify that an email was sent twice?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/notify/SharedNotifyTemplate.java
```diff
@@ -0,0 +1,22 @@
...
+    private String dataPresent;
+    private String fullStop;
+    private String timeFramePresent;
+    private String timeFrameValue;
+    private String urgentHearing;
+    private String nonUrgentHearing;
+    private String firstRespondentName;
+    private String reference;
+    private String caseUrl;
+    private String localAuthority;
```
lewisbirks, line 21: If we make all of these protected/package private then we don't need a getter and can just call the values directly in the other classes

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperControllerSubmittedTest.java
Jamesnelson117: Potentially no, we test the content in the handler itself. I'm happy to remove the assertion around content for the int test.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
Jamesnelson117: Good spot thanks

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -74,88 +76,83 @@ void shouldReturnUnsuccessfulResponseWithMalformedData() {
...
             CAFCASS_EMAIL,
-            completeCafcassParameters,
+            completeCafcassParameters.toMap(mapper),
             CASE_REFERENCE.toString());
 
         verify(notificationClient, never()).sendEmail(
             HMCTS_COURT_SUBMISSION_TEMPLATE,
             "FamilyPublicLaw+ctsc@gmail.com",
-            completeHmctsParameters,
+            completeHmctsParameters.toMap(mapper),
```
lewisbirks, line 102: Shouldn't need the toMap part, sendEmail has an overloaded method that accepts the object which has a base type of `NotifyData` instead of the map, same in other places in this class as well

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/base/NotifyTemplateContentProvider.java
```diff
@@ -2,47 +2,59 @@
...
 import static org.apache.commons.lang.StringUtils.uncapitalize;
 import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
+import static uk.gov.hmcts.reform.fpl.enums.YesNo.NO;
+import static uk.gov.hmcts.reform.fpl.enums.YesNo.YES;
 import static uk.gov.hmcts.reform.fpl.utils.EmailNotificationHelper.formatCaseUrl;
 import static uk.gov.hmcts.reform.fpl.utils.PeopleInCaseHelper.getFirstRespondentLastName;
 
-public abstract class CasePersonalisedContentProvider extends AbstractEmailContentProvider {
-    protected CasePersonalisedContentProvider(String uiBaseUrl, ObjectMapper mapper) {
+public abstract class NotifyTemplateContentProvider extends AbstractEmailContentProvider {
```
lewisbirks, line 26: `SharedNotifyContentProvider`?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/base/SharedNotifyContentProvider.java
```diff
@@ -2,47 +2,59 @@
...
-            .put("timeFrameValue", uncapitalize(timeFrame.orElse("")))
-            .put("urgentHearing", timeFrame.isPresent() && timeFrame.get().equals("Same day") ? "Yes" : "No")
-            .put("nonUrgentHearing", timeFrame.isPresent() && !timeFrame.get().equals("Same day") ? "Yes" : "No")
-            .put("firstRespondentName", getFirstRespondentLastName(caseData.getRespondents1()))
-            .put("reference", String.valueOf(caseId))
-            .put("caseUrl", formatCaseUrl(uiBaseUrl, caseId));
+        // What if we just add the list no matter if it is empty or not,
+        // surely the parts below determine if it is going to be shown or not
+        template.setOrdersAndDirections(!ordersAndDirections.isEmpty() ? ordersAndDirections : List.of(""));
+        template.setDataPresent(!ordersAndDirections.isEmpty() ? YES.getValue() : NO.getValue());
```
lewisbirks, line 45: Did you investigate adding JsonProperty tags to the YES and NO enums so that you can directly pass them instead of getting the value attached?

lewisbirks: Or did it just end up breaking things in other places?

Jamesnelson117: Yes I did and reverted :/ It impacted the placementController and was not keen on adding out of scope changes.

lewisbirks:
2 conditions are not covered in the SharedNotificationContentProvider, probably should be, especially as we now have another ticket to increase conditional coverage

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/base/SharedNotifyContentProvider.java
```diff
@@ -39,11 +39,9 @@ protected SharedNotifyContentProvider(String uiBaseUrl, ObjectMapper mapper) {
...
             .filter(StringUtils::isNotBlank);
 
-        // What if we just add the list no matter if it is empty or not,
-        // surely the parts below determine if it is going to be shown or not
-        template.setOrdersAndDirections(!ordersAndDirections.isEmpty() ? ordersAndDirections : List.of(""));
-        template.setDataPresent(!ordersAndDirections.isEmpty() ? YES.getValue() : NO.getValue());
-        template.setFullStop(!ordersAndDirections.isEmpty() ? NO.getValue() : YES.getValue());
+        template.setOrdersAndDirections(ordersAndDirections);
+        template.setDataPresent(YES.getValue());
+        template.setFullStop(NO.getValue());
```
lewisbirks, line 44: Maybe not now but should we create a ticket to update the notification + this code to remove these redundant fields? Seeing as it just appears to be constant now there isn't much point in having extra conditional logic in the notification template.

lewisbirks: Also if they are always going to be Yes and No respectively, we could just set that directly in the model.

Jamesnelson117: Agreed but I also think we should wait until https://tools.hmcts.net/jira/browse/FPLA-1513 has been played also as it will likely result in notify template changes also.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -74,88 +79,83 @@ void shouldReturnUnsuccessfulResponseWithMalformedData() {
...
     }
 
     @Test
     void shouldBuildNotificationTemplatesWithValuesMissingInCallback() throws Exception {
         CaseDetails caseDetails = enableSendToCtscOnCaseDetails(NO);
 
-        Map<String, Object> expectedHmctsParameters = getInCompleteParameters()
-            .put("court", FAMILY_COURT)
-            .build();
+        SubmitCaseHmctsTemplate expectedHmctsParameters = getIncompleteParameters(new SubmitCaseHmctsTemplate());
```
lewisbirks, line 113: Do you think it would be better to directly change to map here instead of doing the conversion twice in the verify? Same for cafcass one as well.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -74,88 +79,83 @@ void shouldReturnUnsuccessfulResponseWithMalformedData() {
...
             CASE_REFERENCE.toString());
     }
 
     @Test
     void shouldSendNotificationToCtscAdminWhenCtscIsEnabledWithinCaseDetails() throws Exception {
         CaseDetails caseDetails = enableSendToCtscOnCaseDetails(YES);
-        Map<String, Object> expectedHmctsParameters = getInCompleteParameters()
-            .put("court", FAMILY_COURT)
-            .build();
+        SubmitCaseHmctsTemplate expectedHmctsParameters = getIncompleteParameters(new SubmitCaseHmctsTemplate());
```
lewisbirks, line 143: Same here

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
Jamesnelson117: Good idea but we set further properties on the returned templates which are not shared across templates.

lewisbirks: maybe not directly here but maybe a bit later. Just so that we don't have to call the `.toMap()` multiple times.

lewisbirks: in this test we are converting the HMCTS parameters to a map twice when we could just do it once

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -99,63 +106,50 @@ void shouldBuildNotificationTemplatesWithCompleteValues() throws Exception {
...
-            .put("cafcass", CAFCASS_COURT)
-            .build();
-
         postSubmittedEvent(caseDetails);
 
         verify(notificationClient).sendEmail(
             HMCTS_COURT_SUBMISSION_TEMPLATE,
             HMCTS_ADMIN_EMAIL,
-            expectedHmctsParameters,
+            getExpectedIncompleteHmctsParameters(),
```
lewisbirks, line 121: If the aim was to reduce the number of times we are creating the object, by calling the method twice it is still being created twice

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -74,20 +79,22 @@ void shouldReturnUnsuccessfulResponseWithMalformedData() {
 
     @Test
     void shouldBuildNotificationTemplatesWithCompleteValues() throws Exception {
-        Map<String, Object> completeHmctsParameters = getCompleteParameters()
-            .put("court", FAMILY_COURT)
-            .build();
+        SubmitCaseHmctsTemplate submitCaseHmctsTemplate = getCompleteParameters(new SubmitCaseHmctsTemplate());
```
lewisbirks, line 82: Could we do similar to what you have done for the incompleteParams here?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/content/GatekeeperEmailContentProviderTest.java
```diff
@@ -79,10 +118,17 @@ void shouldFormatRecipientLabelCorrectlyWhenMultipleGatekeeperEmailsArePresent()
 
     @Test
     void shouldReturnEmptyStringIfOnlyOneRecipient() {
-        List<String> gatekeeperEmails = List.of("JohnSmith@gmail.com");
+        List<String> gatekeeperEmails = ImmutableList.of("JohnSmith@gmail.com");
         String formattedMessage = gatekeeperEmailContentProvider.buildRecipientsLabel(gatekeeperEmails,
             "JohnSmith@gmail.com");
 
-        assertThat(formattedMessage).isEmpty();
+        assertThat(formattedMessage).isEqualTo("");
```
lewisbirks, line 125: Seems like unnecessary changes, I would revert back to what master has for this

PR #1132, opened 2020-04-28
impr: log unsuccessful submitted request [FPLA-NA]

Gogetter:
### Change description ###

Log unsuccessful `submitted` request to `/callback/add-case-number`

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AddCaseNumberController.java
```diff
@@ -45,7 +47,11 @@ public void handleSubmittedEvent(@RequestBody CallbackRequest callbackRequest) {
         CaseData previousData = mapper.convertValue(callbackRequest.getCaseDetailsBefore().getData(), CaseData.class);
 
         if (isEmpty(previousData.getFamilyManCaseNumber())) {
-            applicationEventPublisher.publishEvent(new CaseNumberAdded(callbackRequest.getCaseDetails()));
+            try {
+                applicationEventPublisher.publishEvent(new CaseNumberAdded(callbackRequest.getCaseDetails()));
+            } catch (Exception exc) {
+                log.error("Robotics notification failed", exc);
+            }
```
tomaszpowroznik, line 54: i think this one is not really needed as it will be logged anyway. I rather though that there should be log when controller does not sent notification due to if condition

Gogetter: updated

PR #1135, opened 2020-04-28
impr: Improve test coverage

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-1678
coverage

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingValidatorService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingValidatorService.java
@@ -1,0 +1,32 @@
+package uk.gov.hmcts.reform.fpl.service;
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.fpl.model.HearingBooking;
+import uk.gov.hmcts.reform.fpl.validation.groups.HearingBookingDetailsGroup;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class HearingBookingValidatorService {
+
+    private final ValidateGroupService validateGroupService;
+
+    public List<String> validateHearingBookings(List<HearingBooking> hearingDetails) {
+        final List<String> errors = new ArrayList<>();
+
+        boolean multipleHearings = hearingDetails.size() > 1;
+
+        for (int i = 0; i < hearingDetails.size(); i++) {
+            HearingBooking hearingDetail = hearingDetails.get(i);
+            for (String message : validateGroupService.validateGroup(hearingDetail, HearingBookingDetailsGroup.class)) {
+                errors.add(multipleHearings ? String.format("%s for hearing %d", message, i + 1) : message);
+            }
+        }
+        return errors;
+    }
+
+}
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
@@ -3,7 +3,6 @@ package uk.gov.hmcts.reform.fpl.service.robotics;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.RequiredArgsConstructor;
-import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
@@ -34,6 +33,7 @@ import java.util.Set;
 
 import static com.google.common.collect.ImmutableSet.of;
 import static java.util.Arrays.asList;
+import static java.util.Optional.ofNullable;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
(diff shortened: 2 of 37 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
```diff
@@ -64,9 +64,8 @@ public RoboticsData prepareRoboticsData(final CaseData caseData, final Long case
             .internationalElement(hasInternationalElement(caseData.getInternationalElement()))
             .allocation(isNotEmpty(caseData.getAllocationProposal())
                 && isNotBlank(caseData.getAllocationProposal().getProposal())
-                ? caseData.getAllocationProposal().getProposal() :  null)
-            .issueDate(isNotEmpty(caseData.getDateSubmitted())
-                ? formatLocalDateToString(caseData.getDateSubmitted(), "dd-MM-yyyy") : "")
+                ? caseData.getAllocationProposal().getProposal() : null)
+            .issueDate(formatDate(caseData.getDateSubmitted(), "dd-MM-yyyy"))
```
lewisbirks, line 68: Do we already have a constant for that format in the DateFormatterHelper? If not do you think it would be sensible to add one?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
```diff
@@ -168,7 +166,7 @@ private Respondent buildRespondent(final uk.gov.hmcts.reform.fpl.model.Responden
             .gender(convertStringToGender(respondentParty.getGender()))
             .address(convertAddress(respondentParty.getAddress()).orElse(null))
             .relationshipToChild(respondentParty.getRelationshipToChild())
-            .dob(formatDob(respondentParty.getDateOfBirth()))
+            .dob(formatDate(respondentParty.getDateOfBirth(), "d-MMM-y"))
```
lewisbirks, line 169: Same here

service/src/test/java/uk/gov/hmcts/reform/fpl/service/HearingBookingValidatorServiceTest.java
```diff
@@ -0,0 +1,75 @@
...
+    LocalValidatorFactoryBean.class})
+class HearingBookingValidatorServiceTest {
+
+    @MockBean
+    private ValidateGroupService validateGroupService;
+
+    @Autowired
+    private HearingBookingValidatorService service;
+
+
```
lewisbirks, line 31: blank line

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/content/AbstractEmailContentProviderTest.java
```diff
@@ -1,7 +1,18 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import static uk.gov.hmcts.reform.fpl.service.email.content.AbstractEmailContentProviderTest.BASE_URL;
 import static uk.gov.hmcts.reform.fpl.utils.EmailNotificationHelper.formatCaseUrl;
 
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {JacksonAutoConfiguration.class})
+
```
lewisbirks, line 14: Blank line

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/PaymentServiceTest.java
```diff
@@ -52,6 +52,7 @@
 
 @ExtendWith(SpringExtension.class)
 @ContextConfiguration(classes = {PaymentService.class})
+@TestPropertySource(properties = {"payment.site_id=TEST_SITE_ID"})
```
lewisbirks, line 55: Do you think it would be useful to add a unit test profile so that we don't need to add this and other properties to the tops of files. We could then also remove the TestConfig class (can't remember name) which sets up a load of test beans and just add what is needed to the context config.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingValidatorService.java
```diff
@@ -0,0 +1,32 @@
...
+
+        for (int i = 0; i < hearingDetails.size(); i++) {
+            HearingBooking hearingDetail = hearingDetails.get(i);
+            for (String message : validateGroupService.validateGroup(hearingDetail, HearingBookingDetailsGroup.class)) {
+                errors.add(multipleHearings ? String.format("%s for hearing %d", message, i + 1) : message);
+            }
+        }
+        return errors;
+    }
+
```
lewisbirks, line 31: Blank line

service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
tomaszpowroznik: It is an option, but this is format required by external service and used only by it. If it is common format then can be used in other places and changed when needed, which o course cause test to fail which is good. Problematic is also naming of such constant as we have already cople of them, how to name them DATE_SHORT_MONTH, it is already there for different format, ROBOTICS_DATE does not really fit into common class?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/HearingBookingValidatorServiceTest.java
tomaszpowroznik: done

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/PaymentServiceTest.java
tomaszpowroznik: Interesting point, but not sure if such approach is correct. Ideally unit test should be real unit tests fully isolated and  should not depends on env or bootstrap spring context (or at least should inject mocks and not real dependencies) Additionally with such approach we have limited options to unit test with different injected parameters, e.g if we have flag in config that turn on/off some logic, like feature toggle, with single test profile it would be hard to provide different variations to be test.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingValidatorService.java
tomaszpowroznik: done

service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
lewisbirks: Fair enough, it definitely does appear to be a different format to what we would normally use so I am happy for it to not be a constant in that class, will sonarqube complain about it not being a constant in this class though?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/payment/PaymentServiceTest.java
lewisbirks: Yeah, that was my sort of concern as well.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/robotics/RoboticsDataService.java
tomaszpowroznik: sonar complains when there are 3 repetition of string i guess

PR #1140, opened 2020-04-29
impr: upgrade ccd chart (FPLA-1620)

muyinatech:
https://tools.hmcts.net/jira/browse/FPLA-1620

### Change description ###
Upgrade CCD helm chart version to 4.2.3 and update values.preview template accordingly.

```
[ ] Yes
[x] No
```


charts/fpl-case-service/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: fpl-case-service
 apiVersion: v1
 home: https://github.com/hmcts/fpl-ccd-configuration
-version: 1.10.11
+version: 1.10.12
```
tomaszpowroznik, line 4: i would update minor version instead of patch

charts/fpl-case-service/values.yaml
```diff
@@ -2,6 +2,9 @@ tags:
   fpl-ccd-stack: false
   ccd-idam-pr: false
```
tomaszpowroznik, line 3: is ccd-idam-pr needed? is it not idam-pr now?

charts/fpl-case-service/values.yaml
```diff
@@ -50,3 +53,6 @@ java:
         - AppInsightsInstrumentationKey
         - scheduler-db-password
   applicationPort: 4000
+
+ccd:
```
tomaszpowroznik, line 57: tags:
  fpl-ccd-stack: false 
in first line should turn off ccd dependency, so this maybe is not needed

charts/fpl-case-service/values.yaml
muyinatech: Not need, so I will remove.

charts/fpl-case-service/Chart.yaml
muyinatech: Yup, I will change it.

charts/fpl-case-service/values.yaml
muyinatech: Agreed, so I will remove.

charts/fpl-case-service/values.yaml
```diff
@@ -1,6 +1,8 @@
 tags:
   fpl-ccd-stack: false
-  ccd-idam-pr: false
+
+idam-pr:
```
tomaszpowroznik, line 4: i would remove this one as 
tags:
 fpl-ccd-stack: false
should disable entire ccd deployment

PR #1148, opened 2020-04-30
impr: hide pba number from logs (FPLA-1684)

lewisbirks:

[FPLA-1684](https://tools.hmcts.net/jira/browse/FPLA-1684)

### Change description ###

Excluded the value from Lombok's generated `toString()`

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fnp/model/payment/CreditAccountPaymentRequest.java
```diff
@@ -36,4 +38,15 @@
     private String siteId;
     @JsonProperty("fees")
     private List<FeeDto> fees;
+
+    @Override
+    public String toString() {
```
Gogetter, line 43: just a suggestion we can use 
```java
return new ToStringBuilder(this, ToStringStyle.MULTI_LINE_STYLE)
            .append(...)
...
```

PR #1156, opened 2020-05-01
feat: only show chosen SDO directions in tab (FPLA-1433)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-1433

### Change description ###

Only show chosen SDO directions in the tab view. This is done by removing directions that have directionNeeded set to "No". Also hid the directionNeeded question in the tab view.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonDirectionService.java
```diff
@@ -238,6 +238,33 @@ public void addEmptyDirectionsForAssigneeNotInMap(Map<DirectionAssignee, List<El
...
+     * @param directions any list of directions.
+     * @return a list of directions that have hidden tab values modified.
+     */
+    public List<Element<Direction>> changeSdoTabView(List<Element<Direction>> directions, String status) {
+        return directions.stream()
+            .map(element -> element(element.getId(), element.getValue().toBuilder()
+                .tabView(status)
+                .build()))
+            .collect(toList());
+    }
```
swalker125, line 253: this could be a method on the Direction java object, i.e a setter for this status... I am not sure I am a fan of tabView field at all though

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CommonDirectionServiceTest.java
```diff
@@ -177,6 +177,38 @@ void numberDirections_shouldApplyCorrectNumberingWhenDirectionsAreMarkedAsNotNee
...
+            .directionNeeded("Yes")
+            .assignee(LOCAL_AUTHORITY)
+            .tabView(null)
+            .build();
+
+        List<Direction> directions = unwrapElements(
+            service.changeSdoTabView(newArrayList(element(direction)), "Yes"));
+
+        List<Direction> expectedDirections = unwrapElements(
+            newArrayList(element(direction.toBuilder().tabView("Yes").build())));
```
swalker125, line 207: unwrapElements on element? why not just newArrayList(direction....)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonDirectionService.java
```diff
@@ -238,6 +238,19 @@ public void addEmptyDirectionsForAssigneeNotInMap(Map<DirectionAssignee, List<El
...
 
+    /**
+     * Removes directions where directionNeeded is set to No. Does not remove custom directions.
+     *
+     * @param directions any list of directions.
+     * @return a list of directions that are marked as needed.
+     */
+    public List<Element<Direction>> removeUnnecessaryDirections(List<Element<Direction>> directions) {
+        return directions.stream()
+            .filter(element -> "Yes".equals(element.getValue().getDirectionNeeded()) || "Yes".equals(
```
lewisbirks, line 249: Maybe refactor out the condition into its own method so that it is a bit easier to read then the flow in this method could just be
```java
        return directions.stream()
            .filter(this::showDirection)
            .collect(toList());
```

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CommonDirectionServiceTest.java
joez-f: woops :D resolved

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonDirectionService.java
joez-f: as discussed, we found a better solution 💪

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CommonDirectionService.java
joez-f: 👍

PR #1162, opened 2020-05-04
impr: common extract fields to map method for cmo (FPLA-1483)

swalker125:
Gonna be making a few different PRs for this to do it in smaller chunks!

https://tools.hmcts.net/jira/browse/FPLA-1483

### Change description ###
Extracted a common extract fields from cmo method

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseManagementOrder.java
```diff
@@ -68,4 +73,23 @@ public void setNextHearingFromDynamicElement(HearingDateDynamicElement nextHeari
...
+            data.put(SCHEDULE.getKey(), schedule);
+        }
+
+        if (recitals != null) {
+            data.put(RECITALS.getKey(), recitals);
+        }
+
+        if (action != null) {
+            data.put(ORDER_ACTION.getKey(), action);
+        }
```
lewisbirks, line 91: I think you could add even if null, I think null will only mean that the case field in not populated in the UI which is fine. It also seems to be the previous behaviour

swalker125: I was getting some null pointer exceptions in some tests, so rather than changing the tests I thought it better to rework the code to get those tests passing

akomisarek: I think it's better not put null values, afaik some maps implementation don't allow nulls.

I would prefer not having keys if the value is is null. @swalker125 I like you like lambdas so maybe yoou culd go with:

data.compute(ORDER_ACTION.getKey(),(k, v) -> action);

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ActionCaseManagementOrderController.java
```diff
@@ -66,13 +66,10 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
                 .build();
         }
 
-        caseDetails.getData().putAll(caseManagementOrderService
-            .extractMapFieldsFromCaseManagementOrder(caseManagementOrder));
-
         draftCMOService.prepareCustomDirections(caseDetails, caseManagementOrder);
 
+        caseDetails.getData().putAll(caseManagementOrder.getCCDFields());
```
akomisarek, line 71: Previous implementation had NPE guard inside. But I guess it's not possible to have null here in Action?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/DraftCMOServiceTest.java
```diff
@@ -94,25 +93,19 @@ void shouldNotReturnHearingDatesWhenHearingDateIsInThePast() {
         hearingDetails = createHearingBookingsFromInitialDate(time.now().minusDays(10));
         caseManagementOrder = CaseManagementOrder.builder().build();
 
-        Map<String, Object> data = draftCMOService.extractCaseManagementOrderVariables(
-            caseManagementOrder, hearingDetails);
+        DynamicList data = service.getHearingDateDynamicList(hearingDetails, caseManagementOrder);
```
akomisarek, line 96: This refactoring doesn't have anything to do with the ticket, but it's fine I don't mind.

swalker125: so this isn't a full implementation of 1483, I just keep getting hit with horrific merge conflicts when I'm working on large tickets. So I am trying a little PR approach. As we have removed the extractCaseManagementOrderVariables method from draftCMO service, I have refactored what the tests were testing (getHearingDateDynamicList method) so that this functionality is still tested. To be revisited in further PRs

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseManagementOrder.java
swalker125: you can't add null value to Map in java so you would never have key with null. Is compute not a bit heavy and OTT for this? We're start a new map when building this, so the values will never already be there or need to be checked etc

akomisarek: It is OTT. I mean it wouldn't be performance problem for sure, just suggestion if you would like to avoid if's, but you know I am never against good old if (or for! ;))

akomisarek: Btw, of course you can add null value to Map in Java! It can be used as both key and value :( 

```
        Map<String, Object> map = new HashMap<>();
        map.put("key",null);
        map.put(null, "test");

        System.out.println(map.get(null));
        System.out.println(map.get("key"));
```

This code doesn't throw any exception.

PR #1163, opened 2020-05-04
impr: simplified dynamic list logic for hearing dates (FPLA-NA)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1483

### Change description ###

Reworked the dynamic list population code

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/dynamic/DynamicList.java
```diff
@@ -26,23 +25,6 @@
...
-     * @return a {@link DynamicList} to be sent to CCD
-     */
-    public static <T extends DynamicElementIndicator> DynamicList toDynamicList(List<T> elements,
-                                                                                DynamicListElement selected) {
-        List<DynamicListElement> items = elements.stream()
-            .map(DynamicElementIndicator::toDynamicElement)
-            .collect(Collectors.toList());
-
-        return DynamicList.builder().listItems(items).value(selected).build();
-    }
```
akomisarek, line 44: I think after those changes interface DynamicElementIndicator is no longer used please remove.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -118,6 +90,42 @@ public void prepareCustomDirections(CaseDetails caseDetails, CaseManagementOrder
...
+            .orElse(DynamicListElement.EMPTY);
+
+        return DynamicList.builder()
+            .listItems(values)
+            .value(selectedValue)
+            .build();
+    }
+
+    //TODO: isAfter method in hearing booking does not filter out todays date as Time is set before LocalDate.now()
+    // is evaluated
```
akomisarek, line 107: Goood point, I think they are discussing to remove that filtering altogether.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/DraftCMOService.java
```diff
@@ -118,6 +90,42 @@ public void prepareCustomDirections(CaseDetails caseDetails, CaseManagementOrder
...
+            .label(formatLocalDateToString(element.getValue().getStartDate().toLocalDate(), FormatStyle.MEDIUM))
+            .code(element.getId())
+            .build();
+    }
+
+    private DynamicListElement getPreselectedDate(List<DynamicListElement> list, UUID id) {
+        return list.stream()
+            .filter(item -> item.getCode().equals(id))
+            .findFirst()
+            .orElse(DynamicListElement.EMPTY);
```
akomisarek, line 126: It seems we are doing that or else EMPTY twice? maybe this could return Optional? Will the other one be invoked?

swalker125: hmm do you not think the two empties are both valid? One in this method is for if there is an id that doesn't match -> which can happen if hearings are deleted (although we don't deal with this currently at all) the other is if order is null. Maybe one empty catches both? I'll take a look

akomisarek: Sure makes sense, happy either way.

But at the same time was thinking client could deal with fact there is no preselected date, or something like that :)

PR #1166, opened 2020-05-05
fix: Reorder list of judges for allocation [FPLA-1720]

mounikahmcts:
[FPLA-1720](https://tools.hmcts.net/jira/browse/FPLA-1720)

List should be in this order:
Circuit Judge
Circuit Judge (Section 9)
District Judge
Magistrate
High Court Judge

### DIFF
--- a/ccd-definition/FixedLists/AllocationProposalList.json
+++ b/ccd-definition/FixedLists/AllocationProposalList.json
@@ -28,7 +28,6 @@
     "DisplayOrder": 4
   },
   {
-    
     "LiveFrom": "01/01/2017",
     "ID": "AllocationProposalList",
     "ListElementCode": "High Court judge",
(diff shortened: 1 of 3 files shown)


ccd-definition/FixedLists/AllocationProposalList.json
```diff
@@ -2,29 +2,29 @@
   {
     "LiveFrom": "01/01/2017",
     "ID": "AllocationProposalList",
-    "ListElementCode": "Lay justices",
-    "ListElement": "Magistrate",
+    "ListElementCode": "Circuit judge",
```
joez-f, line 5: I think each word in this list should start with a capital letter; our other list of judge titles has every first letter capitalised and I think we should do it here too; can check with a BA as it seems the existing allocation list does not do this (but looks like AC does specify capital letters)

PR #1174, opened 2020-05-06
feat: collect judge email addresses (FPLA-1664)

muyinatech:

https://tools.hmcts.net/jira/browse/FPLA-1664

### Change description ###
Add email address field to Allocated Judge screen

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
```diff
@@ -42,6 +42,7 @@ private static JudgeAndLegalAdvisor migrateJudgeAndLegalAdvisor(JudgeAndLegalAdv
             .otherTitle(allocatedJudge.getOtherTitle())
             .judgeLastName(allocatedJudge.getJudgeLastName())
             .judgeFullName(allocatedJudge.getJudgeFullName())
+            .judgeEmailAddress(allocatedJudge.getJudgeEmailAddress())
```
Jamesnelson117, line 45: When we migrate judge objects in create order, notice of proceedings etc the judge object associated with those events does not look like it captures judgeEmailAddress. It looks like this does not break anything (I see you've got a green build) but we might loose emailAddress in these cases. Is that a problem? Can you check wether this is the case?

muyinatech: I will check thanks.

muyinatech: As discussed, I will remove the migration of the judge's email address. As for create order, notice of proceedings etc, only a label for the allocated judge is retained. If there is a later requirement to use the judge's email address on another event other than Allocated Judge, it should be handle then.

muyinatech: Done.

ccd-definition/ComplexTypes/1_JudgeAndLegalAdvisor.json
```diff
@@ -70,5 +70,14 @@
...
     "FieldType": "Text",
     "SecurityClassification": "Public"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "JudgeAndLegalAdvisor",
+    "ListElementCode": "judgeEmailAddress",
+    "FieldType": "Email",
+    "ElementLabel": "Email Address",
+    "FieldShowCondition": "judgeTitle!=\"\"",
```
Jamesnelson117, line 80: Is this needed? Should this not be shown at all times?

muyinatech: When you first land on the screen for the 1st time none of the radio buttons are selected. The name & email fields are hidden. Upon selection these fields are shown.

PR #1184, opened 2020-05-11
feat: enhancement for allocated judge hearing details (FPLA-1598)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1598

### Change description ###
'Add hearing details' event now needs the allocated judge event to be complete before hearings can be added. 
Hearings can now use the case's allocated judge. 

Screenshot 2020-05-11 at 10 50 09

Screenshot 2020-05-11 at 10 14 48

```
[ ] Yes
[x] No
```


ccd-definition/CaseEventToComplexTypes/addHearingBookingDetails/addHearingBookingDetailsGatekeeping.json
```diff
@@ -152,8 +175,8 @@
     "CaseFieldID": "hearingDetails",
     "ListElementCode": "judgeAndLegalAdvisor.otherTitle",
     "EventElementLabel": "Title",
-    "FieldDisplayOrder": 10,
-    "FieldShowCondition": "hearingDetails.judgeAndLegalAdvisor.judgeTitle=\"OTHER\"",
+    "FieldDisplayOrder": 12,
+    "FieldShowCondition": "hearingDetails.judgeAndLegalAdvisor.useAllocatedJudge=\"No\" AND hearingDetails.judgeAndLegalAdvisor.judgeTitle=\"OTHER\"",
```
joez-f, line 179: This condition is essentially the same as the one already defined in the complextype, so I think it is redundant and can be removed. Same for condition on judgeTitle

joez-f: oh I see you've actually changed the complex type show condition, so yes this would be needed - but why change it?

joez-f:
Is the screenshot meant to show an allocated judge label?

ccd-definition/CaseEventToComplexTypes/addHearingBookingDetails/addHearingBookingDetailsGatekeeping.json
Jamesnelson117: Yep you're right, I had assumed the same thing. But I found when I did not include these additional fieldShowConditions there was inconsistencies with how it behaved. i.e 
- UI would work fine but full name etc would not be persisted in the tabs. 
- Full name input would re-appear if I selected 'no' to use allocatedJudge then selected 'yes'.

I'm not sure wether caseEventToComplexTypes behaves differently for collection items or wether this is a CCD bug but I found the workaround was to repeat show / hide logic.

Jamesnelson117: Complex type was changed as data was not persisting in add hearing details event. But this now causes...

- Full name input would re-appear if I selected 'no' to use allocatedJudge then selected 'yes'.

In the other events also :( investigating now.

Jamesnelson117: Resolved in latest commit.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -124,6 +125,8 @@
     private final Order standardDirectionOrder;
 
     @NotNull(message = "You need to enter the allocated judge.", groups = SealedSDOGroup.class)
+    @NotNull(message = "You need to enter a judge before you can add hearing details", groups =
```
Jamesnelson117, line 128: TODO - investigate wether the message can be overwritten from a group.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
```diff
@@ -74,10 +80,60 @@ public HearingBooking getHearingBookingByUUID(List<Element<HearingBooking>> hear
...
+    }
+
     private boolean isPastHearing(Element<HearingBooking> element) {
         return ofNullable(element.getValue())
             .map(HearingBooking::getStartDate)
             .filter(hearingDate -> hearingDate.isBefore(time.now()))
             .isPresent();
     }
+
+    private HearingBooking resetJudgeDetails(HearingBooking hearingBooking) {
```
joez-f, line 123: This process be simplified a bit; check my PR to see a method called prepareJudgeFields which will return a judge in the format we want; you could maybe use this

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Judge.java
```diff
@@ -34,4 +35,12 @@ public String getJudgeName() {
         }
         return judgeLastName;
     }
+
+    @JsonIgnore
+    public boolean equalsJudgeAndLegalAdvisor(JudgeAndLegalAdvisor judgeAndLegalAdvisor) {
```
joez-f, line 40: I've updated this in my PR so take a look (but as discussed, may need to re-think how to handle judge email addresses)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
Jamesnelson117: As discussed, would rather use my current implementation to avoid rebuilding each HearingBooking in the collection unnecessarily.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerAboutToStartTest.java
```diff
@@ -46,21 +70,69 @@ void shouldOnlyReturnHearingsWithFutureStartDateWhenHearingsInThePastExist() {
...
+        );
+
+        CaseDetails caseDetails = caseDetails(data);
+        AboutToStartOrSubmitCallbackResponse response = postAboutToStartEvent(caseDetails);
+        CaseData caseData = mapper.convertValue(response.getData(), CaseData.class);
+
+        HearingBooking updatedHearingBooking = caseData.getHearingDetails().get(0).getValue();
+        JudgeAndLegalAdvisor judgeAndLegalAdvisor = updatedHearingBooking.getJudgeAndLegalAdvisor();
+
+        assertThat(judgeAndLegalAdvisor.getJudgeTitle()).isEqualTo(null);
```
joez-f, line 106: can be .isNull()

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -123,7 +124,8 @@
     private final List<Element<Placement>> placements;
     private final Order standardDirectionOrder;
 
-    @NotNull(message = "You need to enter the allocated judge.", groups = SealedSDOGroup.class)
+    @NotNull(message = "You need to enter the allocated judge.",
```
joez-f, line 127: would prefer something more like "You need to allocate a judge to the case" but it's for content to decide

ccd-definition/CaseField.json
```diff
@@ -428,6 +428,14 @@
     "FieldType": "JudgeAndLegalAdvisor",
     "SecurityClassification": "Public"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "allocatedJudgeLabel",
```
joez-f, line 434: Instead of a new case field we should add a new field judgeLabel to AllocatedJudge complex type which is set in Allocated Judge event. Then we can use interpolation to access this field in CCD and show it within the collection. Current implementation would show it at the top of the page. Maybe should be separate a PR; but if we introduce this new case field and plan to remove it, will it cause some problems?

joez-f: I think this is important, we shouldn't make users scroll up to the top of the screen to view the judgeLabel if we have the ability to add the label in a better location

service/src/test/java/uk/gov/hmcts/reform/fpl/model/JudgeTest.java
```diff
@@ -43,6 +44,34 @@ void shouldReturnAllocatedJudgeLastNameWhenMagistratesIsNotSelected() {
         assertThat(lastName).isEqualTo("Judge Last Name");
     }
 
+    @Test
+    void shouldReturnTrueWhenAllocatedJudgeAndJudgeAndLegalAdvisorAreEqual() {
```
joez-f, line 48: shouldReturnTrueWhenJudgesAreEqual() ?

Bit easier to read, don't like the JudgeAndJudgeAndLegalAdvisor

service/src/test/java/uk/gov/hmcts/reform/fpl/model/JudgeTest.java
```diff
@@ -43,6 +44,34 @@ void shouldReturnAllocatedJudgeLastNameWhenMagistratesIsNotSelected() {
...
+            .judgeFullName("Judge Full Name")
+            .judgeLastName("Judge Last Name")
+            .build();
+
+        Judge allocatedJudge = buildAllocatedJudge(judgeOrMagistrateTitle);
+        assertThat(allocatedJudge.equalsJudgeAndLegalAdvisor(judgeAndLegalAdvisor)).isTrue();
+    }
+
+    @Test
+    void shouldReturnFalseWhenAllocatedJudgeAndJudgeAndLegalAdvisorAreNotEqual() {
```
joez-f, line 63: same as above

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
Jamesnelson117: Yep 👍 I've spoke to UX and they'd like a story created to update this wording. Out of scope for this story.

ccd-definition/CaseField.json
Jamesnelson117: This was the suggested approach for now.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/JudgeAndLegalAdvisor.java
```diff
@@ -35,4 +36,12 @@ public static JudgeAndLegalAdvisor from(final Judge allocatedJudge) {
         }
         return judgeAndLegalAdvisorBuilder.build();
     }
+
+    @JsonIgnore
+    public JudgeAndLegalAdvisor resetJudgeProperties(YesNo useAllocatedJudge) {
```
joez-f, line 41: don't think JsonIgnore is needed, also don't think there will be a situation where we want to pass 'No' into this, so parameter probably isn't needed either (we reset to clear the input fields, we wouldn't want to populate with 'No' and clear the fields because it would delete what the user inputted)

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/JudgeAndLegalAdvisor.java
```diff
@@ -35,4 +35,12 @@ public static JudgeAndLegalAdvisor from(final Judge allocatedJudge) {
         }
         return judgeAndLegalAdvisorBuilder.build();
     }
+
+    @JsonIgnore
+    public JudgeAndLegalAdvisor resetJudgeProperties() {
```
Gogetter, line 40: hmm, could we just have as `reset()`? As 1) its updating `JudgeAndLegalAdvisor` not `Judge` (different object) and 2) it's updating only two properties

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
```diff
@@ -74,10 +79,52 @@ public HearingBooking getHearingBookingByUUID(List<Element<HearingBooking>> hear
...
+    }
+
+    public List<Element<HearingBooking>> resetHearingJudge(List<Element<HearingBooking>> hearingBookings,
+                                                           Judge allocatedJudge) {
+        return hearingBookings.stream()
+            .map(element -> {
+                HearingBooking hearingBooking = element.getValue();
+                JudgeAndLegalAdvisor judgeAndLegalAdvisor = hearingBooking.getJudgeAndLegalAdvisor();
+
+                if (isNotEmpty(judgeAndLegalAdvisor)
```
Gogetter, line 105: minor: could we just have `judgeAndLegalAdvisor  != null`?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/JudgeAndLegalAdvisor.java
joez-f: I'll make this change too in my PR

PR #1185, opened 2020-05-11
fix: persist allocated judge selection in SDO (FPLA-1718)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-1718

### Change description ###

Persist the user's choice of 'Yes/No' regarding the allocated judge question in the draft SDO event. This is so users do not have to select yes/no every time they visit the event.
When draft SDO event is submitted, allocated judge details are removed so that they do not appear in the tab (this de-selects their choice). To persist the details, the recorded judge on the SDO and the allocated judge are compared in about-to-start; if they are equal, then populate with 'Yes', else 'No' - this reflects the user's choice

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
```diff
@@ -53,6 +53,23 @@ public static void removeAllocatedJudgeProperties(JudgeAndLegalAdvisor judgeAndL
...
+            return judgeAndLegalAdvisor;
+        }
+    }
+
+    private static JudgeAndLegalAdvisor resetJudgeFields(String legalAdvisorName) {
+        return JudgeAndLegalAdvisor.builder()
+            .useAllocatedJudge("Yes")
+            .legalAdvisorName(legalAdvisorName)
+            .build();
+    }
```
Jamesnelson117, line 71: Does this make sense more sense to be on the JudgeAndLegalAdvisor object?

service/src/test/java/uk/gov/hmcts/reform/fpl/model/JudgeTest.java
```diff
@@ -43,6 +44,35 @@ void shouldReturnAllocatedJudgeLastNameWhenMagistratesIsNotSelected() {
         assertThat(lastName).isEqualTo("Judge Last Name");
     }
 
+    @Test
+    void shouldReturnTrueWhenAllocatedJudgeAndParameterJudgeHaveEqualFields() {
```
Jamesnelson117, line 48: What do you think of` shouldReturnTrueWhenJudgeAndJudgeAndLegalAdvisorHaveEqualFields`?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToStartTest.java
```diff
@@ -132,6 +132,30 @@ void aboutToStartCallbackShouldSetAssignJudgeLabelWhenAllocatedJudgeIsPopulated(
...
+            .data(Map.of(
+                "allocatedJudge", judge,
+                "standardDirectionOrder", Order.builder()
+                    .directions(buildDirections(directions))
+                    .judgeAndLegalAdvisor(JudgeAndLegalAdvisor.builder()
+                        .judgeTitle(judge.getJudgeTitle())
+                        .judgeLastName(judge.getJudgeLastName())
+                        .build())
+                    .build()))
+            .build();
```
Jamesnelson117, line 150: Maybe extract to private?

service/src/test/java/uk/gov/hmcts/reform/fpl/model/JudgeTest.java
joez-f: shouldReturnTrueWhenJudgesHaveEqualFields? 👍

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToStartTest.java
joez-f: Sure

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
joez-f: I don't know :) What decides if it should be on the model or in a helper class?

Jamesnelson117: For now i've moved it to the JudgeAndLegalAdvisor object. It could equally live on both but I think it makes sense to live on the model.

joez-f: I moved it :)

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Judge.java
```diff
@@ -34,4 +38,11 @@ public String getJudgeName() {
         }
         return judgeLastName;
     }
+
+    public boolean hasEqualJudgeFields(JudgeAndLegalAdvisor judgeAndLegalAdvisor) {
+        return Objects.equals(getJudgeTitle(), judgeAndLegalAdvisor.getJudgeTitle())
+            && StringUtils.equals(getOtherTitle(), judgeAndLegalAdvisor.getOtherTitle())
+            && StringUtils.equals(getJudgeLastName(), judgeAndLegalAdvisor.getJudgeLastName())
+            && StringUtils.equals(getJudgeFullName(), judgeAndLegalAdvisor.getJudgeFullName());
```
Gogetter, line 46: minor: but we might want to prefer `getJudgeLastName().equals(judgeAndLegalAdvisor.getJudgeLastName())`

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/JudgeAndLegalAdvisor.java
```diff
@@ -35,4 +35,11 @@ public static JudgeAndLegalAdvisor from(final Judge allocatedJudge) {
         }
         return judgeAndLegalAdvisorBuilder.build();
     }
+
+    public JudgeAndLegalAdvisor resetJudgeProperties() {
```
Gogetter, line 39: why not static?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Judge.java
joez-f: If the field is null then this will fail (and it is possible/expected) :( Hence why I took the other approach

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/JudgeAndLegalAdvisor.java
joez-f: If static I cannot reference legalAdvisorName and would have to pass it as parameter instead. Which do you prefer? I think I will keep as non-static but i'm curious

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -95,20 +96,21 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
...
+            caseDetails.getData().put(JUDGE_AND_LEGAL_ADVISOR_KEY, prepareJudgePage(caseData));
         }
 
         return AboutToStartOrSubmitCallbackResponse.builder()
             .data(caseDetails.getData())
             .build();
     }
 
-    private JudgeAndLegalAdvisor setAllocatedJudgeLabel(CaseData caseData) {
+    private JudgeAndLegalAdvisor prepareJudgePage(CaseData caseData) {
```
Jamesnelson117, line 107: Why not just `prepareJudge`?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -95,20 +96,21 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
...
         JudgeAndLegalAdvisor judgeAndLegalAdvisor = JudgeAndLegalAdvisor.builder().build();
 
         if (isNotEmpty(caseData.getStandardDirectionOrder())
             && isNotEmpty(caseData.getStandardDirectionOrder().getJudgeAndLegalAdvisor())) {
```
Jamesnelson117, line 111: Do you think it's worthwhile creating a boolean private method from this? Or perhaps on the SDO model itself we could have `boolean hasJudge()`.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
```diff
@@ -53,6 +53,16 @@ public static void removeAllocatedJudgeProperties(JudgeAndLegalAdvisor judgeAndL
         judgeAndLegalAdvisor.setUseAllocatedJudge(null);
     }
 
+    public static JudgeAndLegalAdvisor prepareJudgeFields(JudgeAndLegalAdvisor judgeAndLegalAdvisor,
+                                                          Judge allocatedJudge) {
+        if (allocatedJudge.hasEqualJudgeFields(judgeAndLegalAdvisor)) {
+            return judgeAndLegalAdvisor.resetJudgeProperties();
+        } else {
+            judgeAndLegalAdvisor.setUseAllocatedJudge("No");
```
Jamesnelson117, line 61: Why do we explicitly need to set it to no? is ignoring it and leaving as null not enough?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
joez-f: Yeah that sounds better 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/JudgeAndLegalAdvisorHelper.java
joez-f: Because if the code reaches here then we know that there is a judge on the SDO, which means they must have selected something - if we left it as null we would lose the user's selection of 'No' so they would have to click it again

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
joez-f: Probably not worthwhile as this is the only time we are checking for a judge, but good to keep in mind

PR #1219, opened 2020-05-15
fix: add missing other title judge field to events (FPLA-1767)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-1767

### Change description ###

Added missing 'Title' field under 'Other' option for Judge or magistrate's title to:

- Notice of proceedings (also turned multi-state into single-state)
- Create an order (also turned multi-state into single-state)
- Draft SDO

```
[ ] Yes
[X] No
```


Jamesnelson117:
@joez-f is it worth updating a few e2e tests just to use 'other title'?

joez-f:
@Jamesnelson117 we use other title in hearing details, I don't see the benefit of having more e2e to test it elsewhere. it works there so I think it's fine

PR #1223, opened 2020-05-15
fix: restrict children marker to only final orders (FPLA-NA)

lewisbirks:

N/A

### Change description ###

It was noticed that the children final order issued markers were being updated from EPO and C21.

After chatting with the PO it has been determined that it shouldn't be set for C21 but can be set for EPO. As such the condition has been updated to reflect that

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerAboutToSubmitTest.java
```diff
@@ -308,6 +308,28 @@ void shouldNotSetFinalOrderIssuedForEmergencyProtectionOrder() {
             .containsOnlyNulls();
     }
 
+    @Test
+    void shouldNotUpdateChildrenWhenFeatureIsNotEnabled() {
+        given(featureToggleService.isCloseCaseEnabled()).willReturn(true);
```
muyinatech, line 313: should this be false?

lewisbirks: yeah realised... 🤦

lewisbirks: bad copy paste job

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
```diff
@@ -206,8 +205,9 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
 
         data.put("orderCollection", orders);
 
-        if (featureToggleService.isCloseCaseEnabled() && caseData.getOrderTypeAndDocument().getSubtype() != INTERIM) {
-            List<Element<Child>> updatedChildren = getUpdatedChildren(caseData);
+        if (featureToggleService.isCloseCaseEnabled() && caseData.getOrderTypeAndDocument().isClosable()) {
```
akomisarek, line 208: If we want to align with java world it could be: 
isCloseable!

akomisarek, line 208: I think this logic leaks a bit to the controller... I think it should be service responsibility to determine if the order should be closed, not controller! 

But I am not massively concerned... It's not like we aree going to have different entry point - but the controller test, testing logic are a bit fishy.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerAboutToSubmitTest.java
```diff
@@ -306,6 +311,69 @@ void shouldNotSetFinalOrderIssuedForInterimOrder() {
...
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToSubmitEvent(caseDetails);
+
+        final CaseData caseData = mapper.convertValue(callbackResponse.getData(), CaseData.class);
+
+        assertThat(caseData.getAllChildren()).extracting(element -> element.getValue().getFinalOrderIssued())
+            .containsOnly("Yes");
+    }
+
+    @Test
+    void shouldNotUpdateChildrenWhenFeatureIsNotEnabled() {
```
akomisarek, line 359: Those seems logic tests which all should reside in the service.

PR #1224, opened 2020-05-15
fix: moved logic to mid event and place directions in about to submit (FPLA-1782)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1782

### Change description ###

Combined logic in new mid event for date of issue validation and placing logic.
Moved the persisting of data to about to submit callback.

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
```diff
@@ -0,0 +1,213 @@
...
+import static uk.gov.hmcts.reform.fpl.utils.CaseDataGeneratorHelper.createHearingBooking;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.wrapElements;
+
+@ActiveProfiles("integration-test")
+@WebMvcTest(DraftOrdersController.class)
+@OverrideAutoConfiguration(enabled = true)
+class DraftOrdersControllerDateOfIssueMidEventTest extends AbstractControllerTest {
+
+    @Autowired
+    private Time time;
```
lewisbirks, line 46: Shouldn't need

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -71,28 +72,37 @@
...
+
+        caseDetails.getData().put("dateOfIssue", time.now().toLocalDate());
+
+        return AboutToStartOrSubmitCallbackResponse.builder()
+            .data(caseDetails.getData())
+            .build();
+    }
+
+    @PostMapping("/date-of-issue/mid-event")
+    public AboutToStartOrSubmitCallbackResponse handleMidEventValidateDateOfIssue(
```
lewisbirks, line 91: I would wrap the entire logic in this method on the condition that there were no errors, no point doing something that is going to have no effect and just be done again

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ValidateOrderController.java
```diff
@@ -33,19 +32,6 @@
     private final ObjectMapper mapper;
     private final ValidateGroupService validateGroupService;
 
-    @PostMapping("/date-of-issue/mid-event")
-    public AboutToStartOrSubmitCallbackResponse handleMidEventValidateDateOfIssue(
```
lewisbirks, line 37: Do we still not use this for other orders, such as the orders in the create an order event?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ValidateOrderControllerTest.java
```diff
@@ -36,27 +36,6 @@
         super("validate-order");
     }
 
-    @Test
-    void shouldReturnErrorsWhenTheDateOfIssueIsInFuture() {
```
lewisbirks, line 40: see previous

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ValidateOrderController.java
swalker125: you're right we do, this could be a real messy problem. The mid event at the beginning of the event bulldozes anything you set in the about to start

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
```diff
@@ -0,0 +1,208 @@
...
+    DraftOrdersControllerDateOfIssueMidEventTest() {
+        super("draft-standard-directions");
+    }
+
+    @Nested
+    class DateOfIssueValidation {
+
+        @Test
+        void shouldReturnErrorsWhenTheDateOfIssueIsInFuture() {
+            CaseDetails caseDetails = caseDetails(now().plusDays(1));
```
lewisbirks, line 52: `dateNow()` provides a LocalDate

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -71,28 +72,46 @@
...
```
lewisbirks, line 91: Also this mid event deals with more than just validating the date of issue, maybe there would be a better name for it

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
```diff
@@ -0,0 +1,208 @@
...
+        LocalDateTime hearingDate = LocalDateTime.of(2020, 1, 1, 0, 0, 0);
+
+        CaseDetails caseDetails = CaseDetails.builder()
+            .data(Map.of("hearingDetails", wrapElements(createHearingBooking(hearingDate, hearingDate.plusDays(1)))))
+            .build();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postMidEvent(caseDetails, "date-of-issue");
+
+        Stream.of(DirectionAssignee.values()).forEach(assignee ->
+            AssertionsForClassTypes.assertThat(callbackResponse.getData().get(assignee.toHearingDateField()))
```
lewisbirks, line 87: `AssertionsForClassTypes` we don't need to use that type of assertion, it is for java 8 when it couldn't resolve the proper `assertThat` to call. [See here](https://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/AssertionsForClassTypes.html)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
```diff
@@ -0,0 +1,208 @@
...
+                .isEqualTo("1 January 2020, 12:00am"));
+    }
+
+    @Test
+    void shouldShowEmptyPlaceHolderForHearingDateWhenNoHearingDate() {
+        CaseDetails caseDetails = CaseDetails.builder().data(new HashMap<>()).build();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postMidEvent(caseDetails, "date-of-issue");
+
+        Stream.of(DirectionAssignee.values()).forEach(assignee ->
```
lewisbirks, line 97: Seeing as this stream is used more than one, thoughts of precomputing the resultant array of values at the start of the class creation.
```java
    private static final Object[] DIRECTION_ASSIGNEE_KEYS = Stream.of(DirectionAssignee.values())
        .map(DirectionAssignee::toHearingDateField)
        .toArray();
```
and then the assertion can be
```java
assertThat(callbackResponse.getData()).extracting(DIRECTION_ASSIGNEE_KEYS).containsOnly("Please enter a hearing date");
```
same can then be used for the test above

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -252,7 +254,33 @@ public void handleSubmittedEvent(
         }
     }
 
-    private List<Element<Direction>> getConfigDirectionsWithHiddenValues() throws IOException {
+    private JudgeAndLegalAdvisor prepareJudge(CaseData caseData) {
```
lewisbirks, line 257: Do we want these methods in the controller, would it not be better for them to be in a service?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
swalker125: yeah weird I thought I had removed all of this, good spot

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
swalker125: logic was already there I've just moved it. As this is a bug fix ticket I am not going to move

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToSubmitTest.java
```diff
@@ -78,35 +81,25 @@
     void setup() {
         DocmosisDocument docmosisDocument = new DocmosisDocument(SEALED_ORDER_FILE_NAME, PDF);
 
-        given(documentGeneratorService.generateDocmosisDocument(any(DocmosisData.class), any()))
-            .willReturn(docmosisDocument);
+        given(docmosisService.generateDocmosisDocument(any(DocmosisData.class), any())).willReturn(docmosisDocument);
```
toireasam, line 84: could be more specific here and do eq(SDO) instead of any() for the template

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
```diff
@@ -0,0 +1,208 @@
...
+                    .directions(buildDirections(createDirections()))
+                    .judgeAndLegalAdvisor(JudgeAndLegalAdvisor.builder()
+                        .judgeTitle(judge.getJudgeTitle())
+                        .judgeLastName(judge.getJudgeLastName())
+                        .build())
+                    .build()))
+            .build();
+    }
+
+    private Judge buildAllocatedJudge() {
```
toireasam, line 183: we have a testJudge method in testDataHelper we could probably use now instead of this - don't mind if you want to leave as is as would require changing the asserts for some of the tests

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
swalker125: these tests were already this way, as this is just fixing a bug I will not change

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
swalker125: I'll remove the word validate from the method, it fires after date of issue so makes sense to me

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
swalker125: meh

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToSubmitTest.java
swalker125: as this is a bug fix I am not going to change this

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerDateOfIssueMidEventTest.java
swalker125: I will update to use that, thanks

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftCMOController.java
```diff
@@ -51,9 +52,9 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
             caseDetails.getData().putAll(caseManagementOrder.getCCDFields());
         }
 
-        caseDetails.getData().put(HEARING_DATE_LIST.getKey(),
-            caseManagementOrderService.getHearingDateDynamicList(caseData, caseManagementOrder));
+        DynamicList hearingList = caseManagementOrderService.getHearingDateDynamicList(caseData, caseManagementOrder);
 
+        caseDetails.getData().put(HEARING_DATE_LIST.getKey(), hearingList);
```
lewisbirks, line 57: unnecessary change for the bug fix 👀

PR #1232, opened 2020-05-18
feat: adding allocated judge email (FPLA-1746)

toireasam:

### Change description ###

```
[ ] Yes
[ ] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotifyAllocatedJudgeEventHandler.java
```diff
@@ -0,0 +1,38 @@
...
+    private final NotificationService notificationService;
+    private final AllocatedJudgeContentProvider allocatedJudgeContentProvider;
+    private final ObjectMapper mapper;
+
+    @EventListener
+    public void notifyAllocatedJudge(NotifyAllocatedJudgeEvent event) {
+        EventData eventData = new EventData(event);
+        CaseData caseData = mapper.convertValue(eventData.getCaseDetails().getData(), CaseData.class);
+
+        Map<String, Object> parameters = allocatedJudgeContentProvider
```
lewisbirks, line 30: Could we use custom objects for this instead of a Map, it will be easier to maintain and review

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationEventHandlerTestData.java
```diff
@@ -57,6 +58,15 @@ public static CallbackRequest appendSendToCtscOnCallback() {
             .build();
     }
 
+    public static Map<String, Object> getExpectedAllocatedJudgeNotificationParameters() {
```
lewisbirks, line 61: I would keep this in the `NotifyAllocatedJudgeEventHandlerTest` unless you think that other tests are going to be able to make use of it

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NotifyAllocatedJudgeEventHandler.java
toireasam: cool will change :)

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NotificationEventHandlerTestData.java
toireasam: will do :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AllocatedJudgeContentProvider.java
```diff
@@ -0,0 +1,29 @@
...
+import uk.gov.hmcts.reform.fpl.model.notify.AllocatedJudgeTemplate;
+import uk.gov.hmcts.reform.fpl.service.email.content.base.AbstractEmailContentProvider;
+
+@Service
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class AllocatedJudgeContentProvider extends AbstractEmailContentProvider {
+
+    private final ObjectMapper mapper;
+
+    public AllocatedJudgeTemplate buildAllocatedJudgeNotificationParameters(CaseDetails caseDetails) {
```
lewisbirks, line 18: I feel like this name is overly verbose, from the class name we know that we are going to be using functions that are related to allocated judge so I would probs just call this something like buildNotificationParameters, I feel like this provides all the info necessary without being too wordy

lewisbirks:
Code smell seems to be the public modifier on
```java
public class NotifyAllocatedJudgeEventHandlerTest
```
it can probably just be package private (no modifier) if you want to resolve the code smell.
Also the controller isn't tested, I guess there is very little to actually test in there so I don't know what would actually be a good test there. I guess we could verify that upon calling the method that a notification is sent but that is all I can think of.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/notify/AllocatedJudgeTemplate.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.reform.fpl.model.notify;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
```
Jamesnelson117, line 7: Don't think setter annotation is needed since fields are not final.

lewisbirks: Setter is needed, if the fields were final then the setter wouldn't be needed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/AllocatedJudgeContentProvider.java
toireasam: cool makes sense :)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AllocatedJudgeControllerSubmittedTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.verify;
+import static uk.gov.hmcts.reform.fpl.NotifyTemplates.ALLOCATED_JUDGE_TEMPLATE;
+import static uk.gov.hmcts.reform.fpl.utils.CoreCaseDataStoreLoader.callbackRequest;
+
+@ActiveProfiles("integration-test")
+@WebMvcTest(AllocatedJudgeController.class)
+@OverrideAutoConfiguration(enabled = true)
+public class AllocatedJudgeControllerSubmittedTest extends AbstractControllerTest {
+    private static final String allocatedJudgeEmail = "test@test.com";
```
lewisbirks, line 20: constants should be uppercase

toireasam: fixed :) checkstyle complained initially. weird.

PR #1236, opened 2020-05-19
fix: date persistance when marking sdo direction not needed (FPLA-1790)

swalker125:

N/A

### Change description ###

Moved date population logic to new StandardDirectionsService. Dates now populated on about to submit callback of draft sdo event. This refactor and new service should make it easier to populate standard directions with dates on other events.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/JsonOrdersLookupService.java
```diff
@@ -1,26 +1,86 @@
...
+    }
+
+    private LocalDate addDelta(LocalDateTime date, int delta) {
+        if (delta == 0) {
+            return date.toLocalDate();
+        }
+        return calendarService.getWorkingDayFrom(date.toLocalDate(), delta);
+    }
+
+    private String booleanToYesOrNo(boolean value) {
```
akomisarek, line 83: I am not sure if the above method belongs here, 

this class and interface were just responsible for providing direction config for the file, by adding above the code has got multi responsibilities. What if there is a RESTCALL to provide direction library? 
Now we do lookup + some transformation which probably is not the same class.

swalker125: I have moved to a new service, I fail to see the point in making this into a restcall especially considering this is a bug fix

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToSubmitTest.java
```diff
@@ -11,15 +11,7 @@
...
-import uk.gov.hmcts.reform.fpl.model.Applicant;
-import uk.gov.hmcts.reform.fpl.model.ApplicantParty;
-import uk.gov.hmcts.reform.fpl.model.CaseData;
-import uk.gov.hmcts.reform.fpl.model.Direction;
-import uk.gov.hmcts.reform.fpl.model.HearingBooking;
-import uk.gov.hmcts.reform.fpl.model.Judge;
-import uk.gov.hmcts.reform.fpl.model.Order;
-import uk.gov.hmcts.reform.fpl.model.Respondent;
-import uk.gov.hmcts.reform.fpl.model.RespondentParty;
+import uk.gov.hmcts.reform.fpl.model.*;
```
pd-dudkiewicz, line 14: I'd rather leave it as it was, even if it was longer, we don't need all models here.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToSubmitTest.java
```diff
@@ -39,20 +31,13 @@
...
+import static uk.gov.hmcts.reform.fpl.enums.DirectionAssignee.*;
 import static uk.gov.hmcts.reform.fpl.enums.JudgeOrMagistrateTitle.MAGISTRATES;
 import static uk.gov.hmcts.reform.fpl.enums.OrderStatus.SEALED;
 import static uk.gov.hmcts.reform.fpl.service.HearingBookingService.HEARING_DETAILS_KEY;
 import static uk.gov.hmcts.reform.fpl.utils.DateFormatterHelper.formatLocalDateToString;
 import static uk.gov.hmcts.reform.fpl.utils.DocumentManagementStoreLoader.document;
-import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
-import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.unwrapElements;
-import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.wrapElements;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.*;
```
pd-dudkiewicz, line 40: As above, if we are not using all methods then there's no value in that change. Will not point out similar changes, happy to discuss that.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -140,13 +127,15 @@ public AboutToStartOrSubmitCallbackResponse handleMidEvent(
                 .build())
             .build();
 
-        prepareDirectionsForDataStoreService.persistHiddenDirectionValues(
-            getConfigDirections(), updated.getStandardDirectionOrder().getDirections());
+        Order standardDirectionOrder = updated.getStandardDirectionOrder();
+
+        //add hidden values to directions
+        persistValues(getFirstHearing(caseData.getHearingDetails()), standardDirectionOrder.getDirections());
```
pd-dudkiewicz, line 133: I'd rather call it `persistHiddenValues` to avoid comment.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/PopulateStandardDirectionsHandlerTest.java
```diff
@@ -20,295 +17,174 @@
...
 
     @Test
-    void shouldPopulateStandardDirectionsWhenTextContainsSpecialCharacters() throws IOException {
-        callbackRequest = getCallbackRequestWithCustomHearingOnMonday();
-
-        given(startPopulateStandardDirectionsEvent()).willReturn(getStartEventResponse(callbackRequest));
-
-        given(ordersLookupService.getStandardDirectionOrder())
-            .willReturn(getOrderDefinition(DIRECTION_TEXT_SPECIAL_CHARACTERS, SAME_DAY_AS_HEARING));
+    void shouldAddDirectionsToCaseData() throws IOException {
```
pd-dudkiewicz, line 113: I think it's too detailed and it's not even unit test anymore, because we're not isolating it. We should mock dependencies instead of autowiring them. The only responsibilities of PopulateStandardDirectionsHandler that should be tested are:
- hearingBookingService call
- standardDirectionsService call with data returned from the above (we don't really care if the data returned here is valid, this should be verified in standardDirectionsService unit test)
- commonDirectionService call with data returned from standardDirectionsService
- case data is populated with data returned from commonDirectionService (again, this doesn't need to be valid data from end-to-end perspective)
- ccd event is triggered

Thoughts? I'm happy to merge it like this, because it was already too complex before. I was doing these changes on my branch anyway, so I'm fine with leaving it as it is in your PR.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToSubmitTest.java
swalker125: ah sorry that's fresh intellij settings 🤦

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/PopulateStandardDirectionsHandlerTest.java
swalker125: sure, so I guess mocking out hearing service and standard directions service call will solve your comment?

pd-dudkiewicz: I'd mock commonDirectionService as well, just to have it as isolated as much as possible.

PR #1240, opened 2020-05-19
feat: share a case with local authority users in background thread (FPLA-1758)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-1758

related to: https://github.com/hmcts/fpl-ccd-configuration/pull/1250

### DIFF
--- a/charts/fpl-case-service/values.preview.template.yaml
+++ b/charts/fpl-case-service/values.preview.template.yaml
@@ -36,6 +36,7 @@ java:
         - robotics-notification-recipient
         - ld-sdk-key
         - ctsc-inbox
+        - AppInsightsInstrumentationKey
 
 idam-pr:
   releaseNameOverride: ${SERVICE_NAME}-xui-idam-pr
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
@@ -1,13 +1,13 @@
 package uk.gov.hmcts.reform.fpl.config;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.scheduling.annotation.AsyncConfigurer;
 
 import java.lang.reflect.Method;
 
+@Slf4j
 @Configuration
 public class AsyncConfiguration implements AsyncConfigurer {
 
@@ -18,11 +18,9 @@ public class AsyncConfiguration implements AsyncConfigurer {
 
     public class AsyncExceptionHandler implements AsyncUncaughtExceptionHandler {
 
-        private final Logger logger = LoggerFactory.getLogger(getClass());
-
         @Override
         public void handleUncaughtException(Throwable throwable, Method method, Object... obj) {
-            logger.error("Unexpected error occurred during async execution", throwable);
+            log.error("Unexpected error occurred during async execution", throwable);
         }
     }
 
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/exceptions/GrantCaseAccessException.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/exceptions/GrantCaseAccessException.java
@@ -1,0 +1,29 @@
+package uk.gov.hmcts.reform.fpl.exceptions;
+
+import lombok.EqualsAndHashCode;
+import uk.gov.hmcts.reform.fpl.enums.CaseRole;
+
+import java.util.Set;
+
+@EqualsAndHashCode(callSuper = false)
+public class GrantCaseAccessException extends RuntimeException {
+
+    private String caseId;
+    private String localAuthority;
+    private Set<String> userIds;
+    private Set<CaseRole> caseRoles;
+
+    public GrantCaseAccessException(String caseId, Set<String> userIds, Set<CaseRole> caseRoles) {
+        super(String.format("User(s) %s not granted %s to case %s", userIds, caseRoles, caseId));
+        this.caseId = caseId;
+        this.userIds = userIds;
+        this.caseRoles = caseRoles;
+    }
+
+    public GrantCaseAccessException(String caseId, String localAuthority, Set<CaseRole> caseRoles, Exception e) {
+        super(String.format("User from %s not granted %s to case %s", localAuthority, caseRoles, caseId), e);
+        this.caseId = caseId;
+        this.localAuthority = localAuthority;
+        this.caseRoles = caseRoles;
+    }
+}
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseRoleService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseRoleService.java
@@ -1,0 +1,86 @@
+package uk.gov.hmcts.reform.fpl.service;
+
+import com.google.common.collect.Sets;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
+import uk.gov.hmcts.reform.ccd.client.CaseUserApi;
+import uk.gov.hmcts.reform.ccd.client.model.CaseUser;
+import uk.gov.hmcts.reform.fpl.config.SystemUpdateUserConfiguration;
+import uk.gov.hmcts.reform.fpl.enums.CaseRole;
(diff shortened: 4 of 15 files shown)


charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -36,6 +36,7 @@ java:
         - robotics-notification-recipient
         - ld-sdk-key
         - ctsc-inbox
+        - AppInsightsInstrumentationKey
```
tomaszpowroznik, line 39: temporary: connect preview to aat azure so alerts can be tested

service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
```diff
@@ -18,11 +18,9 @@ public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
 
     public class AsyncExceptionHandler implements AsyncUncaughtExceptionHandler {
 
-        private final Logger logger = LoggerFactory.getLogger(getClass());
-
         @Override
         public void handleUncaughtException(Throwable throwable, Method method, Object... obj) {
-            logger.error("Unexpected error occurred during async execution", throwable);
+            log.error("Unexpected error occurred during async execution", throwable);
```
Gogetter, line 23: any need using the `com.microsoft.applicationinsights` library to also log the errors to appinsights?

tomaszpowroznik: No, it does not work this way. App insight is automatically integrated with standard loggers so all logs lands in azure
https://github.com/Microsoft/ApplicationInsights-Java/tree/v2.3.1/azure-application-insights-spring-boot-starter

service/src/main/java/uk/gov/hmcts/reform/fpl/exceptions/GrantCaseAccessException.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.Set;
+
+@EqualsAndHashCode(callSuper = false)
+public class GrantCaseAccessException extends RuntimeException {
+
+    private String caseId;
+    private String localAuthority;
+    private Set<String> userIds;
+    private Set<CaseRole> caseRoles;
```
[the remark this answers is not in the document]
tomaszpowroznik: I left in on purpose not final, as there are 2 constructors and non of them have all properties available. Of course I can initialize missing property with nulls but it is more code without any significant benefit

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseRoleServiceTest.java
```diff
@@ -0,0 +1,197 @@
...
+            assertThat(actualException).isEqualTo(expectedException);
+        }
+    }
+
+    private void verifyAccessGrantAttempted(String userId) {
+        verify(caseUser).updateCaseRolesForUser(AUTH_TOKEN, SERVICE_AUTH_TOKEN, CASE_ID, userId, caseUser(userId));
+    }
+
+    private void verifyAccessGrantNotAttempted(String userId) {
+        verify(caseUser, never()).updateCaseRolesForUser(any(), any(), any(), eq(userId), any());
```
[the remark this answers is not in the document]
tomaszpowroznik: I see method not only as way of reusing code but a way of adding logical name to what code does. From high perspective i guess it is easier to understand what test does with:

            verifyAccessGrantNotAttempted(USER_2_ID);
            verifyAccessGrantAttempted(USER_1_ID);
            verifyAccessGrantAttempted(USER_3_ID);

rather than :

            verify(caseUser, never()).updateCaseRolesForUser(any(), any(), any(), eq(userId), any());
            verifyAccessGrantAttempted(USER_1_ID);
            verifyAccessGrantAttempted(USER_3_ID);

and looks consistent and nicer :)

service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
Gogetter: thanks for the link @tomaszpowroznik.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseRoleService.java
```diff
@@ -0,0 +1,87 @@
...
+        Set<String> localAuthorityUsers = getUsers(caseId, localAuthority, excludeUsers, roles);
+        grantCaseAccess(caseId, localAuthorityUsers, roles);
+    }
+
+    private void grantCaseAccess(String caseId, Set<String> users, Set<CaseRole> roles) {
+        final Set<String> usersGrantedAccess = new HashSet<>();
+
+        try {
+            final String userToken = idam.authenticateUser(userConfig.getUserName(), userConfig.getPassword());
+            final String serviceToken = authTokenGenerator.generate();
```
akomisarek, line 49: I guess minor, but we are logging twice now - for single user and then for multiple sharing. Would you believe there is benefit of doing this once? Maybe caching it quickly for couple minutes?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseRoleService.java
```diff
@@ -0,0 +1,87 @@
...
+            final Set<String> caseRoles = roles.stream()
+                .map(CaseRole::formattedName)
+                .collect(toSet());
+
+            users.stream().parallel()
+                .forEach(userId -> {
+                    try {
+                        caseUser.updateCaseRolesForUser(
+                            userToken, serviceToken, caseId, userId, new CaseUser(userId, caseRoles));
+                        usersGrantedAccess.add(userId);
```
akomisarek, line 59: Is hashSet add thread-safe operation? I don't think it is as the hashset is not generally thread-safe and you have got parallel here.

tomaszpowroznik: you are right, fixed

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseRoleServiceTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+            verifyAccessGrantNotAttempted(USER_2_ID);
+            verifyAccessGrantAttempted(USER_1_ID);
+            verifyAccessGrantAttempted(USER_3_ID);
+        }
+
+        @Test
+        void shouldGrantAccessToRemainingUsersAndThrowExceptionWhenFailedForSomeUsers() {
+            Set<String> localAuthorityUsers = Set.of(USER_1_ID, USER_2_ID, USER_3_ID);
+            Set<String> excludedUser = emptySet();
```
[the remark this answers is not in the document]
tomaszpowroznik: you are absolutely right, intention was to verify that grant is attempted regardless of single exception and this test does not prove it
fixed

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/assertions/AnnotationAssertion.java
```diff
@@ -0,0 +1,41 @@
...
+
+        var expectedAsyncMethod = Set.of(expectedMethods);
+
+        if (!Objects.equals(actualAsyncMethods, expectedAsyncMethod)) {
+            failWithMessage("Expecting \n\t%s\n to have @Async methods:\n\t %s\n but the following were found: \n\t %s",
+                actual, expectedAsyncMethod, actualAsyncMethods);
+        }
+
+        return this;
+    }
```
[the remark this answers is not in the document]
tomaszpowroznik: designed this way, if there is additional async method which is not passed as param it fail.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseRoleService.java
tomaszpowroznik: i think these are separated method, which here are called from totally separated threads, i think from logical perspective it is fine to keep them like this. Dont want to pass tokens as method params. 
Cacheing can be considered at client level i think, service like this should not bother about cacheing, it is not its responsibility

service/src/main/java/uk/gov/hmcts/reform/fpl/exceptions/GrantCaseAccessException.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public GrantCaseAccessException(String caseId, Set<String> userIds, Set<CaseRole> caseRoles) {
```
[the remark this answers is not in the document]
tomaszpowroznik: I assumed that it is not needed. Second constructor have it as list of users is not know. 
From method perspective:

public void grantAccessToUser(String caseId, String user, Set<CaseRole> roles)

it is not related to any la, can be treated as generic method to give arbitrary roles to any user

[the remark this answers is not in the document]
akomisarek: I see, this scenario is interesting as here we always know LA, was thinking maybe it would be good for support perspective, but let's leave it for now.

PR #1249, opened 2020-05-21
impr: hide deleted cases from case list and filter (FPLA-1798)

Gogetter:
[FPLA-1798](https://tools.hmcts.net/jira/browse/FPLA-1798)

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/ccd-definition/State.json
+++ b/ccd-definition/State.json
@@ -43,7 +43,7 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "ID": "Deleted",
-    "Name": " ",
+    "Name": "Deleted",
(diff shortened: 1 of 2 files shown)


ccd-definition/State.json
```diff
@@ -43,7 +43,7 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "ID": "Deleted",
-    "Name": "Deleted",
+    "Name": " ",
```
tomaszpowroznik, line 46: Is it need?

Gogetter: yes, as one of the AC's said 

> Deleted should not be a filter option

PR #1251, opened 2020-05-21
impr: upgrade spring version to 2.3.0

tomaszpowroznik:


service/src/main/resources/application.yaml
```diff
@@ -8,7 +8,9 @@ management:
   endpoint:
     health:
       show-details: "always"
-
+  health:
+    probes:
+      enabled: false
```
tomaszpowroznik, line 13: new 2.3.0 feature that collides with 
uk.gov.hmcts.reform.health-spring-boot-starter: '0.0.5'

service/src/main/java/uk/gov/hmcts/reform/fpl/config/LocalAuthorityCodeLookupConfiguration.java
```diff
@@ -19,7 +20,7 @@ public LocalAuthorityCodeLookupConfiguration(@Value("${fpl.local_authority_email
     }
 
     public String getLocalAuthorityCode(String emailDomain) {
-        checkNotNull(emailDomain, "Email domain cannot be null");
```
tomaszpowroznik, line 22: deprecated in assertj version required by spring 2.3.0, anyway it was from testing support used in prod code

lewisbirks: Maybe it was intended to use the [guava preconditions](https://github.com/google/guava/wiki/PreconditionsExplained), it has the same method name

service/src/main/resources/application.yaml
lewisbirks: There has been [some discussion](https://hmcts-reform.slack.com/archives/C8SR5CAMU/p1589978499323600) around this already, don't know if it is helpful

service/src/main/java/uk/gov/hmcts/reform/fpl/config/LocalAuthorityCodeLookupConfiguration.java
tomaszpowroznik: could be, anyway prefer use native java counterpart

service/src/main/resources/application.yaml
tomaszpowroznik: good to know, was not aware about this discussion.

PR #1254, opened 2020-05-22
Upgrade spring version to 2.3.0

tomaszpowroznik:


service/build.gradle
```diff
@@ -202,12 +202,12 @@ dependencies {
   compile group: 'org.flywaydb', name: 'flyway-core', version: '6.3.2'
 
   compile group: 'org.springframework.boot', name: 'spring-boot-starter-actuator', version: versions.springBoot
+  compile group: 'org.springframework.boot', name: 'spring-boot-starter-validation', version: versions.springBoot
```
tomaszpowroznik, line 205: now needed as previously it was transitive dependency of health-spring-boot-starter

PR #1256, opened 2020-05-22
chores: upgrade gradle from '5.6.2' to '6.4.1'

Gogetter:
### Change description ###

Upgrade Gradle version from `5.6.2` to `6.4.1`

```
[ ] Yes
[x] No
```


lewisbirks:
Do we want to replace the deprecated functionality as well such as `compile` with `implementation` etc https://tools.hmcts.net/jira/browse/FPLA-1657

tomaszpowroznik:
> Do we want to replace the deprecated functionality as well such as `compile` with `implementation` etc https://tools.hmcts.net/jira/browse/FPLA-1657

I think you are right,  that should be cleaned up, but we can keep it for FPLA-1657

PR #1257, opened 2020-05-22
feat: close case from final order (FPLA-1605)

lewisbirks:

[FPLA-1605](https://tools.hmcts.net/jira/browse/FPLA-1605)

### Change description ###

Added the option to close the case from a final order if all children have a final order issued

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderService.java
```diff
@@ -368,4 +371,33 @@ private String getFormattedRemovalAddress(CaseData caseData) {
         return Optional.ofNullable(caseData.getEpoRemovalAddress())
             .map(address -> address.getAddressAsString(", ")).orElse("");
     }
+
+    public boolean shouldGenerateDocument(OrderTypeAndDocument orderTypeAndDocument,
+                                          FurtherDirections orderFurtherDirections,
+                                          List<Element<Child>> children,
+                                          String closeCaseFromOrder) {
+        // generate order if one of the following is met:
```
joez-f, line 379: Think I would prefer /** * * */ annotation for blocks of comments, and maybe above the method name rather than inside the method (this seems to be the approach we have in the code i.e CommonDirectionService

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerMidEventTest.java
```diff
@@ -314,5 +364,15 @@ private DocumentReference expectedDraftDocument() {
                 .url(document.links.self.href)
                 .build();
         }
+
+        private Element<Child> childWithFinalOrderIssued(String finalOrderIssued) {
```
joez-f, line 368: Can maybe put this in TestDataHelper (next to testChild()) as we use it in both controller + service tests and it may be used more in the future

lewisbirks: Are we remaking CaseDataGeneratorHelper 😆

service/src/main/java/uk/gov/hmcts/reform/fpl/service/GeneratedOrderService.java
lewisbirks: Sure thing

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerAboutToSubmitTest.java
```diff
@@ -128,6 +132,47 @@ void aboutToSubmitShouldNotHaveDraftAppendedToFilename() {
         assertThat(caseData.getOrderCollection().get(0).getValue().getDocument()).isEqualTo(expectedDocument());
     }
 
+    @Test
+    void shouldAddCloseCaseObjectToCaseDataWhenOptionWasSelectedAndChangeState() {
+        JudgeAndLegalAdvisor judgeAndLegalAdvisor = buildJudgeAndLegalAdvisor(NO);
+
+        final CaseDetails caseDetails = buildCaseDetails(
+            commonCaseDetailsComponents(CARE_ORDER, FINAL, judgeAndLegalAdvisor)
+                .caseLocalAuthority(LOCAL_AUTHORITY_CODE)
```
joez-f, line 141: just a heads up from my PR, I put the .caseLocalAuthority inside commonCaseDetailsComponents because it is a common casedetails component, so if mine gets merged before yours make sure you handle that!

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerMidEventTest.java
Jamesnelson117: Least it's no longer in main ;)

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerAboutToSubmitTest.java
```diff
@@ -128,6 +132,47 @@ void aboutToSubmitShouldNotHaveDraftAppendedToFilename() {
...
+    }
+
+    @Test
+    void shouldAddCloseCaseObjectToCaseDataWhenOptionWasNotSelected() {
+        JudgeAndLegalAdvisor judgeAndLegalAdvisor = buildJudgeAndLegalAdvisor(NO);
+
+        final CaseDetails caseDetails = buildCaseDetails(
+            commonCaseDetailsComponents(CARE_ORDER, FINAL, judgeAndLegalAdvisor)
+                .caseLocalAuthority(LOCAL_AUTHORITY_CODE)
+                .children1(createChildren("Fred", "John"))
```
joez-f, line 167: see my PR #1260 for something I added to TestDataHelper: testChildren() - I'd prefer using that over having a separate private method in each test class which does the same thing (since you're already editing these controller tests maybe can swap createChildren() out for testChildren())

lewisbirks: Will have a look

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerMidEventTest.java
joez-f: New and improved! (I think for something simple like this it is ideal, but yes let's try not to recreate that!)

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
```diff
@@ -131,32 +135,57 @@ public AboutToStartOrSubmitCallbackResponse handlePopulateSelectorMidEvent(
 
     @PostMapping("/generate-document/mid-event")
     public AboutToStartOrSubmitCallbackResponse handleMidEvent(@RequestBody CallbackRequest callbackRequest) {
-        CaseDetails caseDetails = callbackRequest.getCaseDetails();
-        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+        // This mid event is called after:
```
joez-f, line 138: Probably should java doc this one too

lewisbirks: Don't think it really fits as java doc but can happily make it a multiline comment

PR #1260, opened 2020-05-22
fix: default orderAppliesToAllChildren to yes when only one child (FPLA-1792)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-1792

### Change description ###

Set orderAppliesToAllChildren to be 'Yes' by default when there is only one child on the case. This flag is usually set by the user on a page that only shows when there are multiple children. If this flag is not set, then the final order issued marker will also not be set, resulting in a bug

```
[ ] Yes
[X] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerAboutToStartTest.java
```diff
@@ -57,6 +61,34 @@ void shouldSetDateOfIssueAsTodayByDefault() {
...
+    @Test
+    void shouldSetOrderAppliesFlagToYesWhenOnlyOneChildOnCase() {
+        CaseDetails caseDetails = CaseDetails.builder()
+            .id(parseLong(CASE_ID))
+            .data(Map.of(FAMILY_MAN_CASE_NUMBER_KEY, FAMILY_MAN_CASE_NUMBER_VALUE, CHILDREN_KEY, List.of(testChild())))
+            .build();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails);
+
+        assertThat(callbackResponse.getErrors()).isEmpty();
```
lewisbirks, line 73: Don't think we need to assert this in both of the new tests in here

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
```diff
@@ -90,7 +90,13 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody Call
             ValidateFamilyManCaseNumberGroup.class);
 
         if (errors.isEmpty()) {
-            childrenService.addPageShowToCaseDetails(caseDetails, caseData.getAllChildren());
+            int numOfChildren = caseData.getAllChildren().size();
+            caseDetails.getData().put("pageShow", numOfChildren <= 1 ? "No" : "Yes");
+
+            if (numOfChildren <= 1) {
+                caseDetails.getData().put("orderAppliesToAllChildren", "Yes");
```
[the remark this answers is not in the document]
lewisbirks: It would have to be `YesNo.Yes.getValue()`, unless something has changed recently. I think I remember changing the JsonProperty of the enum it broke something else. Although this was a couple of months ago so I could be misremembering.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
```diff
@@ -178,8 +179,13 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
         caseDetails.getData().put("orderCollection", orders);
 
         if (featureToggleService.isCloseCaseEnabled() && caseData.getOrderTypeAndDocument().getSubtype() != INTERIM) {
+            String orderAppliesToAllChildren = caseData.getOrderAppliesToAllChildren();
+            if (caseData.getChildren1().size() <= 1) {
+                orderAppliesToAllChildren = "Yes";
+            }
```
tomaszpowroznik, line 185: would not it be better to have this logic inside caseData.getOrderAppliesToAllChildren()?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerAboutToStartTest.java
```diff
@@ -53,10 +57,33 @@ void shouldSetDateOfIssueAsTodayByDefault() {
...
+            .data(Map.of(FAMILY_MAN_CASE_NUMBER_KEY, FAMILY_MAN_CASE_NUMBER_VALUE, CHILDREN_KEY, List.of(testChild())))
+            .build();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseDetails);
+
+        assertThat(callbackResponse.getData().get("pageShow")).isEqualTo("No");
+    }
+
+    @Test
+    void shouldSetPageShowFlagOrderToYesWhenMultipleChildrenOnCase() {
```
tomaszpowroznik, line 76: Order word not needed

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
```diff
@@ -178,8 +179,13 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
         caseDetails.getData().put("orderCollection", orders);
 
         if (featureToggleService.isCloseCaseEnabled() && caseData.getOrderTypeAndDocument().getSubtype() != INTERIM) {
+            String orderAppliesToAllChildren = caseData.getOrderAppliesToAllChildren();
+            if (caseData.getChildren1().size() <= 1) {
+                orderAppliesToAllChildren = "Yes";
```
tomaszpowroznik, line 184: YesNo.Yes.value()

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderController.java
joez-f: sure, that is a good idea

PR #1276, opened 2020-05-28
fix: configure custom thread pool task executor so that async executi…

tomaszpowroznik:
fix for issue introduced in 
https://tools.hmcts.net/jira/browse/FPLA-1758
services invoked from new threads dont have access to request scoped components (RequestData)

based on:
https://stackoverflow.com/questions/23732089/how-to-enable-request-scope-in-async-task-executor

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
@@ -2,10 +2,17 @@ package uk.gov.hmcts.reform.fpl.config;
 
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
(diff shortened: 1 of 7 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
```diff
@@ -24,4 +31,26 @@ public void handleUncaughtException(Throwable throwable, Method method, Object..
         }
     }
 
+    @Override
+    @Bean
+    public Executor getAsyncExecutor() {
```
akomisarek, line 36: Would it be possible to test it somehow by int tests?

tomaszpowroznik: sure, will do

tomaszpowroznik: done

PR #1278, opened 2020-05-28
impr: aligning sdo and cmo (FPLA-NA)

swalker125:
### Change description ###

Renamed order class -> Standard Direction Order
New parent class for cmo and sdo objects with common fields and methods.
Some refactoring of sdo draft logic to use common methods 

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -285,7 +269,7 @@ private HearingBooking getFirstHearing(List<Element<HearingBooking>> hearingBook
     private Map<DirectionAssignee, List<Element<Direction>>> sortDirectionsByAssignee(List<Element<Direction>> list) {
         List<Element<Direction>> nonCustomDirections = commonDirectionService.removeCustomDirections(list);
 
-        return commonDirectionService.sortDirectionsByAssignee(nonCustomDirections);
+        return getMapping(nonCustomDirections);
```
toireasam, line 272: would getMappingOfDirectionToAssignee be a more clear method name? happy to leave if you think it's explanation enough. :)

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/DocmosisTemplates.java
```diff
@@ -4,7 +4,7 @@
     C110A("FL-PLW-APP-ENG-00095.doc", "c110a_application_%s.pdf"),
     C6("FL-PLW-HNO-ENG-00092.doc", "Notice_of_proceedings_c6.pdf"),
     C6A("FL-PLW-HNO-ENG-00093.doc", "Notice_of_proceedings_c6a.pdf"),
-    SDO("FL-PLW-STD-ENG-00099.doc", "standard_directions_order.pdf"),
+    SDO("FL-PLW-STD-ENG-00099.doc", "standard-directions-order.pdf"),
```
toireasam, line 7: is there any reason why we name this pdf with '-' rather than '_' like that others

Jamesnelson117: Hardcoded values where removed from the status enum. Originally using 'standard-direction-order.pdf'. It looks like the format was changed to keep tests passing rather than updating tests.

PR #1285, opened 2020-05-29
impr: refactored directions to comply with logic (FPLA-1470)

swalker125:

https://tools.hmcts.net/jira/browse/FPLA-1470

### Change description ###

Moved logic from service to case data class plus some other minor moving around of logic

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -127,6 +127,19 @@
...
+    public List<Element<Direction>> getDirectionsToComplyWith() {
+        if (getServedCaseManagementOrders().isEmpty() && standardDirectionOrder == null) {
+            return emptyList();
+        }
+
+        if (getServedCaseManagementOrders().isEmpty()) {
+            return standardDirectionOrder.getDirections();
+        }
+
+        return servedCaseManagementOrders.get(0).getValue().getDirections();
```
toireasam, line 140: this could be return unwrapElements(servedCaseManagementOrders).get(0).getDirections(); but is probably fine as is if it's only used the once in this method.

service/src/test/java/uk/gov/hmcts/reform/fpl/model/CaseDataTest.java
```diff
@@ -68,6 +68,46 @@
...
+
+            assertThat(caseData.getDirectionsToComplyWith()).isEqualTo(sdoDirections);
+        }
+
+        @Test
+        void shouldReturnCaseManagementOrderDirectionsWhenServedCaseManagementOrdersIsNotEmpty() {
+            List<Element<Direction>> cmoDirections = wrapElements(directionForParty(LOCAL_AUTHORITY));
+            List<Element<Direction>> sdoDirections = wrapElements(directionForParty(CAFCASS));
+
+            CaseData caseData = CaseData.builder()
```
toireasam, line 91: case data is build here and in method above.. could be extracted and just pass in list as param

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
Jamesnelson117: I think this one is fine as it is.

PR #1293, opened 2020-06-01
fix: generate interim orders when case can be closed from final order

lewisbirks:
### Change description ###

Allow non closable orders to be generated when all children have final order marker issued against them but the case has not been closed.

```
[ ] Yes
[x] No
```


lewisbirks:
should be fixed as part of #1307

PR #1299, opened 2020-06-02
feat: app insights integration with slack (FPLA-1752)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1752

### Change description ###
Adds additional app insights alerts to monitor: 

- Any FPLA exceptions
- Failed Health requests 

```
[ ] Yes
[x] No
```


infrastructure/alerts.tf
```diff
@@ -48,3 +48,39 @@ module "fpl-grant-case-access-failure-alert" {
...
   enabled                    = "${var.enable_alerts}"
 }
+
+module "fpl-exceptions-alert" {
+  source                     = "git@github.com:hmcts/cnp-module-metric-alert"
+  location                   = "${var.appinsights_location}"
+  app_insights_name          = "${var.product}-${var.component}-appinsights-${var.env}"
+  alert_name                 = "${var.product}-exceptions-alert"
+  alert_desc                 = "All exceptions within FPL"
+  app_insights_query         = "exceptions | where operation_Name !contains "health" | project timestamp, severityLevel, operation_Name, outerMessage, ['details'] | sort by severityLevel"
```
akomisarek, line 58: How severrity level is controlled? I thought it is simply loggin all the errors so not sure what information it provides can you provide some info? 

I see details might be too much - it provides the whole stack which we don't need here. 
Please add operation_Id I believe it's most helpful one!

infrastructure/alerts.tf
```diff
@@ -48,3 +48,39 @@ module "fpl-grant-case-access-failure-alert" {
...
+  enabled                    = "${var.enable_alerts}"
+}
+
+module "fpl-health-failure-alert" {
+  source                     = "git@github.com:hmcts/cnp-module-metric-alert"
+  location                   = "${var.appinsights_location}"
+  app_insights_name          = "${var.product}-${var.component}-appinsights-${var.env}"
+  alert_name                 = "${var.product}-health-failure-alert"
+  alert_desc                 = "Failed health requests"
+  app_insights_query         = "requests | where url contains "health" | where resultCode != "200" | project timestamp, resultCode, duration"
```
akomisarek, line 76: Duration might not be super helpful (health is always quick), potentially operationId so we can jump in immediately :).

Thanks

infrastructure/alerts.tf
```diff
@@ -48,3 +48,39 @@ module "fpl-grant-case-access-failure-alert" {
...
+
+module "fpl-health-failure-alert" {
+  source                     = "git@github.com:hmcts/cnp-module-metric-alert"
+  location                   = "${var.appinsights_location}"
+  app_insights_name          = "${var.product}-${var.component}-appinsights-${var.env}"
+  alert_name                 = "${var.product}-health-failure-alert"
+  alert_desc                 = "Failed health requests"
+  app_insights_query         = "requests | where url contains "health" | where resultCode != "200" | project timestamp, operation_Id, resultCode"
+  custom_email_subject       = "Alert: Health failure"
+  frequency_in_minutes       = 5
```
tomaszpowroznik, line 78: maybe we can increase this window, otherwise it will 'spam' constantly when dependant service is down

infrastructure/alerts.tf
```diff
@@ -37,7 +37,7 @@ module "fpl-grant-case-access-failure-alert" {
   app_insights_name          = "${var.product}-${var.component}-appinsights-${var.env}"
   alert_name                 = "${var.product}-grant-case-access-failure"
   alert_desc                 = "Grant case access failure"
-  app_insights_query         = "exceptions | where type contains 'GrantCaseAccessException' | project timestamp , outerMessage"
+  app_insights_query         = "exceptions | where type contains 'GrantCaseAccessException' | project timestamp, operation_Id, outerMessage"
```
tomaszpowroznik, line 40: not sure if now we need this dedicated alert now, the only difference is severity. Now same exception is counted twice

infrastructure/alerts.tf
akomisarek: Well but it will mean our app is not working - cause kubernetes use it as rreadiness check.

We defo need to look at the health check behaviour..

infrastructure/alerts.tf
Jamesnelson117: Yes i'd agree @tomaszpowroznik, only advantage we'd get is grouping I guess. @nkoranne thoughts on removing?

nkoranne: @Jamesnelson117  and @tomaszpowroznik , I am happy to remove it as long as information in alert is visible quickly that 401 error has been reported.  Also agree with the point that this alert can differentiated based on severity.  We can think of adding dedicated alert in future if we see too may such type of exceptions

infrastructure/alerts.tf
nkoranne: I would recommend to have same frequency. It's not important to know which system is down to begin with ... It's important to know that something is down.  Based on this alert our further investigation steps can be started.

PR #1305, opened 2020-06-03
chore(deps): bump launchdarkly-java-server sdk to version 5.0.0

Gogetter:
NA

### Change description ###
Following [launchdarkly updates](https://docs.launchdarkly.com/sdk/server-side/java/migration-4-to-5) dependabot update #1302 failed.

This PR adds a few more resolutions

```
[ ] Yes
[x] No
```


service/build.gradle
```diff
@@ -194,9 +194,8 @@ dependencies {
   compile group: 'org.springframework.security', name: 'spring-security-oauth2-resource-server'
   compile group: 'org.springframework.security', name: 'spring-security-oauth2-jose'
   compile group: 'com.nimbusds', name: 'nimbus-jose-jwt', version: '8.17.1' // override to fix vulnerability issue (CVE-2019-17195)
-  compile group: 'com.squareup.okhttp3', name: 'okhttp', version: '4.7.2'
-  compile group: 'com.launchdarkly', name: 'okhttp-eventsource', version: '1.11.0'
```
Gogetter, line 198: removed both as we have now taken `classifier` from `launchdarkly-java-server-sdk`

service/build.gradle
```diff
@@ -194,9 +194,8 @@ dependencies {
...
-  compile group: 'com.launchdarkly', name: 'launchdarkly-java-server-sdk', version: '4.14.0', classifier: 'thin'
```
Gogetter, line 199: removed `classifier: thin` so we are not forced to use `gson`. See [here](https://docs.launchdarkly.com/sdk/server-side/java/migration-4-to-5#understanding-changes-to-the-json-value-type)

service/src/main/java/uk/gov/hmcts/reform/fpl/config/scheduler/JobFactory.java
```diff
@@ -1,7 +1,6 @@
 package uk.gov.hmcts.reform.fpl.config.scheduler;
 
 import lombok.RequiredArgsConstructor;
-import org.jetbrains.annotations.NotNull;
```
Gogetter, line 4: looks like this was a transitive dep from `com.squareup.okhttp3`

service/build.gradle
akomisarek: I am not sure if I understood your rationing correctly, I don't see how `thin` is "forcing" us to use `gson`. Please see [here]https://docs.launchdarkly.com/sdk/server-side/java/migration-4-to-5#using-other-third-party-dependencies)
Now Gson is simply embedded in the library and we potentially use different version, but I think I am happy with it, if it works I don't see reason to change especially if we can remove one of our dependences.

PR #1313, opened 2020-06-04
impr: remove raw type warnings

lewisbirks:
### Change description ###

Removed instances of raw type warnings. Can check by adding `-Xlint:rawtypes` to

```gradle
tasks.withType(JavaCompile) {
  options.compilerArgs << "-Xlint:unchecked" << "-Werror"
}
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementAboutToSubmitControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementAboutToSubmitControllerTest.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.fpl.controllers;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import org.junit.jupiter.api.Test;
 import org.springframework.boot.test.autoconfigure.OverrideAutoConfiguration;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
@@ -112,8 +113,8 @@ class PlacementAboutToSubmitControllerTest extends AbstractControllerTest {
             .isEqualTo(expectedPlacementWithoutConfidentialDocuments(childPlacement, childApplication));
     }
 
-    private List convertToList(Map<String, Object> updatedCaseDetails, String string) {
-        return mapper.convertValue(updatedCaseDetails.get(string), List.class);
+    private List<?> convertToList(Map<String, Object> updatedCaseDetails, String string) {
+        return mapper.convertValue(updatedCaseDetails.get(string), new TypeReference<>() {});
     }
 
     private Placement placement(Element<Child> child, DocumentReference application) {
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementAboutToSubmitControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementAboutToSubmitControllerTest.java
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.reform.fpl.controllers;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import org.junit.jupiter.api.Test;
 import org.springframework.boot.test.autoconfigure.OverrideAutoConfiguration;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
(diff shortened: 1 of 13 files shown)


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementAboutToSubmitControllerTest.java
```diff
@@ -112,8 +113,8 @@ void shouldAddConfidentialPlacementAndPlacementsWithoutPlacementOrderToCaseDetai
             .isEqualTo(expectedPlacementWithoutConfidentialDocuments(childPlacement, childApplication));
     }
 
-    private List convertToList(Map<String, Object> updatedCaseDetails, String string) {
-        return mapper.convertValue(updatedCaseDetails.get(string), List.class);
+    private List<?> convertToList(Map<String, Object> updatedCaseDetails, String string) {
+        return mapper.convertValue(updatedCaseDetails.get(string), new TypeReference<>() {});
```
Gogetter, line 117: just curious why we using `TypeReference` why not `List.class`?

lewisbirks: List just returns a list without any type value, by adding the wildcard to the return type List no longer matches this and as such a type reference is required to convert to the correct type.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/docmosis/GeneratedOrderTemplateDataGeneration.java
```diff
@@ -41,6 +41,8 @@ public GeneratedOrderTemplateDataGeneration(
         this.localAuthorityNameLookupConfiguration = localAuthorityNameLookupConfiguration;
     }
 
+    // If we use <?, ?> there is then an issue with sonar complaining about not using wildcard types as return types
+    @SuppressWarnings("rawtypes")
```
lewisbirks, line 45: Any suggestions on this would be welcome

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementAboutToSubmitControllerTest.java
Gogetter: this is correct. But behind the scenes `mapper.convertValue` uses `TypeReference`

lewisbirks: will test and see if list works, usually we just use type reference so that is what I did but maybe that is only required when we are requesting a specific type

service/src/main/java/uk/gov/hmcts/reform/fpl/service/docmosis/GeneratedOrderTemplateDataGeneration.java
Gogetter: for now, I think we can use the suppression until we can rethink the Docmosis object-ing approach

lewisbirks: Yeah, the issue is with the use of `@SuperBuilder`. My suggestion would be to move to how we have started doing it with the notifications. There we don't use the builder pattern but instead just use setter methods. If we wanted to, we could even instead of doing lombok setters we could make custom setters that return the object as well allowing chaining of the setter methods.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/PlacementAboutToSubmitControllerTest.java
lewisbirks: Seems like it works with `List.class` works so will change back to that

service/src/main/java/uk/gov/hmcts/reform/fpl/service/docmosis/GeneratedOrderTemplateDataGeneration.java
lewisbirks: Actually, looks like lombok supports this as well
>lombok.accessors.chain = [true | false] (default: false)
If set to true, generated setters will return this (instead of void). An explicitly configured chain parameter of an @Accessors annotation takes precedence over this setting.

Gogetter: yes, something we can definitely explore. Do you have a Tech Enhancement ticket for it?

lewisbirks: not yet

PR #1314, opened 2020-06-04
fix: Handle missing child dob when generating orders

tomaszpowroznik:


service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
```diff
@@ -131,6 +131,24 @@ void shouldReturnChildrenDetailsWhenPopulatedChildren() {
...
+            .party(ChildParty.builder()
+                .firstName("John")
+                .lastName("Smith")
+                .build())
+            .build());
+
+        DocmosisChild expectedChild = DocmosisChild.builder()
+            .name("John Smith")
+            .gender("Male")
+            .dateOfBirth(null)
```
lewisbirks, line 146: Unless you want to be specific, this shouldn't be needed as null is the default value if you don't assign it a value

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
```diff
@@ -131,6 +131,24 @@ void shouldReturnChildrenDetailsWhenPopulatedChildren() {
...
+        List<Element<Child>> listChildren = wrapElements(Child.builder()
+            .party(ChildParty.builder()
+                .firstName("John")
+                .lastName("Smith")
+                .build())
+            .build());
+
+        DocmosisChild expectedChild = DocmosisChild.builder()
+            .name("John Smith")
+            .gender("Male")
```
lewisbirks, line 145: Haven't specified a gender in the built child above

tomaszpowroznik: stupid me :) that explains test failure ;)done

service/src/test/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionServiceTest.java
tomaszpowroznik: test should not silently assume anything, how do you know if code doesnt set it to "" or "unknown" or "N/A". This test check that it does not

lewisbirks: Thought so

PR #1329, opened 2020-06-09
Rename internal-change events to avoid URL double-encoding bug

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-1864

### Change description ###

Changed 'internal-change:' to 'internal-change-' (colon : -> hyphen -)

```
[ ] Yes
[X] No
```


charts/fpl-case-service/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: fpl-case-service
 apiVersion: v1
 home: https://github.com/hmcts/fpl-ccd-configuration
-version: 1.11.5
+version: 1.11.6
```
Gogetter, line 4: do we need this and the other chart changes here?

joez-f: this configured preview to use the same ccd images as prod which let us test that it fixed the problem; I think we can merge the chart changes with this PR and close the other one

tomaszpowroznik:
please double check if this wont break backward compatibility, but i think it wont

PR #1334, opened 2020-06-10
fix: mutation tests (FPLA-1844)

lewisbirks:

Part of [FPLA-1844](https://tools.hmcts.net/jira/browse/FPLA-1844)

### Change description ###

Bumped the gradle pitest plugin to latest version.

Fixed an issue where the integration tests when being run through the pitest read from the test resource directory and as such the tests dependant on ordersConfig.json failed.
When the file was just copied into the `integrationTest/resources` it would still read from the test directory due to what I assume to be a name conflict. As such functionality was introduced to dynamically populate the location of the order configuration file, which can be defined in the `application.yaml`.

A further PR will be created to fix the nested test functionality.

This still needs to be tested in the `nightly-dev` pipeline but I want to get local passing first.

```
[ ] Yes
[x] No
```


lewisbirks:
closing and will open an updated one as a replacement

PR #1343, opened 2020-06-11
fix: missing submitted callback in hearing details (FPLA-1884)

joez-f:
### JIRA Link ###

https://tools.hmcts.net/jira/browse/FPLA-1884

### Change description ###

Added missing submitted callback, was most likely removed by accident during a refactor

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
```diff
@@ -103,7 +104,7 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
     @PostMapping("/submitted")
     public void handleSubmittedEvent(@RequestBody CallbackRequest callbackRequest) {
         CaseData caseData = mapper.convertValue(callbackRequest.getCaseDetails().getData(), CaseData.class);
-        if (standardDirectionsService.hasEmptyDates(caseData)) {
+        if (caseData.getState() == GATEKEEPING && standardDirectionsService.hasEmptyDates(caseData)) {
```
lewisbirks, line 107: Not sure if this field is updated automatically by CCD (would assume so but haven't tested it for anything other than changing the case state) but the `caseDetails.getState()` definitely is

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerSubmittedTest.java
```diff
@@ -57,6 +57,22 @@ void shouldTriggerPopulateDatesEventWhenThereAreEmptyDates() {
...
+                .id(12345L)
+                .jurisdiction(JURISDICTION)
+                .caseTypeId(CASE_TYPE)
+                .data(Map.of("state", "Submitted"))
+                .build())
+            .build();
+
+        postSubmittedEvent(callbackRequest);
+
+        verify(coreCaseDataService, never()).triggerEvent(any(), any(), any(), any(), any());
```
tomaszpowroznik, line 71: be careful with negative assertions like this, when async procession it does not prove it has not be invoked, e.g it could be 100ms later

joez-f: Have changed to use checkThat(), please check if its ok

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
akomisarek: @joez-f does it mean we are changing that code or what we do~?

lewisbirks: @akomisarek turns out that CCD doesn't update that field so it is only used for updating the case state from a controller and nothing more. If you want to work out the current state you need to get the state from case details, which is annoying as that is as a `String` and not as our `State` enum.

service/src/main/resources/application-feature-toggle.yml
```diff
@@ -0,0 +1,4 @@
+spring:
```
lewisbirks, line 1: Should not be committed, the extension is wrong and as such has not been picked up by the gitignore `yml` -> `yaml`

joez-f: oops 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
joez-f: have changed it now, I thought caseData.getState() actually worked when I tested but lewis is right it does not work, I didn't test every scenario (doug spotted it though!)

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
```diff
@@ -103,7 +104,8 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
     @PostMapping("/submitted")
     public void handleSubmittedEvent(@RequestBody CallbackRequest callbackRequest) {
         CaseData caseData = mapper.convertValue(callbackRequest.getCaseDetails().getData(), CaseData.class);
-        if (standardDirectionsService.hasEmptyDates(caseData)) {
+        if (callbackRequest.getCaseDetails().getState().equals(GATEKEEPING.getValue())
```
lewisbirks, line 107: There is a method in CaseDetailsHelper that might be nicer to use here

lewisbirks: think it is called `inState` or something like that

PR #1346, opened 2020-06-15
temp: Temporary fix for dm domain regex

tomaszpowroznik:
temp fix until bug fixed:
https://tools.hmcts.net/jira/browse/RDM-8742


lewisbirks:
> what's expected default?

@akomisarek default is
`https?://dm-store.*/documents/[A-Za-z0-9-]+(?:/binary)?`

PR #1357, opened 2020-06-19
fix: CTSC feature toggle fix (FPLA-1912)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1912

### Change description ###
Fixes a bug in which feature toggle service was setting CTSC to true.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleService.java
```diff
@@ -13,14 +13,15 @@
     private final LDClient ldClient;
     private final LDUser ldUser;
     private final LDUser.Builder ldUserBuilder;
```
tomaszpowroznik, line 15: please remove builder instance

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisDocument.java
```diff
@@ -0,0 +1,4 @@
+package uk.gov.hmcts.reform.fpl.model.docmosis;
```
akomisarek, line 1: This class seems reduntant.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisDocument.java
```diff
@@ -0,0 +1,4 @@
...
+
+public class DocmosisDocument {
+}
```
Jamesnelson117, line 4: Removing, unintentional add.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleService.java
```diff
@@ -13,14 +13,15 @@
...
+    private final String ldUserKey;
 
     @Autowired
     public FeatureToggleService(LDClient ldClient, @Value("${ld.user_key}") String ldUserKey) {
         this.ldClient = ldClient;
         this.ldUserBuilder = new LDUser.Builder(ldUserKey)
             .custom("timestamp", String.valueOf(System.currentTimeMillis()));
```
akomisarek, line 22: Please fix other instances of the problem at the same time - tthis timestamp is broken - it would always send data when the userBuilder was created which seems wrong (it would work if the timestamp would evaluate in lambda)

Notification for judge seems to be broken as well, as Tomek suggested the builder should be remove dfrom class, it's bad design.

Additionoally, please prove in the tests.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -53,8 +64,15 @@ public void shouldMakeCorrectCallForCtsc(Boolean toggleState) {
...
+        featureToggleService.isCtscReportEnabled();
 
-        verify(ldClient).boolVariation(eq("CTSC"), any(LDUser.class), eq(false));
+        verify(ldClient, atLeastOnce()).boolVariation(eq("CTSC"), ldUser.capture(), eq(false));
+
+        LDUser expectedLdUser = getLdBuilder()
+            .custom("localAuthorityName", "")
+            .build();
+
+        assertThat(ldUser.getAllValues().get(0).getCustomAttributes()).isEqualTo(expectedLdUser.getCustomAttributes());
```
akomisarek, line 75: getAllvalues can be replaces with getLastValue (or getValue) which makes the assertion nicer. .

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -73,8 +91,11 @@ public void shouldMakeCorrectCallForFees(Boolean toggleState) {
         givenToggle(toggleState);
```
akomisarek, line 92: Test above (shouldMakeCorrectCallForCtscReport) doesn't validate the submitted request, intended?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
lewisbirks: note that this will change the assertion as `getValue()` calls get last value and as the verify captures 2 calls the last value captured will be from `featureToggleService.isCtscReportEnabled()` instead of `featureToggleService.isCtscEnabled("test name")`

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -53,8 +64,15 @@ public void shouldMakeCorrectCallForCtsc(Boolean toggleState) {
...
 
         assertThat(featureToggleService.isCtscEnabled("test name")).isEqualTo(toggleState);
+        featureToggleService.isCtscReportEnabled();
```
lewisbirks, line 67: Should this be part of this test? We have tests for each of the individual methods with this one being for the `isCtscEnabled` method. Should we be doing this additional verification of objects in a different test as it is testing different functionality.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -73,8 +97,11 @@ public void shouldMakeCorrectCallForFees(Boolean toggleState) {
         givenToggle(toggleState);
 
         assertThat(featureToggleService.isFeesEnabled()).isEqualTo(toggleState);
+        featureToggleService.isCtscReportEnabled();
 
-        verify(ldClient).boolVariation(eq("FNP"), any(LDUser.class), eq(false));
+        verify(ldClient, atLeastOnce()).boolVariation(eq("FNP"), ldUser.capture(), eq(false));
```
lewisbirks, line 102: `atLeastOnce()` is unnecessary. There is only going to be one interaction with the "FNP" feature flag as `isCtscReportEnabled()` uses "CTSC"

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -83,8 +110,11 @@ public void shouldMakeCorrectCallForPayments(Boolean toggleState) {
         givenToggle(toggleState);
 
         assertThat(featureToggleService.isPaymentsEnabled()).isEqualTo(toggleState);
+        featureToggleService.isCtscReportEnabled();
 
-        verify(ldClient).boolVariation(eq("payments"), any(LDUser.class), eq(false));
+        verify(ldClient, atLeastOnce()).boolVariation(eq("payments"), ldUser.capture(), eq(false));
```
lewisbirks, line 115: same here

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -93,8 +123,11 @@ public void shouldMakeCorrectCallForExpertUI(Boolean toggleState) {
         givenToggle(toggleState);
 
         assertThat(featureToggleService.isExpertUIEnabled()).isEqualTo(toggleState);
+        featureToggleService.isCtscReportEnabled();
 
-        verify(ldClient).boolVariation(eq("expert-ui"), any(LDUser.class), eq(false));
+        verify(ldClient, atLeastOnce()).boolVariation(eq("expert-ui"), ldUser.capture(), eq(false));
```
lewisbirks, line 128: and here

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
lewisbirks: I am confused as to why this has added this to every test.

lewisbirks: In my opinion we should not have this added to any of the current tests. We should have a new test called something along the lines of `shouldNotCarryAttributesBetweenRequests` that can call the 2 ctsc flags and check that the attributes are not carried over between calls.

akomisarek: hmm ok, but how would you ensure every methods works as expected? If we call only 2 ctsc flags, we don't get confirmation allocation judge is free from pollution. Unless we are happy with single test + code review process? But I would like somehow to ensure every test sends only what is expected and only it.

Maybe it's matter of adding dummy invocation in before? Or like all dummy invocations to make it easier and not specific. Iterate using reflection :D

lewisbirks: Ahh I see what you mean, I was confused as to why we were making this call on all methods but I see your point, will have a think and see if there is a nice way to implement it

lewisbirks: Also this probably wouldn't have been caught even if we had tested the users being passed as we reset the FeatureToggle instance before each test

lewisbirks: Made [this](https://github.com/hmcts/fpl-ccd-configuration/compare/FPLA-1912-CTSC-Feature-Toggle-Fix...dev/fpla-1912-test-example) as an example of what could be done

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -22,143 +34,246 @@
...
 
-    private FeatureToggleService featureToggleService;
+    @Captor
+    private ArgumentCaptor<LDUser> ldUser;
 
-    @BeforeEach
-    void setup() {
-        featureToggleService = new FeatureToggleService(ldClient, LD_USER_KEY);
+    @AfterEach
+    void tearDown() {
```
lewisbirks, line 49: I think there could possibly be a better name for this method as it is tearing anything down.
I think we use resetInvocations else where

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -22,143 +34,246 @@
...
         givenToggle(toggleState);
 
-        assertThat(featureToggleService.isXeroxPrintingEnabled()).isEqualTo(toggleState);
+        assertThat(service.isXeroxPrintingEnabled()).isEqualTo(toggleState);
+        verify(ldClient).boolVariation(eq("xerox-printing"), ldUser.capture(), eq(false));
+
+        LDUser expectedLdUser = getLdBuilder().build();
 
-        verify(ldClient).boolVariation(eq("xerox-printing"), any(LDUser.class), eq(false));
+        assertThat(ldUser.getValue().getCustomAttributes()).isEqualTo(expectedLdUser.getCustomAttributes());
```
lewisbirks, line 63: Do we want to keep this capture now that we are doing the essentially the same thing in the `shouldNotCarryAttributesBetweenRequests` test (same for other tests as well)?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleServiceTest.java
```diff
@@ -196,22 +156,17 @@ void shouldMakeCorrectCallForAllocatedJudgeNotificationC2Application(Boolean tog
...
     @MethodSource("userAttributesTestSource")
-    void shouldNotCarryAttributesBetweenRequests(Runnable functionToTest, Runnable accumulateFunction,
+    void shouldNotAccumulateAttributesBetweenRequests(Runnable functionToTest, Runnable accumulateFunction,
                                                  List<UserAttribute> attributes) {
         accumulateFunction.run();
         functionToTest.run();
 
         verify(ldClient, times(2)).boolVariation(anyString(), ldUser.capture(), anyBoolean());
-
-        assertThat(ldUser.getValue().getCustomAttributes()).containsExactlyInAnyOrderElementsOf(attributes);
```
lewisbirks: Ahh, we need this, might have got confused in the chat

Jamesnelson117: My bad, got a bit delete happy :)

PR #1364, opened 2020-06-22
fix:add zap timestamp disclousure exception (part of uuid matches tim…

tomaszpowroznik:
…espamp regexpused by rule)

Affected are controllers that in about to start expand collection element giving it random uuid. 
ZAP rule 10096 uses \\b[0-9]{8,10}\\b to find timestamps in response body. Generated uuid can contains such sequence and as a result rule fails randomly
see:
https://github.com/zaproxy/zap-extensions/blob/master/addOns/pscanrules/src/main/java/org/zaproxy/zap/extension/pscanrules/TimestampDisclosureScanner.java


audit.json
```diff
@@ -2,7 +2,6 @@
   "100000_A Server Error response code was returned by the server_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/case-extension/mid-event_POST": "ignore",
   "100000_A Server Error response code was returned by the server_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/placement/mid-event_POST": "ignore",
   "100000_A Server Error response code was returned by the server_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/case-extension/about-to-start_POST": "ignore",
-  "100000_A Server Error response code was returned by the server_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/case-extension/about-to-start_POST": "ignore",
```
tomaszpowroznik, line 5: duplicate of above

PR #1365, opened 2020-06-22
impr: Increase e2e parallelism

tomaszpowroznik:
nightly build e2e tooks over 20 minutes whereas preview less than 10. Difference is number of e2e workers (5 vs 10)

### DIFF
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -16,6 +16,9 @@ properties([
     string(name: 'SECURITY_RULES',
       defaultValue: 'https://raw.githubusercontent.com/hmcts/security-test-rules/FPLA-security-test-rules-DO-NOT-MERGE/conf/security-rules.conf',
       description: 'The security rules to use'),
+    string(name: 'FUNCTIONAL_TESTS_WORKERS',
+      defaultValue: '10',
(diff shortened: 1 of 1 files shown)


Jenkinsfile_nightly
```diff
@@ -46,6 +46,7 @@ withNightlyPipeline(type, product, component) {
 
   before('fullFunctionalTest') {
     env.PROXY = params.PROXY_SERVER
+    env.PARALLEL_CHUNKS = 10
```
lewisbirks, line 49: Want to add it as a build param?

tomaszpowroznik: can be added, but not sure if needed

tomaszpowroznik: done, but looks that numeric types are not supported

lewisbirks: Wasn't sure either but just mentioning the possibility

lewisbirks: > done, but looks that numeric types are not supported

Yeah seems that way from [here](https://www.jenkins.io/doc/book/pipeline/syntax/#parameters). Do you think that any conversions will need to take place or javascript will handle that all for us (thinking the latter but want to double check)

PR #1375, opened 2020-06-24
fix:Fix request data in async task issue

tomaszpowroznik:
Currently RequestData accessed from @Async executor may be corrupted (returns nulls) when main request thread already committed http response (depends on timing). It can affect payments, share a case and notifications.
New approach is doing copy of RequestData in main request thread and attach it to every new thread spanned as ThreadLocal variable so accessing it from executor works fine even if http requested is closed.

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
@@ -1,14 +1,18 @@
 package uk.gov.hmcts.reform.fpl.config;
 
+import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.aop.interceptor.AsyncUncaughtExceptionHandler;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.task.TaskDecorator;
 import org.springframework.scheduling.annotation.AsyncConfigurer;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
-import org.springframework.web.context.request.RequestAttributes;
-import org.springframework.web.context.request.RequestContextHolder;
+import uk.gov.hmcts.reform.fpl.request.RequestData;
+import uk.gov.hmcts.reform.fpl.request.RequestDataCache;
+import uk.gov.hmcts.reform.fpl.request.SimpleRequestData;
 
 import java.lang.reflect.Method;
 import java.util.concurrent.Executor;
@@ -16,7 +20,10 @@ import javax.annotation.Nonnull;
 
 @Slf4j
 @Configuration
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class AsyncConfiguration implements AsyncConfigurer {
+
+    private final ApplicationContext context;
 
     @Override
     public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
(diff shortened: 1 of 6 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/request/CacheAwareRequestData.java
```diff
@@ -0,0 +1,31 @@
...
+
+@Primary
+@Service
+public class CacheAwareRequestData implements RequestData {
+
+    private final HttpServletRequest httpServletRequest;
+
+    @Autowired
+    public CacheAwareRequestData(HttpServletRequest httpServletRequest) {
+        this.httpServletRequest = httpServletRequest;
```
lewisbirks, line 17: Does this servlet request still have a possibility of getting cleaned up before we have used it?

tomaszpowroznik: If RequestData is accessed from async task then AsyncConfiguration guarantees that there will be cached version of RequestData and httpRequest wont be consulted.
If Request data is accessed from standard "request"  thread then it is guaranteed by spring request scope that httpServletRequest is valid.
The question is about when exactly cached version is created, and it created in request thread.

 @Override
        public Runnable decorate(@Nonnull Runnable task) {
            //request thread where we create copy of request data
            SimpleRequestData requestData = new SimpleRequestData(context.getBean(RequestData.class));

           //everything in lambda is executed in new task thread
            return () -> {
                RequestDataCache.add(requestData);
                try {
                    task.run();
                } finally {
                    RequestDataCache.remove();
                }
            };
        }

tomaszpowroznik: To test it i have setup code like this:
    @PostMapping("/submitted")
        System.out.println("Invoking async action " + Thread.currentThread().getName());
        localAuthorityUserService.grantUserAccessWithCaseRole(caseId, caseLocalAuthority);
    }

in AsyncConfiguration
 @Override
        public Runnable decorate(@Nonnull Runnable task) {
            System.out.println("Copying request data " + Thread.currentThread().getName());
            SimpleRequestData requestData = new SimpleRequestData(context.getBean(RequestData.class));

            return () -> {
                System.out.println("Setting thread local"+Thread.currentThread().getName());
                RequestDataCache.add(requestData);
                try {
                    task.run();
                } finally {
                    RequestDataCache.remove();
                }
            };
        }

output is: 
2020-06-24T17:32:15.734 INFO  [http-nio-4000-exec-1] u.g.h.r.logging.filters.RequestStatusLoggingFilter Request POST /callback/case-initiation/about-to-submit processed in 242ms
Invoking async action http-nio-4000-exec-2
Copying request data http-nio-4000-exec-2
Setting thread localgetAsyncExecutor-1

which proves that copy of request data is done in request thread where correct httpRequest is guranteed by spring

lewisbirks: Sounds good to me.
On a slight side note if you care, you can change the name from `localgetAsyncExecutor` to something else in the `@Bean(name="something")`

service/src/test/java/uk/gov/hmcts/reform/fpl/request/RequestDataTest.java
```diff
@@ -9,28 +12,63 @@
...
     @InjectMocks
-    private RequestData requestData;
+    private CacheAwareRequestData requestData;
 
-    @Test
-    public void shouldReturnAuthorisationHeaderFromRequest() {
-        when(httpServletRequest.getHeader("authorization")).thenReturn(AUTH_TOKEN);
-        assertThat(requestData.authorisation()).isEqualTo(AUTH_TOKEN);
+    @Nested
+    class NotCachedRequestData {
```
lewisbirks, line 32: test methods don't need to be public

service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
```diff
@@ -35,20 +41,28 @@ public void handleUncaughtException(Throwable throwable, Method method, Object..
     @Bean
     public Executor getAsyncExecutor() {
         ThreadPoolTaskExecutor taskExecutor = new ThreadPoolTaskExecutor();
-        taskExecutor.setTaskDecorator(new AsyncTaskDecorator());
+        taskExecutor.setTaskDecorator(new AsyncTaskDecorator(context));
```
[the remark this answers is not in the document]
tomaszpowroznik: cannot be injected into method, this method signature is enforced by interface

service/src/test/java/uk/gov/hmcts/reform/fpl/request/RequestDataTest.java
tomaszpowroznik: fixed

service/src/main/java/uk/gov/hmcts/reform/fpl/config/AsyncConfiguration.java
```diff
@@ -18,6 +21,9 @@
 @Configuration
 public class AsyncConfiguration implements AsyncConfigurer {
 
+    @Autowired
+    private ApplicationContext context;
```
lewisbirks, line 25: Do we want to have field injection or is constructor injection better?

tomaszpowroznik: yeah, for consistency it will be better to change it

PR #1380, opened 2020-06-25
feat: Send case application resubmission emails (FPLA-1822)

tomaszpowroznik:
Story: https://tools.hmcts.net/jira/browse/FPLA-1822
Template: https://www.notifications.service.gov.uk/services/12f756df-f01d-4a32-a405-e1ea8a494fbb/templates/d2e44b62-436c-4bb6-b3b9-afa4f2bca7b5


service/src/main/java/uk/gov/hmcts/reform/fpl/model/notify/returnedcase/ReturnedCaseTemplate.java
```diff
@@ -1,11 +1,13 @@
...
+import lombok.Builder;
+import lombok.Data;
 import uk.gov.hmcts.reform.fpl.model.notify.NotifyData;
 
-@Getter
-@Setter
+import java.util.Map;
+
+@Data
+@Builder
```
lewisbirks, line 10: Thoughts on using this instead https://projectlombok.org/features/experimental/Accessors.
So that we don't run into the annoyances described here https://tools.hmcts.net/jira/browse/FPLA-1856

tomaszpowroznik: looks interesting, the only think is that builders promotes usage of immutable objects, with accessors we can nicely mimic builder syntax but immutability is the price

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerSubmittedTest.java
```diff
@@ -277,71 +267,59 @@ void shouldSendFailedPaymentNotificationOnHiddenDisplayAmountToPay() {
...
+            checkUntil(() -> resubmissionNotificationsSent(HMCTS_ADMIN_EMAIL));
+            checkThat(this::paymentNotTakenAndNoMoreEmailsSent);
+        }
 
-    @Test
-    void shouldNotNotifyAdminOfAReturnedCaseWhenTheCaseIsSubmittedFromOpenState() {
-        CaseDetails caseDetails = enableSendToCtscOnCaseDetails(YES);
-        caseDetails.getData().put(DISPLAY_AMOUNT_TO_PAY, NO.getValue());
+        @Test
+        void shouldNotifyCtscAndCafcassWhenCaseIsResubmitted() {
```
akomisarek, line 294: Are we testing feature toggle value here? Not sure if much value testing but if the logic is in the controller we need. to live with it.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -154,9 +154,9 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmitEvent(@RequestBod
 
     @PostMapping("/submitted")
     public void handleSubmittedEvent(@RequestBody @NotNull CallbackRequest callbackRequest) {
-        applicationEventPublisher.publishEvent(new SubmittedCaseEvent(callbackRequest));
-
-        if (isInReturnedState(callbackRequest.getCaseDetailsBefore())) {
+        if (isInOpenState(callbackRequest.getCaseDetailsBefore())) {
+            applicationEventPublisher.publishEvent(new SubmittedCaseEvent(callbackRequest));
+        } else if (isInReturnedState(callbackRequest.getCaseDetailsBefore())) {
```
akomisarek, line 159: Stupid question, for the future - should. we throw illegalStateException in. else block? If someone adds a new state, it will fail silently, and with the exception it might be easier to find, thoughts?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/AmendedReturnedCaseEventHandler.java
```diff
@@ -27,10 +30,22 @@ public void notifyAdmin(AmendedReturnedCaseEvent event) {
...
-        ReturnedCaseTemplate parameters = returnedCaseContentProvider
-            .buildNotificationParameters(eventData.getCaseDetails(), eventData.getLocalAuthorityCode());
+        ReturnedCaseTemplate parameters = contentProvider.parametersWithCaseUrl(eventData.getCaseDetails());
 
-        notificationService.sendEmail(AMENDED_APPLICATION_RETURNED_TO_THE_ADMIN, email, parameters,
+        notificationService.sendEmail(AMENDED_APPLICATION_RETURNED_ADMIN_TEMPLATE, email, parameters,
+            eventData.getReference());
+    }
+
+    @Async
```
akomisarek, line 39: I  like the async Usage, makes it better for the user.

Do you know how big by default is that async thread pool?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -479,4 +479,6 @@ public boolean isClosedFromOrder() {
     public boolean allocatedJudgeExists() {
         return allocatedJudge != null;
     }
+
+    private final DocumentReference submittedForm;
```
akomisarek, line 483: We never had this as a field? Surprised!

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/ReturnedCaseContentProvider.java
```diff
@@ -2,41 +2,62 @@
...
+            .respondentLastName(getFirstRespondentLastName(caseData.getRespondents1()))
+            .familyManCaseNumber(defaultIfNull(caseData.getFamilyManCaseNumber(), ""))
+            .returnedReasons(returnApplication.getFormattedReturnReasons())
+            .returnedNote(returnApplication.getNote());
+    }
 
-        return returnedCaseTemplate;
+    private Map<String, Object> linkToAttachedDocument(final DocumentReference documentReference) {
+        byte[] content = documentDownloadService.downloadDocument(documentReference.getBinaryUrl());
+        return generateAttachedDocumentLink(content).map(JSONObject::toMap).orElse(null);
```
akomisarek, line 61: Are we silently ignoring when there are issues with documents? That makes me a bit worried.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
tomaszpowroznik: It wont fail, it just wont do anything, which i think is fine as it it not clear what would be business req for this future story

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/AmendedReturnedCaseEventHandler.java
tomaszpowroznik: In AsyncConfigrattion we deined to use
ThreadPoolTaskExecutor
which have

public class ThreadPoolTaskExecutor extends ExecutorConfigurationSupport
		implements AsyncListenableTaskExecutor, SchedulingTaskExecutor {

	private final Object poolSizeMonitor = new Object();

	private int corePoolSize = 1;

	private int maxPoolSize = Integer.MAX_VALUE;

PR #1382, opened 2020-06-26
fix: persist representedBy in respondent about to submit (FPLA-1927)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1927

### Change description ###
Fix to an issue where representedBy was not being persisted in amendRespondents about to submit. 

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/RespondentControllerTest.java
```diff
@@ -90,6 +83,23 @@ void shouldReturnNoDateOfBirthErrorsForRespondentWhenValidDateOfBirth() {
...
+            .build();
+
+        AboutToStartOrSubmitCallbackResponse callbackResponse 
+            = postMidEvent(caseDetails, "persist-representatives");
+
+        CaseData caseData = mapper.convertValue(callbackResponse.getData(), CaseData.class);
+        Respondent firstRespondent = caseData.getRespondents1().get(0).getValue();
+        Respondent secondRespondent = caseData.getRespondents1().get(0).getValue();
+
+        assertThat(firstRespondent.getPersistRepresentedBy()).isEqualTo(YES.getValue());
```
akomisarek, line 99: Minor question - do you think it would be worth checking in above about to start or something that this value is not actually YES?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Respondent.java
```diff
@@ -29,6 +29,7 @@
     @NotNull(message = "You need to add details to respondents")
     private final RespondentParty party;
     private final String leadRespondentIndicator;
+    private String persistRepresentedBy;
```
akomisarek, line 32: Probably this could be of type "YesNo"?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -54,6 +56,15 @@ void shouldBuildExpectedLabelWhenEmptyList() {
             assertThat(result).isEqualTo("No respondents on the case");
         }
 
+        @Test
+        void shouldSetPersistRepresentativeFlagToYes() {
+            List<Element<Respondent>> updatedRespondents = service.setPersistRepresentativeFlag(getRespondents());
+            List<Respondent> respondentsList = unwrapElements(updatedRespondents);
+
+            assertThat(respondentsList.get(0).getPersistRepresentedBy()).isEqualTo(YesNo.YES.getValue());
+            assertThat(respondentsList.get(1).getPersistRepresentedBy()).isEqualTo(YesNo.YES.getValue());
```
akomisarek, line 65: As mentioned earlier the persistRepresentedBy could be YesNo type so getValue could be avoided.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/RespondentControllerTest.java
Jamesnelson117: Sorry, do you mean before we call the mid event (not about to start)?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Respondent.java
Jamesnelson117: CCD still treats fieldType 'YesOrNo' value as a string so unfortunately above still needs be a string.

akomisarek: Ah ok we could probably achieve that by having some serializers / desialiers but it's not in scope of this. Would you mind opening tech debt?

lewisbirks: So what could be done here is that the YesNo could have a `@JsonProperty` annotation added to it for "Yes" and "No" respectively, the issue is that this will then break some parts of the service as they work on the assumption that the return value from the enum being serialised is going to be upper case.

PR #1385, opened 2020-06-29
feat: disabled organisation API returning of roles (FPLA-1921)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1921

### Change description ###
Disabled organisation API returning of user roles

```
[ ] Yes
[x] No
```


tomaszpowroznik:
what about /fpla-docker/mocks/wiremock/mappings/organisations.json, does it need to be updated?

service/src/main/java/uk/gov/hmcts/reform/rd/client/OrganisationApi.java
```diff
@@ -19,7 +19,8 @@
     OrganisationUsers findUsersByOrganisation(
         @RequestHeader(AUTHORIZATION) String authorisation,
         @RequestHeader(SERVICE_AUTHORIZATION) String serviceAuthorization,
-        @RequestParam(value = "status") Status status
+        @RequestParam(value = "status") Status status,
+        @RequestParam(value = "returnRoles") Boolean returnRoles
```
akomisarek, line 23: Could we abstract this method to not expose this parameter? As easier Api for use would it be possible to expose client without this method and always pass false?

tomaszpowroznik: not sure if we should, it is interface definition. We have parameter hidden at service layer

PR #1387, opened 2020-06-30
feat: notice of new hearing added (FPLA-1879)

KiranVarma02:

https://tools.hmcts.net/jira/browse/FPLA-1879

### Change description ###
notice of new hearing added

```
[ ] Yes
[* ] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
```diff
@@ -104,9 +116,13 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
     @PostMapping("/submitted")
     public void handleSubmittedEvent(@RequestBody CallbackRequest callbackRequest) {
         CaseData caseData = mapper.convertValue(callbackRequest.getCaseDetails().getData(), CaseData.class);
+
         if (isInGatekeepingState(callbackRequest.getCaseDetails())
             && standardDirectionsService.hasEmptyDates(caseData)) {
             applicationEventPublisher.publishEvent(new PopulateStandardDirectionsOrderDatesEvent(callbackRequest));
         }
+
+        applicationEventPublisher.publishEvent(new NewHearingsAddedEvent(callbackRequest));
```
akomisarek, line 125: Being picky here (from the domain perspective)

If we name the NewHearingsAdded it should be ONLY raised if the new hearings are added.

I remember it was originally called HearingsUpdated which is different (but again we should detect if hearings were updated). It's tricky but the more correct you are with events the easier it's for consumers to consume.

Additionally, I would remove Event from the name, it seems for me as "smurph naming" 
NewHearingsAdded already denots the event (and the package name as well). Please consider refactoring.

service/src/main/java/uk/gov/hmcts/reform/fpl/events/NewHearingsAddedEvent.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Generated;
+import lombok.Value;
+import uk.gov.hmcts.reform.ccd.client.model.CallbackRequest;
+
+
+@Value
+@Generated
+@EqualsAndHashCode(callSuper = true)
+public class NewHearingsAddedEvent extends CallbackEvent {
```
akomisarek, line 12: As mentioned somewhere else I don't like this to have "Event" in the name, past tense + package already suggest that.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NewHearingsAddedEventHandler.java
```diff
@@ -0,0 +1,101 @@
...
+
+        final CaseDetails caseDetails = mapper
+            .convertValue(event.getCallbackRequest().getCaseDetails(), CaseDetails.class);
+        final CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+
+        List<Element<HearingBooking>> hearings = caseData.getHearingDetails();
+        hearings.removeAll(hearingBookingService.getPastHearings(caseData.getHearingDetails()));
+
+        List<Element<HearingBooking>> selectedHearings = hearingBookingService.getSelectedHearings(
+            caseData.getNewHearingSelector(), hearings);
```
akomisarek, line 57: I believe this logic shouldn't be here.

If we have similar handler (i.e. not sending the email, but doing something else), we would need to repeat the logic.

Ideally best if the new hearings added would be already in the event "payload" is it possible?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NewHearingsAddedEventHandler.java
```diff
@@ -0,0 +1,101 @@
...
+
+        selectedHearings.forEach(hearing -> {
+            sendNotificationToLA(eventData, caseDetails, hearing.getValue());
+            sendNotificationToCafcass(eventData, caseDetails, hearing.getValue());
+            sendNotificationToRepresentatives(eventData, caseDetails, hearing.getValue());
```
akomisarek, line 62: Do we want too have 3 notification async? If LA fails others won't be delivered, are we happy with it?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/notify/hearing/NewNoticeOfHearingTemplate.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+import uk.gov.hmcts.reform.fpl.model.notify.NotifyData;
+
+import java.util.Map;
+
+@Getter
+@Setter
+@Builder
```
akomisarek, line 12: Do we need both setter and builder? It doesn't matter, I guess, but could we use Data?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/docmosis/NoticeOfHearingGenerationService.java
```diff
@@ -0,0 +1,42 @@
...
+        HearingVenue venue = hearingVenueLookUpService.getHearingVenue(hearingBooking);
+
+        return DocmosisNoticeOfHearing.builder()
+            .children(dataService.getChildrenDetails(caseData.getChildren1()))
+            .hearingDate(dataService.getHearingDateIfHearingsOnSameDay(hearingBooking).orElse(""))
+            .hearingTime(dataService.getHearingTime(hearingBooking))
+            .hearingType(hearingBooking.getType().getLabel().toLowerCase())
+            .hearingVenue(hearingVenueLookUpService.buildHearingVenue(venue))
+            .judgeAndLegalAdvisor(dataService.getJudgeAndLegalAdvisor(hearingBooking.getJudgeAndLegalAdvisor()))
+            .postingDate(formatLocalDateToString(LocalDate.now(), DATE))
```
akomisarek, line 36: Ideally, date like that should ALWAYS be injectable. As we capture the date only it's not big deal, but if it captured time we would struggle to test it properly.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
```diff
@@ -85,17 +105,9 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
...
-        List<Element<HearingBooking>> pastHearings = service.getPastHearings(hearingDetailsBefore);
-
-        List<Element<HearingBooking>> updatedHearings =
-            service.setHearingJudge(caseData.getHearingDetails(), caseData.getAllocatedJudge());
-
-        List<Element<HearingBooking>> combinedHearingDetails =
-            service.combineHearingDetails(updatedHearings, pastHearings);
-
-        caseDetails.getData().put(HEARING_DETAILS_KEY, combinedHearingDetails);
+        List<Element<HearingBooking>> hearingDetails = service.getHearingDetails(caseDetails, caseData, caseDataBefore);
```
joez-f, line 108: I agree that we shouldn't do too much in controller, but here we just put the whole controller logic into a service method, I don't think this is better

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/NewNoticeOfHearingEmailContentProvider.java
```diff
@@ -0,0 +1,51 @@
...
+            .hearingDate(formatLocalDateToString(hearingBooking.getStartDate().toLocalDate(), FormatStyle.LONG))
+            .hearingVenue(hearingVenueLookUpService.getHearingVenue(hearingBooking).getVenue())
+            .hearingTime(caseDataExtractionService.getHearingTime(hearingBooking))
+            .preHearingTime(caseDataExtractionService.extractPrehearingAttendance(hearingBooking))
+            .caseUrl(getCaseUrl(caseDetails.getId()))
+            .documentLink(linkToAttachedDocument(hearingBooking.getNoticeOfHearing()))
+            .familyManCaseNumber(defaultIfNull(caseData.getFamilyManCaseNumber(), ""))
+            .respondentLastName(getFirstRespondentLastName(caseData.getRespondents1()))
+            .digitalPreference(representativeServingPreferences == DIGITAL_SERVICE ? "Yes" : "No")
+            .caseUrl(representativeServingPreferences == DIGITAL_SERVICE ? getCaseUrl(caseDetails.getId()) : "")
```
akomisarek, line 48: Just asking - do we need to have this logic here? Couldn't the template mage the decision based on digital preference to hide the link?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
```diff
@@ -119,6 +151,99 @@ public HearingBooking getHearingBookingByUUID(List<Element<HearingBooking>> hear
...
+        List<Element<HearingBooking>> updatedHearings =
+            setHearingJudge(caseData.getHearingDetails(), caseData.getAllocatedJudge());
+
+        sendDocumentsForSelectedHearings(caseDetails, caseData, updatedHearings);
+
+        return combineHearingDetails(updatedHearings,
+            getPastHearings(defaultIfNull(caseDataBefore.getHearingDetails(), emptyList())));
+    }
+
+    private void sendDocumentsForSelectedHearings(
```
joez-f, line 224: this isn’t sending documents, it is attaching a notice of hearing document to a hearing booking, we could have a better name

KiranVarma02: I was going to ask you for a better name, what do you suggest please?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
KiranVarma02: Everything was in controller before, its much better for it to be in service. Ideally, i would want that some code from service delegated to other smaller units but as i was saying, we will not be able to do all that today. I was going to suggest we raise a tech debt and address it later. I can work on it today but don't want this to delay delivering business functionality. Happy to hear what others think!

KiranVarma02: if there are no other big review comments to address then yes we can do it today.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandlerTest.java
```diff
@@ -60,6 +75,12 @@
     @MockBean
     private RepresentativeService representativeService;
 
+    @MockBean
+    private DocmosisDocumentGeneratorService docmosisDocumentGeneratorService;
+
+    @MockBean
+    private UploadDocumentService uploadDocumentService;
+
```
akomisarek, line 83: Why we need to modify CMO event handler here?

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderReadyForPartyReviewEventHandlerTest.java
```diff
@@ -55,6 +68,12 @@
     @MockBean
     private CaseManagementOrderEmailContentProvider caseManagementOrderEmailContentProvider;
 
+    @MockBean
+    private DocmosisDocumentGeneratorService docmosisDocumentGeneratorService;
+
+    @MockBean
+    private UploadDocumentService uploadDocumentService;
+
```
akomisarek, line 76: Any reason why this is required?

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/GeneratedOrderEventHandlerTest.java
```diff
@@ -74,6 +89,12 @@
     @MockBean
     private InboxLookupService inboxLookupService;
 
+    @MockBean
+    private DocmosisDocumentGeneratorService docmosisDocumentGeneratorService;
+
+    @MockBean
+    private UploadDocumentService uploadDocumentService;
+
```
akomisarek, line 97: Same question.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/NoticeOfProceedingsServiceTest.java
```diff
@@ -27,14 +30,26 @@
...
 
     @Autowired
     private NoticeOfProceedingsService noticeOfProceedingService;
 
+    @MockBean
+    private DocmosisDocumentGeneratorService docmosisDocumentGeneratorService;
+
+    @MockBean
+    private UploadDocumentService uploadDocumentService;
+
```
akomisarek, line 52: I will stop commenting after this, but I would suggest a nicer approach instead of adding dependencies to the test which we don't use, seems we are doing something wrong here.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
```diff
@@ -119,6 +151,99 @@ public HearingBooking getHearingBookingByUUID(List<Element<HearingBooking>> hear
...
+        if (hearings.isEmpty() || hearingSelector == null) {
+            return List.of();
+        } else {
+            return hearingSelector.getSelected().stream()
+                .map(hearings::get)
+                .collect(toList());
+        }
+    }
+
+    public List<Element<HearingBooking>> getHearingDetails(
```
akomisarek, line 203: As commented in controller I don't think "getHearingDetails" gives this method credit. What hearing details, maybe comment is required with this logic.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
akomisarek, line 224: Not sure if I like "sendDocument" sending where? it uploads to docs store and genarates, I believe better name (or splitting into multiple,would be more beneficial)

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NewHearingsAddedEventHandlerTest.java
```diff
@@ -0,0 +1,131 @@
...
+    private LocalDateTime futureDate;
+
+    @BeforeEach
+    void setUp() {
+        futureDate = time.now().plusDays(1);
+    }
+
+    @Test
+    void shouldSendNotificationToLACafcassAndRepresentativesWhenNewHearingIsAdded() {
+        final ObjectMapper mapper = new ObjectMapper();
```
akomisarek, line 84: I would recommend using the same object mapper the code is using (via Injection), this can lead to differences how the code is serialised / deserialised.

service/src/test/java/uk/gov/hmcts/reform/fpl/handlers/NewHearingsAddedEventHandlerTest.java
```diff
@@ -0,0 +1,131 @@
...
+        List<Element<HearingBooking>> hearingBookings = List.of(
+            element(UUID.randomUUID(), createHearingBooking(futureDate.plusDays(5), futureDate.plusDays(6))));
+        NewNoticeOfHearingTemplate newNoticeOfHearingTemplate =
+            NewNoticeOfHearingTemplate.builder().build();
+
+        given(hearingBookingService.getSelectedHearings(any(), any())).willReturn(hearingBookings);
+        given(inboxLookupService.getNotificationRecipientEmail(caseDetails, LOCAL_AUTHORITY_CODE))
+            .willReturn(LOCAL_AUTHORITY_EMAIL_ADDRESS);
+        given(newNoticeOfHearingEmailContentProvider.buildNewNoticeOfHearingNotification(
+            any(CaseDetails.class), any(HearingBooking.class), any()))
```
akomisarek, line 98: On unit test level we should try to avoid using "any()" we now what class will be passed here we should validate. 

I can pass the wrong class in the code and unit test won't pass.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/HearingBookingServiceTest.java
```diff
@@ -281,13 +373,19 @@ private HearingBooking buildHearingBooking(YesNo useAllocatedJudge) {
             .build();
     }
 
-    private List<Element<HearingBooking>> createHearingBookings() {
-        return List.of(
+    private List<Element<HearingBooking>> createHearingBookings(boolean addHearing) {
```
akomisarek, line 376: I don't like the approach of this unit test, passing boolean into the method, seems invisible what this method is doing. I would much more prefer to have tests adding extra one hearing if required.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
KiranVarma02: agree that name is not great. I was trying to refactor out all this code from controller without knowing much about the code.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NewHearingsAddedEventHandler.java
KiranVarma02: agree to the point. Looking in to it.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NewHearingsAddedEventHandler.java
KiranVarma02: good point. I will follow how we notify in other places and separate notifys.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/NewNoticeOfHearingEmailContentProvider.java
joez-f: gov notify cannot hide dynamic content based on condition, only static ☹️ so I think we must have the logic here, i.e we can use gov notify to hide some text, but for caseUrl we need to set it to null so that it does not render

ccd-definition/AuthorisationCaseField/gatekeeper.json
```diff
@@ -237,6 +237,20 @@
...
+    "UserRole": "caseworker-publiclaw-gatekeeper",
+    "CRUD": "CRUD"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "newHearingSelector",
+    "UserRole": "caseworker-publiclaw-gatekeeper",
+    "CRUD": "CRUD"
+  },
```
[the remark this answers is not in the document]
joez-f: Judge only has R permission to the event

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerAboutToSubmitTest.java
```diff
@@ -138,6 +164,47 @@ void shouldOnlyAddCurrentHearingToListWhenSameIdForHearings() {
...
+                    .data(data)
+                    .build())
+                .caseDetailsBefore(CaseDetails.builder()
+                    .data(Map.of("data", "some data")).build()).build();
+
+            AboutToStartOrSubmitCallbackResponse response = postAboutToSubmitEvent(callbackRequest);
+            CaseData responseData = mapper.convertValue(response.getData(), CaseData.class);
+            assertThat(responseData.getHearingDetails().get(0).getValue().getNoticeOfHearing()).isNull();
+            assertThat(responseData.getHearingDetails().get(1).getValue().getNoticeOfHearing()).isNotNull();
+            assertThat(responseData.getHearingDetails().get(2).getValue().getNoticeOfHearing()).isNull();
```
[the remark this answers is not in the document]
joez-f: yes, good point I can improve that

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsController.java
joez-f: have renamed to NewHearingsAdded and surrounded with if statement

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerAboutToSubmitTest.java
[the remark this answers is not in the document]
joez-f: have pushed up improvement, let me know if anything more is needed

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/HearingBookingDetailsControllerMidEventTest.java
```diff
@@ -93,6 +98,47 @@ void shouldReturnAnErrorWhenExistingBookingIsUpdatedToPastDate() {
...
+
+        assertThat(callbackResponse.getData().get(NEW_HEARING_SELECTOR.getKey())).isEqualTo(expectedSerializedSelector);
+    }
+
+    @Test
+    void shouldResetNewHearingsLabelAndSelectorWhenNoNewHearingsHaveBeenAdded() {
+        LocalDateTime date = now().plusDays(5);
+        UUID hearingId = randomUUID();
+
+        List<Element<HearingBooking>> newHearingBooking = listBookingWithStartDate(hearingId, date);
```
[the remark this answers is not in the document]
joez-f: done 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/HearingBookingService.java
KiranVarma02: updated name

service/src/test/java/uk/gov/hmcts/reform/fpl/service/HearingBookingServiceTest.java
KiranVarma02: changed as suggested.

e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
```diff
@@ -69,12 +69,17 @@ Scenario('Gatekeeper enters hearing details and submits', async (I, caseViewPage
   await addHearingBookingDetailsEventPage.enterHearingDetails(hearingDetails[1]);
   await addHearingBookingDetailsEventPage.enterJudge(hearingDetails[1].judgeAndLegalAdvisor);
   await addHearingBookingDetailsEventPage.enterLegalAdvisor(hearingDetails[1].judgeAndLegalAdvisor.legalAdvisorName);
-  await I.completeEvent('Save and continue', {summary: 'summary', description: 'description'});
+  await I.retryUntilExists(() => I.click('Continue'), '#newHearingSelector_newHearingSelector');
+  addHearingBookingDetailsEventPage.sendNoticeOfHearing(hearingDetails[0].sendNoticeOfHearing);
+  addHearingBookingDetailsEventPage.sendNoticeOfHearing(hearingDetails[1].sendNoticeOfHearing, 1);
+  await I.completeEvent('Save and continue');
```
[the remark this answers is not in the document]
Jamesnelson117: Yeah I agree there's no need to test this flow twice. The e2e test was already there, I had only updated it to make sure it continues to work. Currently we need this event because the mock data does not include hearingDetails data. 

Perhaps we could update this file so fixture does include hearings and test just ensures user can read from tabs.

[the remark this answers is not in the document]
Jamesnelson117: Caught up with Adam and agreed to keep this in for now.

PR #1394, opened 2020-07-01
Support login retries in e2e tests

tomaszpowroznik:


e2e/steps_file.js
```diff
@@ -208,7 +203,11 @@ module.exports = function () {
           output.log(`retryUntilExists(${locator}): element found before try #${tryNumber} was executed`);
           break;
         }
-        await action();
+        try {
+          await action();
+        }catch(error){
```
joez-f, line 208: really minor but a space before catch would look nicer

PR #1401, opened 2020-07-02
chore: updated Java chart version to 3.0.0 (FPLA-1949)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1949

### Change description ###
Updates java chart version to 3.0.0

```
[ ] Yes
[x] No
```


charts/fpl-case-service/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: fpl-case-service
 apiVersion: v1
```
joez-f, line 2: In the Spring boot 2.3.0 upgrade message it says:
You will need to move to apiVersion: v2  in Chart.yaml for the helm chart as well, see the below sample PRs.

Do we need to do this now too? @tomaszpowroznik maybe you can advise

charts/fpl-case-service/Chart.yaml
```diff
@@ -1,7 +1,7 @@
...
 home: https://github.com/hmcts/fpl-ccd-configuration
-version: 1.11.7
+version: 1.11.8
```
joez-f, line 4: should we move to version 1.12 instead? I am not sure, probably doesn't matter

tomaszpowroznik: probably good idea to bump up minor version and not patch
please also update flux to point to new version

charts/fpl-case-service/Chart.yaml
tomaszpowroznik: I think we can change to v2, but it is not really related to upgrading java chart.
Helm doc says:
The apiVersion field should be v2 for Helm charts that require at least Helm 3. Charts supporting previous Helm versions have an apiVersion set to v1 and are still installable by Helm 3.

Looks like not mandatory but better to change

PR #1407, opened 2020-07-03
feat:Fees Register API keyword change to GeneralAppWithoutNotice

nkoranne:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/service/src/main/resources/application.yaml
+++ b/service/src/main/resources/application.yaml
@@ -32,7 +32,7 @@ fees-register:
       event: general application
       jurisdiction1: family
       jurisdiction2: family court
-      keyword: ${FEES_REGISTER_C2_WITHOUT_NOTICE_KEYWORD:without-notice}
+      keyword: ${FEES_REGISTER_C2_WITHOUT_NOTICE_KEYWORD:GeneralAppWithoutNotice}
       service: other
     C2_WITH_NOTICE:
       channel: default
--- a/service/src/main/resources/application.yaml
+++ b/service/src/main/resources/application.yaml
@@ -32,7 +32,7 @@ fees-register:
       event: general application
       jurisdiction1: family
       jurisdiction2: family court
-      keyword: without-notice
+      keyword: ${FEES_REGISTER_C2_WITHOUT_NOTICE_KEYWORD:without-notice}
       service: other
     C2_WITH_NOTICE:
       channel: default


joez-f:
there are some changes to fpla-docker in this PR, are they intentional?

nkoranne:
There is a wire mock change in c2-without-notice.json as well, not sure why that file not showing in this PR :(

tomaszpowroznik:
> There is a wire mock change in c2-without-notice.json as well, not sure why that file not showing in this PR :(
this change is not correct, it reverts some of changes done in fpl docker recently

tomaszpowroznik:
no changes in fpla-docker is needed for this.

service/src/main/resources/application.yaml
```diff
@@ -32,7 +32,7 @@ fees-register:
       event: general application
       jurisdiction1: family
       jurisdiction2: family court
-      keyword: without-notice
+      keyword: ${FEES_REGISTER_C2_WITHOUT_NOTICE_KEYWORD:GeneralAppWithoutNotice}
```
lewisbirks, line 35: Don't think that the whole `${FEES_REGISTER_C2_WITHOUT_NOTICE_KEYWORD:GeneralAppWithoutNotice}` is needed, just `GeneralAppWithoutNotice`.
From my understanding as the env variable should map directly to the property spring should recognise this and give the env variable priority over the `application.yaml` value.

lewisbirks: Ahh it would need to be called `FEES_REGISTER_PARAMETERS_C2_WITHOUT_NOTICE_KEYWORD`

tomaszpowroznik: @lewisbirks are you sure about this? Code does not directly goes this this property

lewisbirks: Chatted with tomasz and we can leave as is, just more work for a temp thing than is needed.

tomaszpowroznik: i would prefer to keep explicit env variable, if somebody decide to refactor name in application xml then  env variable wont match and prod config can be broken without even noticing it. 
Thats nice feature, but i guess quite risky if you dont now context

akomisarek: @nkoranne  As a safer approach we could leave old default and ovewrite with the new value in flux demo. 

But that would work if the overwriting is confirmed and we have got flux ready for prod to use the old value. (which can be merged beforehand)

PR #1423, opened 2020-07-14
hack: fix OWASP configuration (temp solution to get builds passing)

akomisarek:
N/A

### Change description ###

Fix config for owasp. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -17,10 +17,6 @@ apply plugin: 'org.springframework.boot'
 apply plugin: 'io.spring.dependency-management'
 apply plugin: 'info.solidsoft.pitest'
 
-subprojects {
-  apply plugin: 'org.owasp.dependencycheck'
-}
-
 // https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html
 dependencyCheck {
   // Specifies if the build should be failed if a CVSS score above a specified level is identified.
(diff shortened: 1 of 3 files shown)


service/build.gradle
```diff
@@ -16,7 +16,24 @@ plugins {
 apply plugin: 'org.springframework.boot'
 apply plugin: 'io.spring.dependency-management'
 apply plugin: 'info.solidsoft.pitest'
-apply plugin: 'org.owasp.dependencycheck'
+
+subprojects {
```
timja, line 20: isn't this already in allprojects, why is this needed again?

akomisarek: I was playing with the config, I removed that now. Seems to work locally as well.

service/build.gradle
```diff
@@ -16,7 +16,20 @@ plugins {
 apply plugin: 'org.springframework.boot'
 apply plugin: 'io.spring.dependency-management'
 apply plugin: 'info.solidsoft.pitest'
-apply plugin: 'org.owasp.dependencycheck'
+
+// https://jeremylong.github.io/DependencyCheck/dependency-check-gradle/configuration.html
+dependencyCheck {
+  // Specifies if the build should be failed if a CVSS score above a specified level is identified.
```
banderous, line 22: this comment is obsolete

akomisarek: Thanks, removed and cleaned up a bit so it shows less changes.

PR #1425, opened 2020-07-15
impr: improve Dependency config

akomisarek:

N/A

### Change description ###

Remove dependency duplicate config from subproject

```
[ ] Yes
[X] No
```


build.gradle
```diff
@@ -6,13 +6,15 @@ buildscript {
 
 plugins {
   id 'java'
-  id 'org.owasp.dependencycheck' version '5.3.2.1' apply true
+  id 'org.owasp.dependencycheck' version '5.3.2.1'
 }
 
+// Enables code navigation from IntelliJ
+apply plugin: 'org.owasp.dependencycheck'
```
banderous, line 13: not needed, it'll be automatically applied if you declare it in the plugins block

akomisarek: Thanks, removed. For some reason IntelliJ code navigation didn't work for me initially, but now it even works when I have got them inlined 🤔

PR #1428, opened 2020-07-16
refactor: consolidate CMO issued notification (FPLA-1816)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-1816

### Change description ###
Replaces the following GOV notify templates -
**LA + digital rep** (people with case access): https://www.notifications.service.gov.uk/services/12f756df-f01d-4a32-a405-e1ea8a494fbb/templates/64cc7ea5-ca74-4679-912f-6d84d1c7e4cd
**Email rep**:(people without case access) https://www.notifications.service.gov.uk/services/12f756df-f01d-4a32-a405-e1ea8a494fbb/templates/931dcd86-d331-4273-abaa-208daed0761a

with 
**CMO issued template** - https://www.notifications.service.gov.uk/services/12f756df-f01d-4a32-a405-e1ea8a494fbb/templates/1587187b-7c5a-4665-880c-3d52073124bc

CMO issued now uses Java objects for template data.
Int and unit tests updated.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
```diff
@@ -1,24 +1,92 @@
...
+                eventData.getCaseDetails(), EMAIL);
+
+        final String cafcassEmail = cafcassLookupConfiguration.getCafcass(eventData.getLocalAuthorityCode()).getEmail();
+
+        notificationService.sendEmail(CMO_ORDER_ISSUED_NOTIFICATION_TEMPLATE, cafcassEmail, cafcassParameters,
+            eventData.getReference());
+    }
+
+    private void sendToRepresentatives(final EventData eventData, RepresentativeServingPreferences servingPreference) {
+        CaseData caseData = mapper.convertValue(eventData.getCaseDetails().getData(), CaseData.class);
```
joez-f, line 77: I think we should call sendToRepresentativesByServedPreference from RepresentativeNotificationService in here instead as it is doing the same thing; you need to pass in parameters but the parameters don't change depending on the representative so you can generate them and then just pass them in

Jamesnelson117: Agree with this change but `sendToRepresentativesByServedPreference` takes `final Map<String, Object> templateParameters` which when using java object for templates would require me to call `toMap`on my template object before the `NotificationService` tries to call toMap again. 

It would require a refactor of `RepresentativeNotificationService.sendToRepresentativesByServedPreference` to accept an overload to handle generic `notifyData`. Out of scope for this ticket but happy to make a tech deb and clean this and others up.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
```diff
@@ -1,24 +1,92 @@
...
-        caseLinkNotificationHandler.sendNotifications(eventData);
-        documentLinkNotificationHandler.sendNotifications(eventData,
-            event.getDocumentContents());
+        sendToLocalAuthority(eventData);
+        sendToCafcass(eventData);
+        sendToRepresentatives(eventData, DIGITAL_SERVICE);
+        sendToRepresentatives(eventData, EMAIL);
+
+        // TODO
+        // Refactor issuedOrderAdminNotificationHandler to use Java object for template so we can avoid passing byte
```
lewisbirks, line 49: Should this be a tech debt ticket?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
```diff
@@ -1,24 +1,92 @@
...
+        notificationService.sendEmail(CMO_ORDER_ISSUED_NOTIFICATION_TEMPLATE, cafcassEmail, cafcassParameters,
+            eventData.getReference());
+    }
+
+    private void sendToRepresentatives(final EventData eventData, RepresentativeServingPreferences servingPreference) {
+        CaseData caseData = mapper.convertValue(eventData.getCaseDetails().getData(), CaseData.class);
+        List<Representative> representatives = representativeService.getRepresentativesByServedPreference(
+            caseData.getRepresentatives(), servingPreference);
+
+        representatives.stream()
```
lewisbirks, line 81: If there is an issue with sending one of the emails this would block all emails. Is that what we want? Do we want it to be done in parallel? Would that be out of the scope of the ticket?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
```diff
@@ -1,24 +1,92 @@
...
+    private final ObjectMapper mapper;
+    private final InboxLookupService inboxLookupService;
+    private final NotificationService notificationService;
+    private final RepresentativeService representativeService;
+    private final CaseManagementOrderEmailContentProvider caseManagementOrderEmailContentProvider;
+    private final CafcassLookupConfiguration cafcassLookupConfiguration;
+    private final IssuedOrderAdminNotificationHandler issuedOrderAdminNotificationHandler;
 
     @EventListener
     public void sendEmailsForIssuedCaseManagementOrder(final CaseManagementOrderIssuedEvent event) {
```
lewisbirks, line 40: Should we make this method async like we are starting to do with the rest of the notification stuff?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/content/CaseManagementOrderEmailContentProviderTest.java
```diff
@@ -35,6 +39,8 @@
     @Autowired
     private CaseManagementOrderEmailContentProvider caseManagementOrderEmailContentProvider;
 
+    private static final byte[] APPLICATION_BINARY = TestDataHelper.DOCUMENT_CONTENT;
```
lewisbirks, line 42: Do we need this variable? Could we not just directly use `TestDataHelper.DOCUMENT_CONTENT` instead? The name seems slightly more fitting as well for its uses.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
Jamesnelson117: Yes I was planning too. I will remove comment and create tech debt.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
Jamesnelson117: Agreed to make tech debt.

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
Jamesnelson117: Agreed to make part of tech debt.

PR #1430, opened 2020-07-17
fix: fixed test names code comments (FPLA-1622)

nkoranne:

[FPLA-1622](https://tools.hmcts.net/jira/browse/FPLA-1622)

### Change description ###

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -69,10 +69,11 @@ public String getHearingTime(HearingBooking hearingBooking) {
         return hearingTime;
     }
 
-    public Optional<String> getHearingDateIfHearingsOnSameDay(HearingBooking hearingBooking) {
+    public Optional<String> getHearingDate(HearingBooking hearingBooking) {
```
toireasam, line 72: You need to update the classes that use this service. NoticeOfHearingGenerationService.java for example is now breaking as it's trying to reference getHearingDateIfHearingsOnSameDay

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
```diff
@@ -69,10 +69,11 @@ public String getHearingTime(HearingBooking hearingBooking) {
...
         String hearingDate = null;
 
-        // If they aren't on the same date return nothing
+        // If they are on same day, then return formatted date
+        // and If they aren't on the same day, then return nothing
```
toireasam, line 76: Do we need this comment or is the code self explanatory? I will leave this to your discretion :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
nkoranne: yeah, sorry ... before check-in this file was not changed.. I need to merge it before any commits, will keep that in mind

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
nkoranne: This was one the point in tech debt , that's why added it :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
toireasam: np :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
toireasam: cool lets leave then :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
akomisarek: Is this change related to tech debt? If yes please link to the commit / ticket name (so we know in history why the change was introduced)

Alternatively do we agree with the change? I don't know why this is being renamed, the name was nicely telling what this method is doing, now the name is worse...

akomisarek:
I read now the TechDebt - not sure if I agree, @toireasam / @lewisbirks what do you think? Is the getHearingDate better?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseDataExtractionService.java
nkoranne: - The branch name itself says it's a tech debt, so I feel that's good enough indication 
- There is another method getHearingTime, that also has the same logic , but the method name is still getHearingTime
so from consistency point of view , method names are fine ...  if we need to change, then we need to change both the method names 
  
thoughts?

lewisbirks:
> I read now the TechDebt - not sure if I agree, @toireasam / @lewisbirks what do you think? Is the getHearingDate better?

I did like the more verbose name as it explained the reason for the optional return value

toireasam:
@akomisarek @lewisbirks I also now think I prefer the old name too. Explains better what is happening. So I think lets revert to old method name in service. and we can leave the test names changes as they are now as they incorrectly where named before. sound good? :) @nkoranne

nkoranne:
> @akomisarek @lewisbirks I also now think I prefer the old name too. Explains better what is happening. So I think lets revert to old method name in service. and we can leave the test names changes as they are now as they incorrectly where named before. sound good? :) @nkoranne

yeah ... that's fine @toireasam .. I will make changes

akomisarek:
Great, thanks everyone :), sometimes we need to double-check if the tickets make sense, but yeah test method names should be correct, so already a good improvement.

PR #1445, opened 2020-07-22
impr: cleanup events to use DocumentReference

joez-f:
### Change description ###

Pass DocumentReference to events in submitted callbacks instead of passing byte[] documentContents. This allows event handlers/content providers to better utilise linkToAttachedDocument in AbstractEmailContentProvider (less duplication of code invoking documentDownloadService). This also fixes an issue with document links of generated orders being links directly to the document instead of a gov notify download of the attachment.

```
[ ] Yes
[X] No
```


joez-f:
Parking ticket as advised by @akomisarek, needs a bit more discussion next week/week after about how to best implement the improvement @tomaszpowroznik

PR #1454, opened 2020-07-28
impr: default "share a case" to manage org (FPLA-1980)

lewisbirks:

[FPLA-1980](https://tools.hmcts.net/jira/browse/FPLA-1980)

### Change description ###

Switched the ordering of the calls to manage org and our local mapping for organisations.
Updated relevant tests.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -32,16 +32,14 @@
     private final RequestData requestData;
 
     public Set<String> findUserIdsInSameOrganisation(String localAuthorityCode) {
-
         try {
-            return Set.copyOf(getUsersFromSameOrganisationBasedOnAppConfig(localAuthorityCode));
-        } catch (UnknownLocalAuthorityCodeException ex) {
+            return Set.copyOf(getUsersFromSameOrganisationBasedOnReferenceData(requestData.authorisation()));
+        } catch (FeignException e) {
```
tomaszpowroznik, line 37: exception is not even logged, we should be able to distinguish between situation where theres is failure because org is not configured and between failures like Service unavailable.

Anyway I would add more logs which helps us tell, which path was used

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrganisationServiceTest.java
```diff
@@ -57,21 +61,20 @@
 
     @BeforeEach
     void setup() {
-        LocalAuthorityUserLookupConfiguration laUserLookupConfig =
-            new LocalAuthorityUserLookupConfiguration("SA=>1,2,3");
-        organisationService = new OrganisationService(laUserLookupConfig, organisationApi,
-            authTokenGenerator, requestData);
+        lookupSpy = Mockito.spy(new LocalAuthorityUserLookupConfiguration("SA=>1,2,3"));
```
tomaszpowroznik, line 64: you can try:

@ExtendWith(MockitoExtension.class)
@MockitoSettings(strictness = LENIENT)
class OrganisationServiceTest {

    @Mock
    private OrganisationApi organisationApi;

    @Mock
    private AuthTokenGenerator authTokenGenerator;

    @Mock
    private RequestData requestData;

    @Spy
    private LocalAuthorityUserLookupConfiguration lookupSpy = new LocalAuthorityUserLookupConfiguration("SA=>1,2,3");

    @InjectMocks
    private OrganisationService organisationService;

    private static final Request REQUEST = Request.create(GET, EMPTY, Map.of(), new byte[]{}, UTF_8, null);
    private static final String AUTH_TOKEN_ID = "Bearer authorisedBearer";
    private static final String SERVICE_AUTH_TOKEN_ID = "Bearer authorised service";
    private static final String USER_EMAIL = "test@test.com";
    private static final Organisation EMPTY_ORGANISATION = Organisation.builder().build();
    private static final Organisation POPULATED_ORGANISATION = buildOrganisation();

    @BeforeEach
    void setup() {
        when(authTokenGenerator.generate()).thenReturn(SERVICE_AUTH_TOKEN_ID);
        when(requestData.authorisation()).thenReturn(AUTH_TOKEN_ID);
    }

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrganisationServiceTest.java
```diff
@@ -80,21 +83,28 @@ void shouldReturnUsersFromOrganisationIfExistsInRefData() {
...
 
-        Set<String> userIds = organisationService
-            .findUserIdsInSameOrganisation("AN");
+        Set<String> userIds = organisationService.findUserIdsInSameOrganisation("AN");
+
+        assertThat(userIds).containsExactlyInAnyOrder("40", "41");
+    }
+
+    @Test
+    void shouldNotCheckLocalMappingWhenOrganisationExistsInRefData() {
```
tomaszpowroznik, line 92: this test is not needed, as it does not verify service functionality,  it is enough to add
verifyNoInteractions(lookupSpy);
in shouldReturnUsersFromOrganisationIfExistsInRefData

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrganisationServiceTest.java
```diff
@@ -80,21 +83,28 @@ void shouldReturnUsersFromOrganisationIfExistsInRefData() {
...
-        assertThat(userIds)
-            .containsExactlyInAnyOrder("40", "41");
+        verify(lookupSpy, never()).getUserIds(any());
     }
 
     @Test
     void shouldReturnEmptyListWhenTheLAIsNotKnownAndTheApiReturnsNotFound() {
         when(organisationApi.findUsersByOrganisation(any(), any(), any(), any()))
-            .thenThrow(new FeignException.NotFound("No organisation", REQUEST, new byte[]{}));
+            .thenThrow(new FeignException.NotFound("No organisation", REQUEST, new byte[] {}));
```
tomaszpowroznik, line 105: i think it will be Forbidden. PRD wont return notfound as if organisation exists and user belongs to it there must be at lease one user, the one who executed endpoint

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OrganisationServiceTest.java
lewisbirks: Wonder what I did differently, last time I tried to do this the spy wasn't created properly and was trying therefore failing initialisation. 🤷 
That works so will change to that.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OrganisationService.java
```diff
@@ -32,18 +32,23 @@
     private final RequestData requestData;
 
     public Set<String> findUserIdsInSameOrganisation(String localAuthorityCode) {
+        try {
+            return Set.copyOf(getUsersFromSameOrganisationBasedOnReferenceData(requestData.authorisation()));
+        } catch (FeignException.NotFound | FeignException.Forbidden unregisteredException) {
+            log.warn("User not registered in any org in MO", unregisteredException);
+        } catch (FeignException prdFailureException) {
+            log.error("PRD failed", prdFailureException);
```
tomaszpowroznik, line 40: Message could be a bit better, Maybe 'Request for user in same organisation failed'

lewisbirks: Oops forgot to change that, will do now

PR #1456, opened 2020-07-28
bug: confidential others loosing data after edit(FPLA-2035)

KiranVarma02:

https://tools.hmcts.net/jira/browse/FPLA-2035

### Change description ###
Fix to confidential others loosing data after edit

There is also another bug in master when we add 2 other as confidential then we only show 1 other 2 times which is wrong. 

```
[ ] Yes
[ *] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/OthersControllerAboutToStartTest.java
```diff
@@ -33,7 +33,7 @@
     private static final UUID ADDITIONAL_OTHER_ID = randomUUID();
 
     @Test
-    void shouldReturnOthersWhenTheyAreMarkedAsConfidential() {
+    void shouldReturnOthersWithoutConfidentialDataWhenTheyAreMarkedAsConfidential() {
```
KiranVarma02, line 36: I think this test can be removed. Let me know what you think?

Jamesnelson117: I agree it's similar to the unit test but I don't think it should be removed. This test tests how caseDetails should behave after mutation so I think there's value in it. 

But is this test not wrong? Maybe i'm understanding it wrong but is the idea not if an other has 'detailsHidden' yes then it should have contact details hidden from the object? Using `otherWithDetailsRemoved` does not really test that?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsServiceTest.java
```diff
@@ -461,7 +461,7 @@ void shouldReturnOthersIfOthersIsPrePopulated() {
         }
 
         @Test
-        void shouldPrepareOtherWithConfidentialValuesWhenConfidentialOthersIsNotEmpty() {
+        void shouldPrepareOtherWithConfidentialValuesRemoved() {
```
Jamesnelson117, line 464: This only tests the one 'other' can extend or have a new test to ensure multiple others have confidential fields persisted.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/OthersControllerAboutToStartTest.java
Jamesnelson117: No I think there's value in keeping this test still as it test's what CaseDetails after mutation. But is this test right? The object you pass to firstOther does not have confidentialFields to hide so were not really testing anything no? Maybe i'm misunderstanding the logic.

KiranVarma02: i have made slight changes to possibly make it right :man_shrugging but i still don't see point of testing confidential details in controller when same test is done unit tests.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsServiceTest.java
KiranVarma02: done

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/OthersControllerAboutToStartTest.java
Jamesnelson117: I still don't understand what we are testing here. We aren't testing as your name applies that confidential fields are removed since. 
`assertThat(caseData.getOthers().getFirstOther()).isEqualTo(other());` asserts that other() still contains address etc.

toireasam: I think there's benefit in keeping this test although yes agreed it's not testing what name says. it should work like this:

1) case data should have an other added with confidential details removed

2) assert should check that the confidential details has been added again after call of about to start

essentially - the old test

PR #1470, opened 2020-08-03
bug: Confidential others loosing data after edit

KiranVarma02:
https://tools.hmcts.net/jira/browse/FPLA-2035

### Change description ###
Confidential others loosing data after edit

```
[ ] Yes
[* ] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
@@ -95,13 +95,7 @@ public class Other implements Representable, ConfidentialParty<Other> {
 
     @Override
     public Other addConfidentialDetails(Party party) {
-        Telephone telephoneNumber = ofNullable(party.getTelephoneNumber()).orElse(Telephone.builder().build());
-
-        return this.toBuilder()
-            .name(party.getFirstName())
-            .address(party.getAddress())
-            .telephone(ofNullable(telephoneNumber.getTelephoneNumber()).orElse(""))
-            .build();
+        return null;
     }
 
     @Override
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
@@ -18,7 +18,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
-import static java.util.Optional.ofNullable;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.unwrapElements;
 
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
@@ -12,6 +12,7 @@ import java.util.List;
 import static java.util.stream.Collectors.toList;
 import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
 import static uk.gov.hmcts.reform.fpl.enums.ConfidentialPartyType.OTHER;
+import static uk.gov.hmcts.reform.fpl.utils.ConfidentialDetailsHelper.getItemToAdd;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
 
 @Service
@@ -78,13 +79,5 @@ public class ConfidentialDetailsService {
             });
         }
         return collection;
-    }
-
-    private <T> T getItemToAdd(List<Element<T>> confidential, Element<T> element) {
-        return confidential.stream()
-            .filter(item -> item.getId().equals(element.getId()))
-            .map(Element::getValue)
-            .findFirst()
-            .orElse(element.getValue());
     }
 }
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
@@ -12,7 +12,7 @@ import java.util.List;
 import static java.util.stream.Collectors.toList;
 import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
 import static uk.gov.hmcts.reform.fpl.enums.ConfidentialPartyType.OTHER;
-import static uk.gov.hmcts.reform.fpl.utils.ConfidentialDetailsHelper.getItemToAdd;
+import static uk.gov.hmcts.reform.fpl.utils.ConfidentialDetailsHelper.addConfidentialDetailsForItems;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
 
 @Service
@@ -69,7 +69,7 @@ public class ConfidentialDetailsService {
             all.forEach(element -> {
                 T party = element.getValue();
                 if (party.containsConfidentialDetails()) {
-                    T partyToAdd = getItemToAdd(confidential, element);
+                    T partyToAdd = addConfidentialDetailsForItems(confidential, element);
                     T confidentialParty = party.addConfidentialDetails((partyToAdd.toParty()));
 
                     collection.add(element(element.getId(), confidentialParty));
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
@@ -12,7 +12,7 @@ import java.util.List;
 import static java.util.stream.Collectors.toList;
 import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
 import static uk.gov.hmcts.reform.fpl.enums.ConfidentialPartyType.OTHER;
-import static uk.gov.hmcts.reform.fpl.utils.ConfidentialDetailsHelper.addConfidentialDetailsForItems;
+import static uk.gov.hmcts.reform.fpl.utils.ConfidentialDetailsHelper.addConfidentialDetailsForElement;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
 
 @Service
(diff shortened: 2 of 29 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/OthersService.java
```diff
@@ -31,6 +40,61 @@ public String buildOthersLabel(Others others) {
...
+            }
+        });
+
+        return Others.builder()
+            .firstOther(getFirstOther(caseData.getConfidentialOthers(), others))
+            .additionalOthers(getAdditionalOthers(others))
+            .build();
+    }
+
+    private Element<Other> findConfidentialOther(List<Element<Other>> confidentialOthers, Element<Other> element) {
```
toireasam, line 62: We can use the similar method in confidential details service called getItemToAdd

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OthersService.java
```diff
@@ -31,6 +40,61 @@ public String buildOthersLabel(Others others) {
...
+        return confidentialOthers.stream()
+            .filter(confidentialOther -> confidentialOther.getId().equals(element.getId()))
+            .findFirst()
+            .orElse(element);
+    }
+
+    private Other addConfidentialDetails(Element<Other> confidentialOther, Element<Other> other) {
```
toireasam, line 69: We can remove this method and and use addConfidentialDetails method from model class

KiranVarma02: this method in model class does not get correct telephone and address like here...

service/src/main/java/uk/gov/hmcts/reform/fpl/service/OthersService.java
KiranVarma02: moved it to Helper class

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
```diff
@@ -74,14 +69,9 @@
             all.forEach(element -> {
                 T party = element.getValue();
                 if (party.containsConfidentialDetails()) {
-                    T partyToAdd = getItemToAdd(confidential, element);
```
toireasam, line 77: Why have we renamed this method? It isn't adding confidentialDetails, it is getting the confidential item to add

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
```diff
@@ -95,13 +94,7 @@ public Other extractConfidentialDetails() {
...
     @Override
     public Other addConfidentialDetails(Party party) {
-        Telephone telephoneNumber = ofNullable(party.getTelephoneNumber()).orElse(Telephone.builder().build());
-
-        return this.toBuilder()
-            .name(party.getFirstName())
-            .address(party.getAddress())
-            .telephone(ofNullable(telephoneNumber.getTelephoneNumber()).orElse(""))
-            .build();
+        return null;
```
toireasam, line 97: This method shouldn't return null? We could update this method to the new way we are using it here perhaps, this one..

    private Other addConfidentialDetails(Other confidentialOther, Element<Other> other) {
        return other.getValue().toBuilder()
            .telephone(confidentialOther.getTelephone())
            .address(confidentialOther.getAddress())
            .build();
    }

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/ConfidentialDetailsHelperTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+class ConfidentialDetailsHelperTest {
+
+    private static final String CONFIDENTIAL = "Yes";
+    private static final UUID ID = randomUUID();
+    private static final String TELEPHONE_NUMBER = "01227 831393";
+
+    @Test
+    void shouldAddConfidentialDetailsForOther() {
+        List<Element<Other>> others = List.of(otherWithConfidentialFields(ID, CONFIDENTIAL));
```
toireasam, line 24: This isn't testing the correct thing here. It should test that the matching element passed in is found in the confidential list. 

Something like this 

'    @Test
    void shouldFindItemToAddWhenInConfidential() {
         UUID firstUUID = fromString("d733b442-1b70-4abb-87b6-47e406d2e32a");
         UUID secondUUID = fromString("d733b442-1b70-4abb-87b6-47e406d2e32b");

        List<Element<Other>> confidential = List.of(otherWithConfidentialFields(firstUUID, CONFIDENTIAL, TELEPHONE_NUMBER),
            otherWithConfidentialFields(secondUUID, CONFIDENTIAL, "02830860107"));
        Element<Other> element = otherWithRemovedConfidentialFields(firstUUID);

        Other confidentialOther = addConfidentialDetailsForElement(confidential, element);

        assertThat(confidentialOther).isEqualToComparingFieldByField(confidential.get(0));
    }'

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsService.java
KiranVarma02: yes, you are right. I will call it getConfidentialItemToAdd..

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/ConfidentialDetailsHelperTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    void shouldNotAddConfidentialDetailsForNonMatchingElement() {
+        List<Element<Other>> others = List.of(otherWithConfidentialFields(ID, CONFIDENTIAL));
+        Element<Other> othersNotConfidential = otherWithRemovedConfidentialFields(randomUUID());
+
+        Other confidentialOthers = addConfidentialDetailsForElement(others, othersNotConfidential);
+
+        assertThat(confidentialOthers).isEqualTo(otherWithRemovedConfidentialFields(ID).getValue());
```
toireasam, line 39: Again it's not adding any confidential details here, if it doesn't find a match in confidential it returns the value of the element you passed in

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
KiranVarma02: we can't update method because addConfidentialDetails is implementation of interface. I either have to add new method to interface and implement dummy in Child and Respondent or do it this way..

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/ConfidentialDetailsHelperTest.java
KiranVarma02: Changed test name and modified test to test fieldByField instead of 1 field.

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/ConfidentialDetailsHelperTest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private Element<Other> otherWithRemovedConfidentialFields(UUID id) {
+        return element(id, baseOtherBuilder(CONFIDENTIAL).build());
+    }
+
+    private Element<Other> otherWithConfidentialFields(UUID id, String detailsHidden) {
+        return element(id, baseOtherBuilder(detailsHidden)
+            .address(Address.builder().addressLine1("Address Line 1").build())
+            .telephone(TELEPHONE_NUMBER)
```
toireasam, line 59: We don't have any other fields like address or name as static fields so i'd either add this telephone hard coded or i'd make them all static constants

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
toireasam: OK sure then maybe this is only option :)

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/ConfidentialDetailsHelperTest.java
KiranVarma02: you are right.. i used static TELEPHONE_NUMBER somewhere else before but now i don't. I will remove that.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OthersServiceTest.java
```diff
@@ -77,4 +84,109 @@ void shouldBuildExpectedLabelWhenEmptyOthers() {
...
+        CaseData caseData = buildCaseDataWithOthers(firstOther, additionalOther, confidentialOther);
+
+        Others others = service.prepareOthers(caseData);
+
+        assertThat(others.getAdditionalOthers()).containsOnly(additionalOther.get(0));
+    }
+
+    @Test
+    void shouldAddExpectedRespondentWhenHiddenDetailsMarkedAsNo() {
+        Other firstOther = otherWithDetailsHiddenValue("No");
```
toireasam, line 137: Not sure what this method is trying to do and name doesn't reflect it either

toireasam: On second review functionality looks correct but name should be shouldAddExpectedOtherWhenHiddenDetailsMarkedAsNo

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsServiceTest.java
```diff
@@ -448,75 +447,6 @@ void shouldNotRemoveConfidentialDetailsWhenNotMarkedAsConfidential() {
...
-            List<Element<Other>> others = service.combineOtherDetails(caseData.getAllOthers(),
-                caseData.getConfidentialOthers());
-
-            assertThat(others.get(0).getValue())
-                .isEqualTo(otherWithConfidentialFields(ID, NOT_CONFIDENTIAL).getValue());
-        }
-
-        @Test
-        void shouldMaintainOrderingOfOthersWhenComplexScenario() {
-            UUID otherId = randomUUID();
```
toireasam, line 494: Did we want to add similar test like this to the othersService? I think should be the same sort of logic except calling prepareOthers instesd of combineOtherDetails

service/src/test/java/uk/gov/hmcts/reform/fpl/service/OthersServiceTest.java
KiranVarma02: done

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ConfidentialDetailsServiceTest.java
KiranVarma02: done

PR #1480, opened 2020-08-07
feat: add in progress tag to task list (FPLA-1991)

tomaszpowroznik:
Screenshot 2020-08-07 at 09 31 12


service/src/test/java/uk/gov/hmcts/reform/fpl/service/validators/AllocationProposalCheckerIsStartedTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+@ExtendWith(MockitoExtension.class)
+class AllocationProposalCheckerIsStartedTest {
+
+    @InjectMocks
+    private AllocationProposalChecker allocationProposalChecker;
+
+    @ParameterizedTest
+    @NullSource
+    @MethodSource("emptyAllocationProposal")
```
KiranVarma02, line 25: Test passes even without this methodSource, i guess because of NullSource. So, do we need method source?

KiranVarma02: same in couple of other tests..

KiranVarma02:
other than my comment on tests, everything else looks good to me. 

Now lot of other teams are going to use your code on task list i guess :smiley:

service/src/test/java/uk/gov/hmcts/reform/fpl/service/validators/AllocationProposalCheckerIsStartedTest.java
tomaszpowroznik: yes, we need. If MerhodSource is remove only null values will be checked, if MethodSource is there then nullsources and all combinations from emptyAllocationProposal will be tested

service/src/main/java/uk/gov/hmcts/reform/fpl/service/TaskListRenderer.java
```diff
@@ -38,9 +38,28 @@
...
 
-    public TaskListRenderer(@Value("${resources.images.baseUrl}") String imagesBaseUrl) {
+    public TaskListRenderer(
+            @Value("${resources.images.baseUrl}") String imagesBaseUrl,
+            FeatureToggleService featureToggleService) {
         this.imagesBaseUrl = imagesBaseUrl;
+        this.featureToggleService = featureToggleService;
+    }
+
+    //TODO consider templating solution like mustache
```
Jamesnelson117, line 50: Are we doing anything with this TODO @tomaszpowroznik? Is this a tech debt, spike etc?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/ChildrenChecker.java
```diff
@@ -0,0 +1,65 @@
...
+                childParty.getKeyDates(),
+                childParty.getCareAndContactPlan(),
+                childParty.getAdoption(),
+                childParty.getMothersName(),
+                childParty.getFathersName(),
+                childParty.getFathersResponsibility(),
+                childParty.getSocialWorkerName(),
+                childParty.getAdditionalNeeds(),
+                childParty.getShowAddressInConfidentialTab(),
+                childParty.getLitigationIssues())
```
Jamesnelson117, line 60: I guess we could create a custom validator `isChildEmpty` or something but I guess both ways we can't avoid large statements like this.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/validators/AllocationProposalCheckerIsStartedTest.java
KiranVarma02: aha ok. Good to know. Thanks for explaining 👍

service/src/test/java/uk/gov/hmcts/reform/fpl/service/validators/ChildrenCheckerIsStartedTest.java
```diff
@@ -0,0 +1,172 @@
...
+                                .contactDirection("")
+                                .telephoneNumber("")
+                                .build())
+                        .socialWorkerTelephoneNumber(Telephone.builder()
+                                .telephoneUsageType("")
+                                .contactDirection("")
+                                .telephoneNumber("")
+                                .build())
+                        .build())
+                .build();
```
Jamesnelson117, line 96: minor: Could be private.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/EventsChecker.java
```diff
@@ -0,0 +1,127 @@
...
+        eventCheckers.put(OTHERS, othersChecker);
+        eventCheckers.put(COURT_SERVICES, courtServiceChecker);
+        eventCheckers.put(SUBMIT_APPLICATION, caseSubmissionChecker);
+    }
+
+    public List<String> validate(Event event, CaseData caseData) {
+        return ofNullable(eventCheckers.get(event))
+                .map(validator -> validator.validate(caseData))
+                .orElse(emptyList());
+    }
```
Jamesnelson117, line 108: So we have this `ValidateGroupService` with a `validateGroup`, maybe I missed this in the original PR but why did we not re-use?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/TaskListRenderer.java
tomaszpowroznik: afaik, we need approval from architects, Adam already started some discussion with Khaled

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/ChildrenChecker.java
tomaszpowroznik: there were couple options on table, one was to use reflection to get all declared fields and validates if they are empty or null, but there are couple of problems with this. One is that, when user open event which generates e.g default element in collection and there is nested compound element as Address it is not null, it is Address object with all empty fields. To handle this recursive solution should be implemented. Another issue is when object contains List<Elements<>>, elements can be initialised by ccd to have random ids.

One nice option is to created default object instance e.g Child emptyChild = Child.builder.build() and then do emptyChild.equals(actualChild) to check that child is empty, but again it requires equals method to be manually written to handle null, empty, empty Address objects and things like this.

Another problem is that not all fields in object are available to user, and they can be set in backend, but not visible to user. If we user automated solution like reflection, we would detect it as non empty object, but from user pov it is empty

Considered options i choose on purpose to write it this way.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/EventsChecker.java
tomaszpowroznik: ValidateGroupService validates entire case and returns list of all errors for entire case. Here we need errors for part of case data which are modified by particular event (e.g if event is AmendChildren, we need to get errors only for Children and not for entire case)

Jamesnelson117: Could we have solved this by creating more validation groups perhaps? That way we could of used `ValidateGroupService(caseData, childrenGroup.class)`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/TaskListRenderer.java
```diff
@@ -117,20 +122,27 @@ public String render(List<Task> allTasks) {
...
         });
 
         return section;
     }
 
-    private List<String> renderTask(Task task) {
+    private List<String> renderTask(Task task, boolean showInProgress) {
         final List<String> lines = new LinkedList<>();
 
         switch (task.getState()) {
```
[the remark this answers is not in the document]
tomaszpowroznik: what if toggle change in the middle? Not quite sure how it affect performance, definitely it can affect LD statistics which shows how many times toggle has been evaluated. This code basically sits inside loop

PR #1485, opened 2020-08-10
fix: pass c2 document instead of c110a in c2 notification

akomisarek:

https://tools.hmcts.net/jira/browse/FPLA-2081

### Change description ###

Fixes the above defect and provides expected latest c2 bundle. 

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsSubmittedControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsSubmittedControllerTest.java
@@ -21,7 +21,6 @@ import uk.gov.hmcts.reform.fpl.model.common.DocumentReference;
 import uk.gov.hmcts.reform.fpl.model.notify.c2uploaded.C2UploadedTemplate;
 import uk.gov.hmcts.reform.fpl.service.DocumentDownloadService;
 import uk.gov.hmcts.reform.fpl.service.payment.PaymentService;
-import uk.gov.hmcts.reform.fpl.utils.TestDataHelper;
 import uk.gov.hmcts.reform.idam.client.IdamApi;
 import uk.gov.hmcts.reform.idam.client.models.UserInfo;
 import uk.gov.service.notify.NotificationClient;
@@ -59,7 +58,8 @@ class UploadC2DocumentsSubmittedControllerTest extends AbstractControllerTest {
     private static final String LOCAL_AUTHORITY_CODE = "example";
     private static final Long CASE_ID = 12345L;
     private static DocumentReference applicationDocument;
-    private static final byte[] APPLICATION_BINARY = TestDataHelper.DOCUMENT_CONTENT;
+    private static DocumentReference latestC2Document;
+    private static final byte[] C2_BINARY = {5, 4, 3, 2, 1};
 
     @MockBean
     private NotificationClient notificationClient;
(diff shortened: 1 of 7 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProvider.java
```diff
@@ -24,14 +25,15 @@
...
+    public C2UploadedTemplate buildC2UploadNotificationTemplate(final CaseDetails caseDetails,
+                                                                final DocumentReference latestC2) {
         CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
 
         C2UploadedTemplate adminTemplateForC2 = new C2UploadedTemplate();
         adminTemplateForC2.setCallout(buildCallout(caseData));
         adminTemplateForC2.setRespondentLastName(getFirstRespondentLastName(caseData.getRespondents1()));
         adminTemplateForC2.setCaseUrl(getCaseUrl(caseDetails.getId()));
-        adminTemplateForC2.setDocumentLink(linkToAttachedDocument(caseData.getSubmittedForm()));
+        adminTemplateForC2.setDocumentLink(linkToAttachedDocument(latestC2));
```
KiranVarma02, line 36: looks like what we need is already in caseData.. would this work?
    adminTemplateForC2.setDocumentLink(linkToAttachedDocument(caseData.getLastC2DocumentBundle().getDocument()));

akomisarek: Probably, but to be fair I prefer having the information "what was added" in the event, which makes it more decoupled. Probably wouldn't want to change this right now, but happy to refactor in the "restrict" bug if you insist :) 

@tomaszpowroznik what are your thoughts?

tomaszpowroznik: agree, relying on the fact that something has been added as last is a bit slippery. In theory if we send notification  with @async, it could happened that something else has been added meanwhile to the case.

akomisarek: Ok, I will leave as it is if you don't mind. Even if there is more code I believe my approach is easier to follow and more logically correct - we raise an event mentioning what was changed. But I am not much fussed and if you press I can refactor in the restrict "story" :)

KiranVarma02: i agree what we have here is logically correct 👍

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsSubmittedControllerTest.java
```diff
@@ -59,7 +58,8 @@
     private static final String LOCAL_AUTHORITY_CODE = "example";
     private static final Long CASE_ID = 12345L;
     private static DocumentReference applicationDocument;
```
tomaszpowroznik, line 60: probably not relevant for this test, can be removed or if still want to have it for data completness, can be hidden in buildCommonNotificationParameters

service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProviderTest.java
```diff
@@ -30,14 +29,15 @@
     @Autowired
     private C2UploadedEmailContentProvider c2UploadedEmailContentProvider;
 
-    private static final byte[] APPLICATION_BINARY = TestDataHelper.DOCUMENT_CONTENT;
+    private static final byte[] C2_DOCUMENT_BINARY = {5};
```
tomaszpowroznik, line 32: TestDataHelper.DOCUMENT_CONTENT can be used

PR #1487, opened 2020-08-11
chore: Remove child dob validation dead code

tomaszpowroznik:


tomaszpowroznik:
> LGTM but is validation not required? or is it done somewhere else?

yes, validation is done on case submission, see ChildParty


    @Override
    @NotNull(message = "Tell us the date of birth of all children in the case",
        groups = {Default.class, SealedSDOGroup.class})
    @PastOrPresent(message = "Date of birth is in the future. You cannot send this application until that date")
    public LocalDate getDateOfBirth() {
        return super.getDateOfBirth();
    }

PR #1497, opened 2020-08-12
fix: replace c2 download document link (FPLA-2087)

toireasam:
https://tools.hmcts.net/jira/browse/FPLA-2087

### Change description ###
Replaces c2 download link with link to c2 document

```
[ ] Yes
[ X ] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsSubmittedControllerTest.java
```diff
@@ -316,19 +314,6 @@ private CaseDetails createCase(Map<String, Object> data) {
             "caseUrl", "http://fake-url/cases/case-details/12345");
     }
 
-    private Map<String, Object> buildExpectedNotificationParams() {
-        C2UploadedTemplate c2UploadedTemplate = new C2UploadedTemplate();
```
toireasam, line 320: Removed this as I think we agreed that controllers didn't need fine level details of the parameters and instead could just use anyMap() so didn't see point in maintaining this as it's covered in unit tests - happy if someone disagrees.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseUrlService.java
```diff
@@ -8,6 +9,7 @@
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 @Service
+@Getter
```
KiranVarma02, line 12: not an issue but in theory is it okay to have Getter on a service? but then i am not sure if this class is a service 🤷

toireasam: hmm, good point. I added this here because i wanted to access xuiBaseUrl but i think i could add this property into the class I need it and remove the dependency on this service. So will remove. and don't think it should be here :)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/C2UploadedEmailContentProvider.java
```diff
@@ -33,7 +39,8 @@ public C2UploadedTemplate buildC2UploadNotificationTemplate(final CaseDetails ca
         adminTemplateForC2.setCallout(buildCallout(caseData));
         adminTemplateForC2.setRespondentLastName(getFirstRespondentLastName(caseData.getRespondents1()));
         adminTemplateForC2.setCaseUrl(getCaseUrl(caseDetails.getId()));
-        adminTemplateForC2.setDocumentLink(linkToAttachedDocument(latestC2));
+        adminTemplateForC2.setDocumentUrl(concatUrlAndMostRecentUploadedDocumentPath(
+            latestC2.getBinaryUrl(), xuiBaseUrl));
```
akomisarek, line 43: It might be only me but I would prefer other order of the parameters - outcome of this method xuiBaseUrl + latestC2.getBinary... However the order passed to. the method is different, minor issue, but I think it reads better other way araound.

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/DocumentsHelperTest.java
```diff
@@ -49,6 +50,27 @@ void shouldReturnFalseIfDocumentHasNotBinaries() {
...
+
+    @Test
+    void shouldCatchExceptionAndReturnEmptyStringWhenDocumentLinkIsNotInCorrectFormat() {
+        String invalidMostRecentDocumentLink = "http://dm-store:8080/documents/b28f859b-7521-4c84-9057-47e56afd773f/binary^";
+        String url = "http://fake-url";
+
+        String documentUrl = concatUrlAndMostRecentUploadedDocumentPath(invalidMostRecentDocumentLink, url);
+
+        assertThat(documentUrl).isEqualTo("");
+    }
```
akomisarek, line 72: I am actually not sure what's better here-  returning empty string or just. throwing the exception. It's helper class which shouldn't decide how the errors are handled, that should be calleer responsibility. But I guess I am happy in this instance. As it is only used there for email notification.

toireasam: Yes I can see what you mean here! I think in order to get this bug fix into testing asap let's agree to leave this for now (a couple of our other helpers actually throw exceptions / follow this same format) and if we wish to iterate back on it again we can. and yes only used for email notification so nothing major either!

PR #1503, opened 2020-08-14
tmp: Add Add case number to cases in prepare for hearing state (FPLA-2098)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2098

### DIFF
--- a/ccd-definition/CaseEvent/CareSupervision/MultiState.json
+++ b/ccd-definition/CaseEvent/CareSupervision/MultiState.json
@@ -88,7 +88,7 @@
     "Name": "Add case number",
     "Description": "Add case number",
     "DisplayOrder": 5,
-    "PreConditionState(s)": "Submitted",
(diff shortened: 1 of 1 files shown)


ccd-definition/CaseEvent/CareSupervision/MultiState.json
```diff
@@ -88,7 +88,7 @@
     "Name": "Add case number",
     "Description": "Add case number",
     "DisplayOrder": 5,
-    "PreConditionState(s)": "Submitted",
+    "PreConditionState(s)": "Submitted;PREPARE_FOR_HEARING",
```
joez-f, line 91: this event triggers call to robotics, is that OK in prepare for hearing state?

akomisarek: It doesn't send if the number was already there.

PR #1506, opened 2020-08-15
refactor: Use CaseData object when preparing data to build response instead of Map (FPL-2094)

KiranVarma02:

https://tools.hmcts.net/jira/browse/FPLA-2094

### Change description ###
remove use of strings from controllers

```
[ ] Yes
[* ] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AddNoteController.java
```diff
@@ -26,21 +25,22 @@
...
 
         CaseNote caseNote = service.buildCaseNote(requestData.authorisation(), caseData.getCaseNote());
         List<Element<CaseNote>> caseNotes = service.addNoteToList(caseNote, caseData.getCaseNotes());
 
-        caseDetails.getData().put("caseNotes", caseNotes);
-        caseDetails.getData().remove("caseNote");
+        CaseData updatedCase = caseData.toBuilder()
+            .caseNotes(caseNotes)
+            .caseNote(null)
+            .build();
```
KiranVarma02, line 40: this scenario is covered in unit test

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AllocationDecisionController.java
```diff
@@ -10,49 +9,47 @@
...
-        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+        CaseData caseData = caseDataConverter.convertToCaseData(callbackRequest);
 
         Allocation allocationDecision = service.createDecision(caseData);
 
-        Map<String, Object> data = caseDetails.getData();
-        data.put("allocationDecision", allocationDecision);
+        CaseData updatedCase = caseData.toBuilder()
+            .allocationDecision(allocationDecision)
+            .build();
```
KiranVarma02, line 34: covered in existing unit test

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AllocationDecisionController.java
```diff
@@ -10,49 +9,47 @@
...
-        CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
+        CaseData caseData = caseDataConverter.convertToCaseData(callbackRequest);
 
         Allocation allocationDecision = service.setAllocationDecisionIfNull(caseData);
 
-        Map<String, Object> data = caseDetails.getData();
-        data.put("allocationDecision", allocationDecision);
+        CaseData updatedCase = caseData.toBuilder()
+            .allocationDecision(allocationDecision)
+            .build();
```
KiranVarma02, line 49: covered in existing unit test

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -72,31 +71,34 @@
...
         Order standardDirectionOrder = caseData.getStandardDirectionOrder();
 
         if (standardDirectionOrder != null && standardDirectionOrder.getDateOfIssue() != null) {
             dateOfIssue = parseLocalDateFromStringUsingFormat(standardDirectionOrder.getDateOfIssue(), DATE);
         }
 
-        caseDetails.getData().put("dateOfIssue", dateOfIssue);
+        CaseData updatedCaseData = caseData.toBuilder()
+            .dateOfIssue(dateOfIssue)
+            .build();
```
KiranVarma02, line 91: existing test

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersControllerAboutToSubmitTest.java
```diff
@@ -127,6 +105,16 @@ void shouldPopulateHiddenCCDFieldsInStandardDirectionOrderToPersistData() {
         assertThat(fileName.getValue()).isEqualTo(SEALED_ORDER_FILE_NAME);
     }
 
+    @Test
+    void shouldRemoveJudgeAndLegalAdvisorAndDateOfIssueFromCaseData() {
```
KiranVarma02, line 109: if there is a better name, please suggest

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -193,7 +192,11 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
             .build();
 
         //add sdo to case data for document generation
-        CaseData updated = caseData.toBuilder().standardDirectionOrder(order).build();
+        CaseData updated = caseData.toBuilder()
+            .standardDirectionOrder(order)
+            .judgeAndLegalAdvisor(null)
+            .dateOfIssue(null)
+            .build();
```
KiranVarma02, line 199: covered in new and existing test

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -193,7 +192,11 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
             .build();
 
         //add sdo to case data for document generation
-        CaseData updated = caseData.toBuilder().standardDirectionOrder(order).build();
+        CaseData updated = caseData.toBuilder()
+            .standardDirectionOrder(order)
+            .judgeAndLegalAdvisor(null)
+            .dateOfIssue(null)
```
KiranVarma02, line 198: covered in new test

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/DraftOrdersController.java
```diff
@@ -193,7 +192,11 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
             .build();
 
         //add sdo to case data for document generation
-        CaseData updated = caseData.toBuilder().standardDirectionOrder(order).build();
+        CaseData updated = caseData.toBuilder()
+            .standardDirectionOrder(order)
```
KiranVarma02, line 196: covered in existing test

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverter.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.ccd.client.model.CallbackRequest;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.hmcts.reform.fpl.model.CaseData;
+
+import java.util.Map;
+
+@Component
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class CaseDataConverter {
```
tomaszpowroznik, line 15: maybe just CaseConverter

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverter.java
```diff
@@ -0,0 +1,31 @@
...
+    private final ObjectMapper mapper;
+
+    @SuppressWarnings("unchecked")
+    public Map<String, Object> convertToMap(CaseData caseData) {
+        return mapper.convertValue(caseData, Map.class);
+    }
+
+    public CaseData convertToCaseData(CallbackRequest callbackRequest) {
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        return mapper.convertValue(caseDetails.getData(), CaseData.class);
```
tomaszpowroznik, line 25: use other method:
convertToCaseData(callbackRequest.getCaseDetails())

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverterTest.java
```diff
@@ -0,0 +1,64 @@
...
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {JacksonAutoConfiguration.class, CaseDataConverter.class})
+class CaseDataConverterTest {
+
+    @Autowired
+    private CaseDataConverter caseDataConverter;
+
+    @Test
+    void shouldConvertCaseDataToMap() {
+        assertThat(caseDataConverter.convertToMap(getCaseData())).containsKey("caseLocalAuthority");
```
tomaszpowroznik, line 31: i found this a bit weird, why we checking particular properties only?
I would say that ObjectMapper should be Mock/Spy so we can assert that ObjectMapper has been  invoked with correct arguments and what is returned by this method is what we get back from object mapper. We dont want to test objectmapper

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverterTest.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        @Test
+        void shouldConvertCallbackRequestToCaseData() {
+            assertThat(caseDataConverter.convertToCaseData(request).getCaseLocalAuthority()).isEqualTo(localAuthority);
+        }
+
+        @Test
+        void shouldConvertCaseDetailsToCaseData() {
+            final CaseDetails caseDetails = request.getCaseDetails();
```
tomaszpowroznik, line 53: not used at all

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverterTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        @BeforeEach
+        void setup() {
+            localAuthority = "LA1";
+            request = callbackRequest(Map.of("localAuthority", localAuthority));
+        }
+
+        @Test
+        void shouldConvertCallbackRequestToCaseData() {
+            assertThat(caseDataConverter.convertToCaseData(request).getCaseLocalAuthority()).isEqualTo(localAuthority);
```
tomaszpowroznik, line 48: same here, why only some of properties are checked?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverterTest.java
```diff
@@ -0,0 +1,64 @@
...
+            final CaseDetails caseDetails = request.getCaseDetails();
+
+            assertThat(caseDataConverter.convertToCaseData(request).getCaseLocalAuthority()).isEqualTo(localAuthority);
+        }
+
+    }
+
+    private CaseData getCaseData() {
+        LocalDateTime dateTime = LocalDateTime.of(2099, 1, 1, 10, 0, 0);
+        return buildCaseDataForCMODocmosisGeneration(dateTime);
```
tomaszpowroznik, line 62: looks weird to have CMODocmosis stuff here

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AddNoteController.java
tomaszpowroznik, line 40: this entire thing is over engineered tbh. 
1. No need to pass requestData.authorisation to service, service can inject RequestData on its own,
2. having addNoteToList which just add element to list is overengineered

Why not having just single method, public CaseData service.addNote(CaseDate caseData){
 //get note, enrich with user info, add to list, reset note, return case data
}

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AddNoteController.java
```diff
@@ -26,21 +25,22 @@
 public class AddNoteController {
     private final CaseNoteService service;
     private final RequestData requestData;
-    private final ObjectMapper mapper;
+    private final CaseDataConverter caseDataConverter;
```
tomaszpowroznik, line 28: if we are looking for generic solution then we could have abstract controller which is extended by all our controlers. It can have e.g method saveCase(CaseData caseData){
            .data(caseDataConverter.convertToMap(updatedCase))
            .build();
}

then our controller at the end just invoke saveCase(CaseData caseData)

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverter.java
KiranVarma02: agree, don't know what i was thinking...

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverterTest.java
KiranVarma02: removed

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverterTest.java
KiranVarma02: I was not sure if tests add much value for this entire class but did half hearted attempt :). 

I have added as you suggested, do you think there is value in this now?

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDataConverterTest.java
KiranVarma02: removed

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AddNoteController.java
KiranVarma02: as discussed, my aim of this Spike/PR is just to remove strings while building case data.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/AddNoteController.java
KiranVarma02: I like this too but maybe i will try to leave current scope to just removing strings.

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseConverter.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Map;
+
+@Component
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class CaseConverter {
+    private final ObjectMapper mapper;
+
+    @SuppressWarnings("unchecked")
+    public Map<String, Object> convertToMap(CaseData caseData) {
+        return mapper.convertValue(caseData, Map.class);
```
lewisbirks, line 19: I would suggest changing `Map.class` to `new TypeReference<>(){}` which would then allow the removal of the suppress warnings

lewisbirks:
If we want to go ahead with this approach we would need to ensure that there are no unintended side effects. Methods could be altering collections or other objects that use the same reference as the property in the case data class in ways that had results that weren't intended to be returned back to the user.

lewisbirks:
Another issue would be for fields that aren't persisted in case data, if there is a field that exists in the case field json file but not in the CaseData class then using this new approach the field would be lost. I am not sure how many fields this would effect but I believe we don't store the `sendToCtsc` flag in case data.

KiranVarma02:
@lewisbirks - I have discussed your point with @akomisarek about field in json file and not in casedata class. Even if we remove using strings for the fields which exist in casedata class, i think that's still better than what we have now. 

About your other point, to see if change will have other unintended issues, in this PR we are starting with 3 controllers (to take less risk). This is also the reason why i am asking everyone for review and we are going to test. I don't see any issues in this PR at this moment. 

I am interested to see what others think!

lewisbirks:
>@lewisbirks - I have discussed your point with @akomisarek about field in json file and not in casedata class (My understandis is you are talking about temp fields). Even if we remove using strings for the fields which exist in casedata class, i think that's still better than what we have now.

My point was not about temporary fields. There is at least one instance of a flag having been set in the codebase, `sendToCtsc`, that does not appear in the case data class. If we go with your approach then the data stored in these fields will be lost as they wouldn't have been stored when the data map was deserialised into the class.

KiranVarma02:
Thanks @lewisbirks , I was not aware of that. I don't understand why that is the case!. Speaking to @akomisarek and he suggested raising a tech debt to correct that! (hopefully not too many cases). Let me know what you think!

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseConverterTest.java
```diff
@@ -0,0 +1,50 @@
...
+import uk.gov.hmcts.reform.fpl.model.CaseData;
+
+import java.util.Map;
+
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {JacksonAutoConfiguration.class, CaseConverter.class})
+class CaseConverterTest {
```
lewisbirks, line 22: I am not sure about these tests, surely it would be better to assert that the expected output is correct rather than just that the object mapper was invoked

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseConverter.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+import uk.gov.hmcts.reform.fpl.model.CaseData;
+
+import java.util.Map;
+
+@Component
+@RequiredArgsConstructor(onConstructor = @__(@Autowired))
+public class CaseConverter {
```
lewisbirks, line 15: Not sure wether this should be a component or just added to the abstract controller class

lewisbirks: It feels like a lot of extra code which could simply be done in the abstract class

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseConverterTest.java
KiranVarma02: assertion will be testing object mapper itself (not necessary as @tomaszpowroznik mentioned above) as this method doesn't do much. 

🤷‍♂️🤷‍♂️🤷‍♂️ I personally don't see lot of value for tests in this case

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseConverter.java
KiranVarma02: agree with you on this.

KiranVarma02:
close temporarily until tests are fixed

PR #1522, opened 2020-08-20
fix: Update e2e tests to check payment made on application submission…

tomaszpowroznik:
… and C2 upload

### DIFF
--- a/e2e/fixtures/gatekeeping.json
+++ b/e2e/fixtures/gatekeeping.json
@@ -118,7 +118,7 @@
             "partyId": "3a8785c6-178d-4138-bd1a-5eea0debed25",
             "jobTitle": "Legal adviser",
             "partyType": "ORGANISATION",
-            "pbaNumber": "PBA1234567",
+            "pbaNumber": "PBA0082848",
             "clientCode": "8888",
             "mobileNumber": {
               "telephoneNumber": "07000000000"
--- a/e2e/pages/paymentHistory.page.js
+++ b/e2e/pages/paymentHistory.page.js
@@ -1,0 +1,20 @@
+const {I} = inject();
+const config = require('../config');
+
+const paymentsLocator = '//span[text()="Processed payments"]';
+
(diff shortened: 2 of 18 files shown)


e2e/fixtures/gatekeeping.json
```diff
@@ -118,7 +118,7 @@
             "partyId": "3a8785c6-178d-4138-bd1a-5eea0debed25",
             "jobTitle": "Legal adviser",
             "partyType": "ORGANISATION",
-            "pbaNumber": "PBA1234567",
+            "pbaNumber": "PBA0082848",
```
KiranVarma02, line 121: why this change in multiple files?

akomisarek: This is the test pba which works for us on aat and provide "real" results.

e2e/pages/paymentHistory.page.js
```diff
@@ -0,0 +1,20 @@
+const {I} = inject();
+const config = require('../config');
+
+const paymentsLocator = '//span[text()="Processed payments"]';
+
+const buildPaymentLocator = amount => config.mockedPayment ? `${paymentsLocator}/../table//tr[.//td[text()="Success"]]` : `${paymentsLocator}/../table//tr[.//td[text()="${amount}"] and .//td[text()="Success"]]`;
```
tomaszpowroznik, line 6: config.mockedPayment is true on local env and false in pipeline. On local environment we use static wiremock mappings to simulate payment sercvice, thus fee is always same regardless of what has been actually requested to be paid. For that reason, when working with mocks, fee amount is not checked

e2e/fixtures/gatekeeping.json
tomaszpowroznik: potentially can be moved to config, but it requires refactoring data json files like this, as we cannot inject config value directly into json

akomisarek (approved):

PR #1527, opened 2020-08-20
fix: fix flow logic for Review CMO in mid-event

akomisarek:
https://tools.hmcts.net/jira/browse/FPLA-2118


tomaszpowroznik:
is it not enough to change ReviewCMOService.getSelectedCMO

if (("MULTI").equals(caseData.getNumDraftCMOs())) {
to
if (getCMOsReadyForApproval(caseData).size()>1) {

if the root cause is that it returns wrong selecttion due to lost numDraftCMOs, then maybe it is enough

PR #1535, opened 2020-08-24
Revert "chore(deps-dev): bump codeceptjs from 2.6.8 to 2.6.9"

akomisarek:
Reverts hmcts/fpl-ccd-configuration#1529

Seems e2e are a lot more flakey.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "FPL CCD e2e tests",
   "devDependencies": {
     "babel-eslint": "^10.1.0",
-    "codeceptjs": "2.6.9",
+    "codeceptjs": "2.6.8",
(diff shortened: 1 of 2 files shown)


KiranVarma02 (approved):

KiranVarma02:
What is the reason for your suspicion?

lewisbirks:
> What is the reason for your suspicion?

Looking at the changelog in the previous PR I guess it could be something to do with this
> [Puppeteer][Playwright] SessionStorage is now cleared in after hook.
None of the other listed changes effect us I think.

@akomisarek Do we want to try updating to the most recent version (2.6.10) and give that a go to see if it is any more stable?

PR #1538, opened 2020-08-25
feat: Retry on case sharing failures (FPLA-2129)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2129

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
@@ -4,6 +4,8 @@ import feign.FeignException;
 import feign.Request;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockito.BDDMockito;
+import org.mockito.stubbing.Answer;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.OverrideAutoConfiguration;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
@@ -42,6 +44,7 @@ import static org.mockito.ArgumentMatchers.anyMap;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static uk.gov.hmcts.reform.fpl.CaseDefinitionConstants.CASE_TYPE;
 import static uk.gov.hmcts.reform.fpl.CaseDefinitionConstants.JURISDICTION;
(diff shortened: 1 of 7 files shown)


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
```diff
@@ -230,13 +270,29 @@ private void verifyTaskListUpdated(CaseDetails caseDetails) {
...
+        });
     }
 
     private void givenPRDWillReturn(List<String> userIds) {
-        given(organisationApi.findUsersByOrganisation(USER_AUTH_TOKEN, SERVICE_AUTH_TOKEN, Status.ACTIVE, false))
-            .willReturn(organisation(userIds));
+        givenPRDWillAnswer(invocation -> organisation(userIds));
+    }
+
+    private void givenPRDWillAnswer(Answer... answers) {
```
lewisbirks, line 283:
```suggestion
    private void givenPRDWillAnswer(Answer<?>... answers) {
```

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationControllerTest.java
```diff
@@ -230,13 +270,29 @@ private void verifyTaskListUpdated(CaseDetails caseDetails) {
...
+        BDDMockito.BDDMyOngoingStubbing<OrganisationUsers> stub = given(organisationApi.findUsersByOrganisation(
+            USER_AUTH_TOKEN,
+            SERVICE_AUTH_TOKEN,
+            Status.ACTIVE,
+            false));
+        for (Answer answer : answers) {
```
lewisbirks, line 289:
```suggestion        
        for (Answer<?> answer : answers) {
```

service/src/main/java/uk/gov/hmcts/reform/fpl/exceptions/GrantCaseAccessException.java
```diff
@@ -21,7 +21,7 @@ public GrantCaseAccessException(String caseId, Set<String> userIds, Set<CaseRole
     }
 
     public GrantCaseAccessException(String caseId, String localAuthority, Set<CaseRole> caseRoles, Exception e) {
-        super(String.format("User from %s not granted %s to case %s", localAuthority, caseRoles, caseId), e);
+        super(String.format("Users from %s not granted %s to case %s", localAuthority, caseRoles, caseId), e);
```
lewisbirks, line 24: The above has it as `User(s)`, we should probably be consistent and go with your change for both

lewisbirks (approved):
Think it is okay from my brief look through the retryable documentation, would be good if someone else had a look over it though

service/src/main/resources/application.yaml
```diff
@@ -221,6 +221,8 @@ fees-register:
   api:
     url: http://fake-fees-register-api
 
+retry:
+  delay: 50
```
akomisarek, line 225: Maybe bigger? is there exponential backoff?

akomisarek (approved):
Looks good for me, not sure if 50 is good enough delay.

service/src/main/resources/application.yaml
lewisbirks: I think it could be exponential if the backoff is changed to
```java
  backoff = @Backoff(
    delayExpression = "#{${retry.delay:1000}}",
    multiplier = 2,
    maxDelay = 900000
  )
```

tomaszpowroznik: it is for integration test only

PR #1544, opened 2020-08-27
feat: hide C21 orders generated by mistake (FPLA-2107)

lewisbirks:

[FPLA-2107](https://tools.hmcts.net/jira/browse/FPLA-2107)

### Change description ###

Introduced a new event for a super user to remove C21 orders generated by mistake and display them in their own collection that only the superuser has visibility to.

```
[ ] Yes
[x] No
```


e2e/tests/hmctsSuperUserAdministersCase_test.js
```diff
@@ -0,0 +1,22 @@
...
+
+Scenario('HMCTS super user removes an order from a case', async (I, caseViewPage, loginPage, removeOrderEventPage) => {
+  await caseViewPage.goToNewActions(config.superUserActions.removeOrder);
+  let order = orderCaseData.caseData.orderCollection[0];
+  const labelToSelect = order.value.title + ' - ' + order.value.dateOfIssue;
+  removeOrderEventPage.selectOrderToRemove(labelToSelect);
+  await I.retryUntilExists(() => I.click('Continue'), removeOrderEventPage.fields.reason);
+  removeOrderEventPage.addRemoveOrderReason('Entered incorrect order');
+  await I.completeEvent('Submit');
+  // await orderFunctions.assertOrder(I, caseViewPage, order, defaultIssuedDate);
```
toireasam, line 21: Does this need uncommented if we want to assert? or removed?

toireasam:
FPLA docker has been added by mistake :)

e2e/tests/hmctsSuperUserAdministersCase_test.js
lewisbirks: That is the final part that needs doing, @KiranVarma02 is doing this now

lewisbirks:
> FPLA docker has been added by mistake :)

@ toireasam needed for the super user, the PR in fpla-docker has just been merged for it so once it is updated in the repo I can undo the change here

e2e/config.js
```diff
@@ -31,6 +31,10 @@ module.exports = {
     email: 'hmcts-admin@example.com',
     password: process.env.CA_USER_PASSWORD || defaultPassword,
   },
+  hmctsSuperUser: {
+    email: 'hmcts-superuser@example.com',
+    password: process.env.CA_USER_PASSWORD || defaultPassword,
+  },
```
toireasam, line 37: are we happy for this to have the court admin user password? or do we need another one like SUPER_USER_PASSWORD?

e2e/pages/events/removeOrderEvent.page.js
```diff
@@ -0,0 +1,13 @@
+const {I} = inject();
+
+module.exports = {
+  fields: {
+    reason: '#removeOrderReason',
+  },
+
+
+
```
toireasam, line 13: couple of spaces we could remove maybe :)

e2e/tests/hmctsSuperUserAdministersCase_test.js
toireasam: oh sure np :)

typings/codeceptjs.d.ts
```diff
@@ -507,6 +507,11 @@ declare namespace CodeceptJS {
     addNote(): void,
 
   }
+
+  export interface removeAnOrderPage {
+    selectOrderToRemove(): void,
+    enterReasonToRemoveOrder(): void
+  }
```
toireasam, line 514: Is this needed? I'm not sure it's used

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RemoveOrderController.java
```diff
@@ -0,0 +1,91 @@
...
+        List<Element<GeneratedOrder>> orders = caseData.getOrderCollection();
+        List<Element<GeneratedOrder>> hiddenOrders = caseData.getHiddenOrders();
+
+        service.hideOrder(orders, hiddenOrders, caseData.getRemovableOrderList(), caseData.getReasonToRemoveOrder());
+
+        data.put("orderCollection", orders);
+        data.put("hiddenOrders", hiddenOrders);
+        removeTemporaryFields(
+            caseDetails,
+            "removableOrderList",
```
toireasam, line 79: code smell complaining about "removableOrderList" being referenced like this three times.. is it something we want to put at top of class like as static variable or do we want to ignore the code smell?

service/src/test/java/uk/gov/hmcts/reform/fpl/model/order/generated/GeneratedOrderTest.java
```diff
@@ -0,0 +1,56 @@
...
+            .build();
+
+        assertThat(order.isRemovable()).isFalse();
+    }
+
+    @Test
+    void shouldReturnTitleAppendedByDateOfIssueWhenTitlePresent() {
+        GeneratedOrder order = GeneratedOrder.builder()
+            .title("do you remember")
+            .dateOfIssue("21st September")
```
toireasam, line 41: nice song ;) haha!

e2e/pages/events/removeOrderEvent.page.js
lewisbirks: This is an outdated version

typings/codeceptjs.d.ts
lewisbirks: Not sure either, @KiranVarma02 did you add this?

e2e/config.js
lewisbirks: Probably should be, do you know if we set those env variables anywhere?

e2e/tests/hmctsSuperUserAdministersCase_test.js
KiranVarma02: done

e2e/config.js
KiranVarma02: I didn't find where we set ! Does anyone know? or its probably not set and using defaultPassword everywhere 😄

typings/codeceptjs.d.ts
KiranVarma02: removed.

e2e/config.js
lewisbirks: Looks like the only place we set the vars are for the smoke test and that is for the `SMOKE_TEST_LA_USER_USERNAME` and `SMOKE_TEST_LA_USER_PASSWORD`. I think keeping it different is sensible however as if we do decide to set the variable it may be different from the admin one

toireasam: I can see we set SYSTEM_UPDATE_USER_USERNAME in jenkins cnp which we read from key vault. for the rest seems to me like not set and use defaultPassword but not 100% sure (looks that way though)

e2e/helpers/generated_order_helper.js
```diff
@@ -168,11 +168,13 @@ module.exports = {
...
-  async assertOrder(I, caseViewPage, order, defaultIssuedDate, hasAllocatedJudge = false) {
+  async assertOrder(I, caseViewPage, order, defaultIssuedDate, hasAllocatedJudge = false, isOrderRemoved = false) {
     caseViewPage.selectTab(caseViewPage.tabs.orders);
     const numberOfOrders = await I.grabNumberOfVisibleElements('//*[text() = \'Type of order\']');
-    const orderHeading = `Order ${numberOfOrders}`;
-    I.seeInTab([orderHeading, 'Type of order'], order.fullType);
+    let orderHeading = `Order ${numberOfOrders}`;
+
+    isOrderRemoved ?
+      orderHeading = `Removed orders ${numberOfOrders}`: orderHeading = `Order ${numberOfOrders}`;
```
toireasam, line 177: minor: whitespace between ' and :

e2e/tests/hmctsSuperUserAdministersCase_test.js
```diff
@@ -20,7 +21,10 @@ Scenario('HMCTS super user removes an order from a case', async (I, caseViewPage
...
   await I.completeEvent('Submit');
   I.seeEventSubmissionConfirmation(config.superUserActions.removeOrder);
-  const defaultIssuedDate = new Date();
-
-  await orderFunctions.assertOrder(I, caseViewPage, order, defaultIssuedDate, true, true);
+  const issuedDate = new Date(2020, 4,26, 14, 33);
+  order = {...blankOrder, ...{
+    document: 'Blank order (C21).pdf',
+    reasonForRemoval: 'Entered incorrect order',
+  } };
```
toireasam, line 28: Maybe we could just change the fixture's document property to the above + add reasonforRemoval into the fixture also? I don't think this would affect any other test (maybe i am wrong :) )

lewisbirks: removal reason is fine to add however the document name that the system generates is different to the document name that is constructed in the case data used for this test

toireasam: ok sure :)

KiranVarma02: done

e2e/helpers/generated_order_helper.js
KiranVarma02: done

e2e/config.js
KiranVarma02: don't know where we are going to set password but i have added new env variable - SUPER_USER_PASSWORD

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RemoveOrderController.java
KiranVarma02: moved as a static variable

e2e/config.js
toireasam: Think it's just if we want to set it in future the env variable is there :)

toireasam (approved):
LGTM

e2e/helpers/generated_order_helper.js
```diff
@@ -168,11 +168,13 @@ module.exports = {
...
-  async assertOrder(I, caseViewPage, order, defaultIssuedDate, hasAllocatedJudge = false) {
+  async assertOrder(I, caseViewPage, order, defaultIssuedDate, hasAllocatedJudge = false, isOrderRemoved = false) {
     caseViewPage.selectTab(caseViewPage.tabs.orders);
     const numberOfOrders = await I.grabNumberOfVisibleElements('//*[text() = \'Type of order\']');
-    const orderHeading = `Order ${numberOfOrders}`;
-    I.seeInTab([orderHeading, 'Type of order'], order.fullType);
+    let orderHeading = `Order ${numberOfOrders}`;
+
+    isOrderRemoved ?
+      orderHeading = `Removed orders ${numberOfOrders}` : orderHeading = `Order ${numberOfOrders}`;
```
akomisarek, line 177: Cannot this just be single line with const?

e2e/tests/hmctsSuperUserAdministersCase_test.js
```diff
@@ -0,0 +1,29 @@
...
+  removeOrderEventPage.selectOrderToRemove(labelToSelect);
+  await I.retryUntilExists(() => I.click('Continue'), removeOrderEventPage.fields.reason);
+  removeOrderEventPage.addRemoveOrderReason('Entered incorrect order');
+  await I.completeEvent('Submit');
+  I.seeEventSubmissionConfirmation(config.superUserActions.removeOrder);
+  const issuedDate = new Date(2020, 4,26, 14, 33);
+  order = {...blankOrder, ...{
+    document: 'Blank order (C21).pdf',
+  } };
+  await orderFunctions.assertOrder(I, caseViewPage, order, issuedDate, false, true);
```
akomisarek, line 28: I think it's hard to read - it would be much better if method names says what we are asserting. Impossible to say now what is happening without going deeper, test should be a bit cleaner.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/order/generated/GeneratedOrder.java
```diff
@@ -26,4 +30,14 @@
     private final String courtName;
     @JsonSerialize(contentConverter = BasicChildConverter.class)
     private final List<Element<Child>> children;
+    private String removalReason;
+
+    @JsonIgnore
+    public boolean isRemovable() {
+        return BLANK_ORDER.getLabel().equals(type);
```
akomisarek, line 37: I think it would be better if getLabel is not used bu there is dedicated flag on the order telling us that information. 

Something like:
`
GeneratedOrderType.fromType(type).isRemovable();
`
thoughts?

akomisarek (approved):

e2e/helpers/generated_order_helper.js
KiranVarma02: you mean for orderHeading? I don't think so, because i am changing orderHearing based on isOrderRemoved.

e2e/tests/hmctsSuperUserAdministersCase_test.js
KiranVarma02: I tried to reuse existing assertion which asserts order, now method asserts for removed order too which is exactly same thing but with different title. It made sense to me but happy to take suggestion and work on it as a small follow up PR.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/order/generated/GeneratedOrder.java
lewisbirks: It would be nice to have an internal flag that keeps the order type, I will have a look into it

lewisbirks: Actually no, your way with a few tweaks might be good

e2e/helpers/generated_order_helper.js
lewisbirks: Have changed to
```js
const orderHeading = isOrderRemoved ? `Removed orders ${numberOfOrders}` : `Order ${numberOfOrders}`;
```

KiranVarma02: ahh ok :), i misunderstood that.

e2e/tests/hmctsSuperUserAdministersCase_test.js
akomisarek: Please do, false / true seems as magic values - when reading, no clue what's really being asserted. Such helpers should have more readable variations:
assertFinalOrder
assertOrderHidden

(something along the lines)

service/src/main/java/uk/gov/hmcts/reform/fpl/model/order/generated/GeneratedOrder.java
akomisarek: 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/model/order/generated/GeneratedOrder.java
```diff
@@ -26,4 +30,14 @@
     private final String courtName;
     @JsonSerialize(contentConverter = BasicChildConverter.class)
     private final List<Element<Child>> children;
+    private String removalReason;
+
+    @JsonIgnore
+    public boolean isRemovable() {
+        return GeneratedOrderType.fromType(type).isRemovable();
```
akomisarek, line 37: I believe this can throw an exception, do we need to worry about it, or os it impossible with the current flows to get it? What about the future?

akomisarek (approved):

service/src/main/java/uk/gov/hmcts/reform/fpl/model/order/generated/GeneratedOrder.java
lewisbirks: I believe at the moment it is impossible as I have tested for all possibilities of orders, i.e. interim/final care/supervision, epo, discharge, blank. If there are more added in the future there might be issues but at the moment I don't think there should be.

PR #1551, opened 2020-08-28
fix:Do not take C2 payment when user select not to use PBA

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-1971


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadC2DocumentsController.java
```diff
@@ -116,21 +118,24 @@ public void handleSubmittedEvent(
...
+        } else {
+            if (displayAmountToPay(caseDetails)) {
+                try {
+                    paymentService.makePaymentForC2(caseDetails.getId(), caseData);
+                } catch (FeeRegisterException | PaymentsApiException paymentException) {
+                    log.error("C2 payment for case {} failed", caseDetails.getId());
+                    applicationEventPublisher.publishEvent(new FailedPBAPaymentEvent(callbackRequest, C2_APPLICATION));
+                }
+            } else if (NO.getValue().equals(caseDetails.getData().get(DISPLAY_AMOUNT_TO_PAY))) {
+                log.warn("C2 payment for case {} not taken as payment fee not shown to user", caseDetails.getId());
```
akomisarek, line 136: Just thinking, shouldn't this be an error as well? I believe the warns appear only in a trace, so we could potentially miss it.

akomisarek (approved):

PR #1560, opened 2020-09-02
chore: Add id and state to CaseData

tomaszpowroznik:
- add case id to CaseData
- add state to CaseData
- add CaseConverter to convert CaseDetials to CaseData
limit usage of CaseDetails and CallbackRequest to Controllers only, services and event handlers operates on CaseData
- remove EventData
- abstract CallbackController to provide common utilities


service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -642,4 +643,6 @@ public boolean hasFutureHearing(List<Element<HearingBooking>> hearingBookings) {
             .filter(hearingBooking -> hearingBooking.getStartDate().isAfter(currentCmoStartDate))
             .min(comparing(HearingBooking::getStartDate));
     }
+
+    private String sendToCtsc;
```
KiranVarma02, line 647: nice. I created a sub-task for this and other scenarios like this - 

https://tools.hmcts.net/jira/browse/FPLA-2148

service/src/main/java/uk/gov/hmcts/reform/fpl/events/CaseEvent.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.hmcts.reform.fpl.events;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import uk.gov.hmcts.reform.fpl.model.CaseData;
+
+@Getter
+@EqualsAndHashCode
+public class CaseEvent {
```
lewisbirks, line 9: I don't think that this has been used, are all events supposed to extend it?

tomaszpowroznik: you are right this one is orphaned, will remove

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CoreCaseDataStoreLoader.java
```diff
@@ -13,10 +20,21 @@
 
 public class CoreCaseDataStoreLoader {
 
-    private static final ObjectMapper mapper = new ObjectMapper();
+    private static final ObjectMapper mapper = JsonMapper.builder()
```
akomisarek, line 23: I wonder if we could just inject the ObjectMapper from spring.

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CoreCaseDataStoreLoader.java
```diff
@@ -13,10 +20,21 @@
...
+        .addModule(new ParameterNamesModule())
+        .addModule(new Jdk8Module())
+        .addModule(new JavaTimeModule())
+        .build();
+
+    private static final CaseConverter caseConverter = new CaseConverter(mapper);
 
     private CoreCaseDataStoreLoader() {
-        // NO-OP
+        new ObjectMapper();        // NO-OP
```
akomisarek, line 33: Really, does sonar complains? Even in tests? ;o 

What if there is no constructor, it's test class so I wouldn't worry..

akomisarek (approved):
I don't want to see this change anymore :P. 

Couple of comments, but feel free to change in follow up PR, and as we agreed the PopulateDirectionsHandler could use some refactoring as well to remove even more code.

PR #1577, opened 2020-09-07
chore: added new fingerprints

akomisarek:

N/A

### Change description ###

Adding new fingerprints to allow security build to pass. 

```
[ ] Yes
[ ] No
```


audit.json
```diff
@@ -94,6 +96,7 @@
   "10096_Timestamp Disclosure - Unix_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/callback/add-hearing-bookings/about-to-start_POST": "ignore",
   "10096_Timestamp Disclosure - Unix_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/notify-gatekeeper/about-to-start_POST": "ignore",
   "90033_Loosely Scoped Cookie_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/case-submission/mid-event_POST": "ignore",
+  "90033_Loosely Scoped Cookie_http://fpl-case-service-aat.service.core-compute-aat.internal/callback/remove-order/about-to-start_POST": "ignore",
```
lewisbirks, line 99: Do you wanna add cookie stuff for the new mid event and about to submit as well?

tomaszpowroznik (dismissed):

audit.json
lewisbirks: Just wondering whether we should try and be proactive or reactionary with them

akomisarek: Good point, I was thinking the same as well. I will add.

akomisarek: I think we should be running script just to add everything :P

lewisbirks (approved):

PR #1587, opened 2020-09-08
fix: cmo progression e2e in AAT

lewisbirks:
### Change description ###

Allow for documents to be generated for the sealed cmos so that the doc store urls are valid in AAT.
Select the correct jurisdiction and case type when in AAT (only case type was noticed being wrong but best to double check)

```
[ ] Yes
[x] No
```

### DIFF
--- a/e2e/fixtures/issueResolution.json
+++ b/e2e/fixtures/issueResolution.json
@@ -319,11 +319,7 @@
       {
         "id": "749f29dc-c200-4377-9e12-a5b4bad631b3",
         "value": {
-          "order":{
-            "document_url":"http://dm-store:8080/documents/7196dcc4-dd17-49ba-9924-7480b91e620e",
-            "document_filename":"mockFile.pdf",
-            "document_binary_url":"http://dm-store:8080/documents/7196dcc4-dd17-49ba-9924-7480b91e620e/binary"
-          },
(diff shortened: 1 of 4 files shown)


e2e/fixtures/issueResolution.json
```diff
@@ -319,11 +319,7 @@
       {
         "id": "749f29dc-c200-4377-9e12-a5b4bad631b3",
         "value": {
-          "order":{
-            "document_url":"http://dm-store:8080/documents/7196dcc4-dd17-49ba-9924-7480b91e620e",
-            "document_filename":"mockFile.pdf",
-            "document_binary_url":"http://dm-store:8080/documents/7196dcc4-dd17-49ba-9924-7480b91e620e/binary"
-          },
+          "order":"TO BE FILLED WITH VALID DATA",
```
tomaszpowroznik, line 322: should we just update url of doc with correct value, it should be availbale in config

tomaszpowroznik, line 322: do we need it at all? 
it is overriden anyway

lewisbirks: Was just being consistent with everything else

tomaszpowroznik (approved):

PR #1594, opened 2020-09-09
impr: increase resiliance of draft sdo e2e

lewisbirks:
### Change description ###

Added ability to override test retries at the scenario level

```
[ ] Yes
[ ] No
```

### DIFF
--- a/e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
+++ b/e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
@@ -123,7 +123,7 @@ Scenario('Gatekeeper drafts standard directions', async (I, caseViewPage, draftS
   caseViewPage.selectTab(caseViewPage.tabs.draftOrders);
   I.seeInTab(['Standard directions order', 'File'], 'draft-standard-directions-order.pdf');
   I.seeInTab(['Standard directions order', 'Date of issue'], dateFormat(today, 'd mmmm yyyy'));
-});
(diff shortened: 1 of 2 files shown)


e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
```diff
@@ -123,7 +123,7 @@ Scenario('Gatekeeper drafts standard directions', async (I, caseViewPage, draftS
   caseViewPage.selectTab(caseViewPage.tabs.draftOrders);
   I.seeInTab(['Standard directions order', 'File'], 'draft-standard-directions-order.pdf');
   I.seeInTab(['Standard directions order', 'Date of issue'], dateFormat(today, 'd mmmm yyyy'));
-});
+}).retry(1); // Send letter is async for the hearing details event, if things are running slow then a data altered outside of transaction error may be raised at the end of this scenario.
```
KiranVarma02, line 126: is this easy to read maybe?
`Scenario('name....', { retries: 1 }, (I) => {});`

karthikkannankainos (approved):

e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
lewisbirks: I was indecisive between the 2 and went with what I have in the end because it looks safer in my opinion. Using an objects method rather than what looks to be directly injecting a property

PR #1597, opened 2020-09-10
feat: Cache idam call for system user (FPLA-2189)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2189


akomisarek (approved):

lewisbirks (commented):
I might be missing something but are you going to implement this part
> If the request fails, we could invalidate and retry should log in again, so hopefully that should work 

Currently it seems like we just consider the happy path.

PR #1610, opened 2020-09-11
feat: Access case data from e2e test

tomaszpowroznik:
Usage:

const caseData =  await I.getCase(caseId);
console.log(caseData.case_data.sendToCtsc);

### DIFF
--- a/e2e/helpers/case_helper.js
+++ b/e2e/helpers/case_helper.js
@@ -32,6 +32,14 @@ const post = async (url, data, headers, retry = 2, backoff = 500) => {
   });
 };
 
+const get = async (url, headers) => {
+  return fetch(url, {
+    method: 'GET',
+    headers: headers,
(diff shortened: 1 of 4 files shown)


e2e/helpers/case_helper.js
```diff
@@ -83,6 +91,17 @@ const populateWithData = async (caseId, data) => {
   return post(url, data, headers);
 };
 
+const getCaseData = async (caseId) => {
+  const authToken = await getAuthToken();
```
akomisarek, line 95: Couldn't we use currently logged in token? Use the system user as last resort? Aren't the users perms enough?

tomaszpowroznik: i think sys user is better, we want as much filed as possible and not restricted to the actual user.

lewisbirks:
no longer required

PR #1618, opened 2020-09-14
feat: Order alphabetically list of local authorities (FPLA-2188)

tomaszpowroznik:
LAs
https://tools.hmcts.net/jira/browse/FPLA-2188


nkoranne (commented):
Just for my reference : does all future LAs needs to be inserted manually based on the alphabetical order?  and also do we need to change DisplayOrder every time when we insert any LA in between?

tomaszpowroznik:
> Just for my reference : does all future LAs needs to be inserted manually based on the alphabetical order? and also do we need to change DisplayOrder every time when we insert any LA in between?

yes, should be added manually in correct place. Yes reindexing will be required, unless we want to do gaps in numbering (e.g 5)

nkoranne (dismissed):

joez-f (approved):

nkoranne (approved):

ccd-definition/FixedLists/CareSupervision/AuthorityFixedList.json
```diff
@@ -2,115 +2,134 @@
...
+    "ListElementCode": "ERY",
+    "ListElement": "East Riding of Yorkshire Council",
+    "DisplayOrder": 10
   },
   {
     "LiveFrom": "01/01/2017",
     "ID": "AuthorityFixedList",
-    "ListElementCode": "STF",
-    "ListElement": "Staffordshire County Council"
+    "ListElementCode": "FPLA",
```
lewisbirks, line 19: I know that it was already in there but should this be here, does it describe us or is it for something else? If it is us would it not make more sense for it to be in a -nonprod file.

tomaszpowroznik: @lewisbirks , confirmed with Adam, this is legacy org used long time ago to do smoke testing in prod. Currently not used. Cannot submit case for such org, see fpl:
  local_authority_codes_forbidden_case_submission: FPLA

PR #1634, opened 2020-09-16
chore: Group access control - part 1

tomaszpowroznik:
instead of UserRole new elements can be used:
UserRoles: ["role1", "role2"]

AccessControl:[
{
UserRoles: ["role1","role2"],
CRUD: "C"
},
{
UserRoles: ["role3"],
CRUD: "D"
}
]

### DIFF
--- a/ccd-definition/AuthorisationCaseEvent/CareSupervision/AuthorisationCaseEvent.json
+++ b/ccd-definition/AuthorisationCaseEvent/CareSupervision/AuthorisationCaseEvent.json
@@ -17,43 +17,46 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseEventID": "deleteApplication",
-    "UserRole": "[LASOLICITOR]",
-    "CRUD": "CRU"
-  },
-  {
-    "LiveFrom": "01/01/2017",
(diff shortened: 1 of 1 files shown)


lewisbirks (commented):
`ccd-definition/AuthorisationCaseField/CareSupervision/fileds.json` -> `ccd-definition/AuthorisationCaseField/CareSupervision/fields.json`

Also what sort of organisation do you have in mind for this format? It would probably be nicer and easier to search for stuff if we had them split and not all in one file.

tomaszpowroznik:
> `ccd-definition/AuthorisationCaseField/CareSupervision/fileds.json` -> `ccd-definition/AuthorisationCaseField/CareSupervision/fields.json`
> 
> Also what sort of organisation do you have in mind for this format? It would probably be nicer and easier to search for stuff if we had them split and not all in one file.

I think it is more natural to define something and saying who can access it. Also i think that in most cases we are interested in information who has access to filed, and not what filed particular user has access to. But understand your point, i can restore filed configuration and focus on events

lewisbirks (approved):
I am also happy for us to eventually extend it to fields, I would just like it to be nice and organised when we do so. Maybe we can organise it by what event they are used in or something like that. I don't want to run into the same situation we have with the current caseField.json where we have almost all the fields in one file so it is a bit harder to find an individual one.

PR #1639, opened 2020-09-17
chore: zap exceptions

tomaszpowroznik:


lewisbirks (approved):
It's annoying that this didn't appear in the last scan, the endpoint was there so it should have picked it up then

PR #1644, opened 2020-09-18
feat: Add zap hook to disable rule (FPLA-2224)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2224


lewisbirks (commented):
Do we want to test this on the nightly Dev branch first?

zap-hooks.py
```diff
@@ -0,0 +1,5 @@
+def zap_started(zap, target):
+  zap.pscan.set_scanner_alert_threshold(id='90033', alertthreshold='OFF')
+  zap.pscan.set_scanner_alert_threshold(id='10023', alertthreshold='OFF')
+  zap.pscan.set_scanner_alert_threshold(id='10010', alertthreshold='OFF')
+  zap.pscan.set_scanner_alert_threshold(id='10054', alertthreshold='OFF')
```
lewisbirks, line 5: Do you think something like this is easier to extend or do you think that the additional clarity provided by listing each out is better?
```suggestion
def zap_started(zap, target):
  rules_to_ignore = ('90033', '10023', '10010', '10054')
  for rule in rules_to_ignore:
    zap.pscan.set_scanner_alert_threshold(id=rule, alertthreshold='OFF')
```

tomaszpowroznik: yeah, why not, done

lewisbirks (approved):
LGTM unless anyone wants any other rule ignored by default

PR #1647, opened 2020-09-18
Fpla 1983 postgress upgrade v11

nkoranne:
https://tools.hmcts.net/jira/browse/FPLA-1983

### Change description ###
changed a cron job for frequent db calls to check postgreSQL v11 upgrade working fine or not

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/main/resources/application.yaml
+++ b/service/src/main/resources/application.yaml
@@ -132,7 +132,8 @@ scheduler:
       cronGroup: "NIGHTLY_CRON"
       description: 'Finds upcoming hearings'
       jobClass: 'uk.gov.hmcts.reform.fpl.jobs.UpcomingHearingsFinder'
-      cron: ${UPCOMING_HEARINGS_CRON:0 0 2 ? * MON-FRI *}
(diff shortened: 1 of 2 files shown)


service/src/main/resources/application.yaml
```diff
@@ -132,7 +132,8 @@ scheduler:
       cronGroup: "NIGHTLY_CRON"
       description: 'Finds upcoming hearings'
       jobClass: 'uk.gov.hmcts.reform.fpl.jobs.UpcomingHearingsFinder'
-      cron: ${UPCOMING_HEARINGS_CRON:0 0 2 ? * MON-FRI *}
+      cron: ${UPCOMING_HEARINGS_CRON:0 0/15 * ? * MON-FRI *}
```
tomaszpowroznik, line 135: it is not needed now, upgrade was done on Friday, so there should be already job run. If we still need to change cron expression it should be rather done via UPCOMING_HEARINGS_CRON in flux

nkoranne: ok. Thanks Tomek ... I will close this PR

PR #1650, opened 2020-09-21
impr: update smoke test and clean some e2e steps

lewisbirks:

FPLA-N/A

### Change description ###

Update the smoke test to actaully touch the system.
Attempt to reduce the number of logins during the e2e run time and navigation to other pages to reduce possible failures.
Clean up gradle file during exploration of smoke test.

```
[ ] Yes
[x] No
```


e2e/tests/gatekeeperAdministersCaseAfterGatekeeping_test.js
```diff
@@ -60,7 +60,7 @@ Scenario('Gatekeeper enters allocation decision', async (I, caseViewPage, enterA
   I.seeEventSubmissionConfirmation(config.applicationActions.enterAllocationDecision);
 });
 
-Scenario('Gatekeeper enters hearing details and submits', async (I, caseViewPage, loginPage, addHearingBookingDetailsEventPage) => {
```
lewisbirks, line 63: Do we need to do this test here? We do the same test in `e2e/tests/hmctsAdministersCaseAfterSubmission_test.js:161`. Should one of them be removed?

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -13,32 +13,37 @@ const c2Payment = require('../fixtures/c2Payment.js');
...
 
 BeforeSuite(async (I) => {
   caseId = await I.submitNewCaseWithData(mandatoryWithMultipleChildren);
   submittedAt = new Date();
-
-  await I.navigateToCaseDetailsAs(config.hmctsAdminUser, caseId);
 });
 
-Before(async I => await I.navigateToCaseDetailsAs(config.hmctsAdminUser, caseId));
+Before(async I => {
```
tomaszpowroznik, line 30: tbh, I dont like this one. Looks like performance improvement. E2e should be really ease to read, thats why all of this I.see something. Should feel like plain english. It can cause problem if dev want to skip some tests.

lewisbirks: Yeah this is the one that I was not sure about either, I wanted to reduce the number of times logging in so I first just changed it from `navigate...As` to `navigate...` but then I noticed that some of the tests use switch users to la users. I guess a better way would be for the `steps_file.js` to remember the current logged in user and then if it is the same user not log in again.

e2e/steps_file.js
```diff
@@ -153,6 +153,15 @@ module.exports = function () {
       this.dontSeeElement(caseListPage.locateCase(normalizeCaseId(caseId)));
     },
 
+    async seeEndStateForEvent(eventName, state) {
+      try {
+        await this.waitForSelector(`//tr[@class="EventLogTable-Selected" and td[contains(., "${eventName}")]]`);
+      } catch (notFound) {
+        this.click(`//table[@class="EventLogTable"]//tr[td[contains(., "${eventName}")]][1]`);
```
tomaszpowroznik, line 160: cannot we always click on a row and then check? it looks dodgy a bit with catch

lewisbirks: probably, will give it a go. Myself and joe also thought it looked dodgy.

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
tomaszpowroznik: yeah, lets keep such improvement for later maybe

e2e/steps_file.js
lewisbirks: do you think it would be worth adding an optional index parameter? Speaking of which the index should probably be `[0]` not `[1]`

lewisbirks: turns out I am wrong on that last part, xpath has a 1 based index

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
lewisbirks: sure thing, will revert the change here and then open another PR later for that

tomaszpowroznik (approved):

PR #1666, opened 2020-09-23
impr: reduce sign ins on e2e

lewisbirks:

FPLA-N/A

### Change description ###

Store the current user and check if they are the same as the user currently trying to sign in.
If they are then don't sign in.

```
[ ] Yes
[x] No
```


e2e/steps_file.js
```diff
@@ -15,6 +15,7 @@ const normalizeCaseId = caseId => caseId.toString().replace(/\D/g, '');
 const baseUrl = process.env.URL || 'http://localhost:3333';
 const signedInSelector = 'exui-header';
 const signedOutSelector = '#global-header';
+let currentUser = {};
```
tomaszpowroznik, line 18: hmm, how does it works when multiple executors?

lewisbirks: seems to work fine, was worried about it as well but it never seemed to log in unnecessarily. Will run again though to check.

tomaszpowroznik: it is a bit slippery, assumes that all login/logout happens via function signIn from this file

tomaszpowroznik: https://codecept.io/parallel/

should be fine as long as we use run-multiple

e2e/steps_file.js
```diff
@@ -25,19 +26,25 @@ function log(msg) {
 module.exports = function () {
   return actor({
     async signIn(user) {
-      await this.retryUntilExists(async () => {
-        this.amOnPage(baseUrl);
+      if (currentUser !== user) {
+        this.say(`Logging in as ${user.email}`);
```
tomaszpowroznik, line 30: maybe we should do currentUser={} here, just in case that entire process fail in the middle?
user logout but did not manage to log in

lewisbirks: makes sense, especially as that may happen in the nightly build.

e2e/steps_file.js
```diff
@@ -15,29 +15,34 @@ const normalizeCaseId = caseId => caseId.toString().replace(/\D/g, '');
 const baseUrl = process.env.URL || 'http://localhost:3333';
 const signedInSelector = 'exui-header';
 const signedOutSelector = '#global-header';
+let currentUser = {};
 
 'use strict';
 
-function log(msg) {
```
lewisbirks, line 21: See https://codecept.io/hooks/#output

tomaszpowroznik (approved):

PR #1667, opened 2020-09-23
chore: Run subset of e2e in aat functional tests stage

tomaszpowroznik:
https://codecept.io/advanced/#tags

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -136,7 +136,7 @@ withPipeline(type, product, component) {
   }
 
   before('functionalTest:aat') {
-    env.SKIP_FUNCTIONAL_TESTS = 'true'
+    env.TESTS_SELECTOR="@smoke-tests"
   }
 
   before('functionalTest:demo') {
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "lint": "eslint e2e",
     "typings": "codeceptjs def && mv steps.d.ts typings/codeceptjs.d.ts",
     "test": "codeceptjs run-multiple parallel",
-    "test:functional": "MOCHAWESOME_REPORTFILENAME=functional codeceptjs run-multiple parallel --invert --grep @smoke-tests --reporter mocha-multi",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "lint": "eslint e2e",
     "typings": "codeceptjs def && mv steps.d.ts typings/codeceptjs.d.ts",
     "test": "codeceptjs run-multiple parallel",
-    "test:functional": "MOCHAWESOME_REPORTFILENAME=functional codeceptjs run-multiple parallel --invert --grep @smoke-tests --reporter mocha-multi",
+    "test:functional": "MOCHAWESOME_REPORTFILENAME=functional codeceptjs run-multiple parallel --grep \"${TESTS_SELECTOR:=(?=.*)^(?!.*@smoke-tests)}\" --reporter mocha-multi",
```
tomaszpowroznik, line 25: https://codecept.io/advanced/#tags
last example how to exclude test

KiranVarma02 (approved):

Jenkinsfile_CNP
```diff
@@ -136,7 +136,7 @@ withPipeline(type, product, component) {
   }
 
   before('functionalTest:aat') {
-    env.SKIP_FUNCTIONAL_TESTS = 'true'
+    env.TESTS_SELECTOR="@smoke-tests"
```
lewisbirks, line 139: Would this not just be doing the same as the smoke test step? It goes deploy aat -> smoke aat -> functional aat right?

tomaszpowroznik: yes, it will do the same what smoke test. We just need to run anything there, agreed with Adam and Robin that we run there smoke test for now

lewisbirks: do we want to have an aat tag which we can then add to the tests which we can then use in the future?

tomaszpowroznik: we can add it when there will be need, for now we dont even know if we will be running e2e or api tests. It is to have something there to not be marked in red in program reports

PR #1668, opened 2020-09-23
chore: Generate ccd definition from perftest branch

tomaszpowroznik:
chore: Generate ccd definition from perftest branch

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -172,6 +172,10 @@ withPipeline(type, product, component) {
     }
   }
 
+  before('buildinfra:perftest') {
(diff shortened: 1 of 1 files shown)


lewisbirks (dismissed):

bin/variables/load-perftest-environment-variables.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
lewisbirks, line 1: Don't think this file is necessary, `build-release-ccd-definition.sh` contains
```sh
export CCD_DEF_CASE_SERVICE_BASE_URL=http://fpl-case-service-${environment}.service.core-compute-${environment}.internal
```

PR #1689, opened 2020-09-29
Bug/upload replacement sdo bug fix

KiranVarma02:

No Jira

### Change description ###

**Scenario**:
Upload SDO and Select option  No, just save it on the system

Then try to do the same event for proceed to final SDO then you will see this issue

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderService.java
@@ -16,6 +16,7 @@ import uk.gov.hmcts.reform.idam.client.models.UserInfo;
 import java.time.LocalDate;
 
 import static org.apache.commons.lang3.ObjectUtils.defaultIfNull;
+import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
 import static uk.gov.hmcts.reform.fpl.utils.DateFormatterHelper.DATE;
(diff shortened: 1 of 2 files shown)


service/src/test/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderServiceTest.java
```diff
@@ -96,6 +96,20 @@ void shouldUseReplacementSDOForTemporaryStandardDirectionOrder() {
         assertThat(order).isEqualTo(expectedOrder);
     }
```
KiranVarma02, line 98: Could you please check what above test is doing? is it valid?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderService.java
```diff
@@ -44,7 +45,8 @@ public StandardDirectionOrder buildTemporarySDO(CaseData caseData) {
             // been through once, either pull from replacement doc or SDO if that isn't present
             document = defaultIfNull(
                 caseData.getReplacementSDO(),
-                caseData.getStandardDirectionOrder().getOrderDoc()
+                isNotEmpty(caseData.getStandardDirectionOrder())
```
lewisbirks, line 48: I am worried about what will be happening in the scenario that a user goes through the event once uploading a document then goes through again but doesn't upload a replacement document. From my understanding of the bug this will just mean that the new SDO is populated with a null document instead of the previous one

tomaszpowroznik:
fix provided in https://github.com/hmcts/fpl-ccd-configuration/pull/1691/files

PR #1691, opened 2020-09-29
fix: sdo regression

lewisbirks:

FPLA-N/A

### Change description ###

Use case details before instead of current case data. CCD now removes fields that appear on the next screen after a mid event (see [ccd-data-store](https://github.com/hmcts/ccd-data-store-api/blob/24b7e06f8b195867e03e7909bda94662c9d5a14d/src/main/java/uk/gov/hmcts/ccd/domain/service/createevent/MidEventCallback.java#L78))

#### Scenarios
- goes through event once
- goes though event twice + uploads new on second pass
- goes though event twice + keeps current doc on second pass (nothing uploaded)

Will need to check if anything breaks with old SDO flow; hopefully we can trust e2e though, it spotted this issue for us.

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AbstractControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AbstractControllerTest.java
@@ -236,6 +236,13 @@ abstract class AbstractControllerTest {
             .build();
     }
 
+    CallbackRequest toCallBackRequest(CaseDetails caseDetails, CaseDetails caseDetailsBefore) {
+        return CallbackRequest.builder()
+            .caseDetails(caseDetails)
+            .caseDetailsBefore(caseDetailsBefore)
+            .build();
+    }
+
     LocalDateTime now() {
         return time.now();
     }
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderService.java
@@ -37,14 +37,14 @@ public class StandardDirectionsOrderService {
         return dateOfIssue;
     }
 
-    public StandardDirectionOrder buildTemporarySDO(CaseData caseData) {
+    public StandardDirectionOrder buildTemporarySDO(CaseData caseData, StandardDirectionOrder previousSDO) {
         DocumentReference document = caseData.getPreparedSDO();
 
         if (document == null) {
             // been through once, either pull from replacement doc or SDO if that isn't present
             document = defaultIfNull(
                 caseData.getReplacementSDO(),
-                caseData.getStandardDirectionOrder().getOrderDoc()
+                previousSDO.getOrderDoc()
             );
         }
         return StandardDirectionOrder.builder()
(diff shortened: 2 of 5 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderService.java
```diff
@@ -37,14 +37,14 @@ public LocalDate generateDateOfIssue(StandardDirectionOrder order) {
...
-    public StandardDirectionOrder buildTemporarySDO(CaseData caseData) {
+    public StandardDirectionOrder buildTemporarySDO(CaseData caseData, StandardDirectionOrder previousSDO) {
         DocumentReference document = caseData.getPreparedSDO();
 
         if (document == null) {
             // been through once, either pull from replacement doc or SDO if that isn't present
             document = defaultIfNull(
                 caseData.getReplacementSDO(),
-                caseData.getStandardDirectionOrder().getOrderDoc()
+                previousSDO.getOrderDoc()
```
tomaszpowroznik, line 47: alternatively it could use caseData.currentSDO which should have actual doc

service/src/test/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderServiceTest.java
```diff
@@ -98,15 +99,16 @@ void shouldUseReplacementSDOForTemporaryStandardDirectionOrder() {
 
     @Test
     void shouldUseCurrentSDODocumentForTemporaryStandardDirectionOrder() {
+        StandardDirectionOrder previousSDO = StandardDirectionOrder.builder()
+            .orderDoc(SEALED_DOC)
```
tomaszpowroznik, line 103: it is a bit misleading, at this point sdo should not be sealed yet, maybe just WORD_DOC

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AbstractControllerTest.java
```diff
@@ -236,6 +236,13 @@ CaseDetails asCaseDetails(CaseData caseData) {
             .build();
     }
 
+    CallbackRequest toCallBackRequest(CaseDetails caseDetails, CaseDetails caseDetailsBefore) {
```
tomaszpowroznik, line 239: maybe it would be a bit more benefitial to have overloaded method that accepts CaseData as both params

tomaszpowroznik (approved):

service/src/test/java/uk/gov/hmcts/reform/fpl/service/sdo/StandardDirectionsOrderServiceTest.java
```diff
@@ -98,15 +99,16 @@ void shouldUseReplacementSDOForTemporaryStandardDirectionOrder() {
 
     @Test
     void shouldUseCurrentSDODocumentForTemporaryStandardDirectionOrder() {
+        StandardDirectionOrder previousSDO = StandardDirectionOrder.builder()
```
tomaszpowroznik, line 102: maybe method name should be shouldUsePreviouslyUploadedSDO, now it is confusing that tests says current but it is previous doc

PR #1696, opened 2020-09-30
fix: fix for the hearing booking bug

toireasam:
https://tools.hmcts.net/jira/browse/FPLA-2249

### Change description ###
Changing e2e test data to reference new test data as old was causing issue with creating hearing bookings as old test data did not contain sdo directions

```
[ ] Yes
[x] No
```


e2e/fixtures/caseData/gatekeepingNoHearingDetails.json
```diff
@@ -38,68 +38,6 @@
       "reference": "reference",
       "telephone": "00000000000"
     },
-    "allParties": [
```
lewisbirks, line 41: please revert as discussed

service/src/main/java/uk/gov/hmcts/reform/fpl/service/StandardDirectionsService.java
```diff
@@ -47,6 +47,7 @@ public boolean hasEmptyDates(CaseData caseData) {
                 caseData.getCafcassDirections(),
                 caseData.getOtherPartiesDirections(),
                 caseData.getCourtDirections())
+            .filter(Objects::nonNull)
```
lewisbirks, line 50: As discussed I don't think this is the correct solution, the issue should be caused else where as it should not be possible to get here without having these directions, which would be populated in the notify gatekeeper event

lewisbirks: As discussed we have worked out that the issue is some bad test data. Please update that test data and then everything should be fine.

service/src/test/java/uk/gov/hmcts/reform/fpl/service/StandardDirectionsServiceTest.java
```diff
@@ -91,6 +91,20 @@ void shouldReturnFalseWhenThereAreNoEmptyDates() {
         assertThat(service.hasEmptyDates(caseData)).isFalse();
     }
 
+    @Test
+    void shouldReturnFalseWhenCafcassDirectionsAreNull() {
```
lewisbirks, line 95: Can be reverted

lewisbirks (changes requested):
FYI @samuelzerozero if you are interested

service/src/main/java/uk/gov/hmcts/reform/fpl/service/StandardDirectionsService.java
toireasam: changed based on discussion

lewisbirks (approved):

PR #1707, opened 2020-10-02
chore: support running e2e on demo environment

tomaszpowroznik:


lewisbirks (commented):
docker stuff added by mistake?

tomaszpowroznik:
> docker stuff added by mistake?

if you asking about bin/build-release-ccd-definition.sh it is on purpose. This script is used to create definition files for envs. On demo we want to include definitions that support prepopulation of case, but on prod not. On demo and prod we want to have same toggles, thats why need to distinguish toggles and testing support definitions

lewisbirks (approved):
It logs in for me 🥳

PR #1723, opened 2020-10-07
chore: give system update "U" permissions to risks (FPLA-2268)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-2268

### Change description ###
DM on production failed to - 
2020-10-06T13:34:29.857 ERROR [main] u.g.hmcts.reform.migration.CaseMigrationProcessor Case 1600945960951828 update failed due to: status 404 reading CoreCaseDataApi#submitEventForCaseWorker(String,String,String,String,String,String,boolean,CaseDataContent)

This was due to 
ccd-data-store-api_1              | 2020-10-07T10:02:27.821 INFO  [http-nio-4452-exec-2] u.g.h.c.d.s.common.CompoundAccessControlService Simple child physicalHarm of risks has data update but no Update ACL

PR updates system user's access to the risks field. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/ccd-definition/AuthorisationCaseField/CareSupervision/system-update.json
+++ b/ccd-definition/AuthorisationCaseField/CareSupervision/system-update.json
@@ -130,7 +130,7 @@
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "risks",
     "UserRole": "caseworker-publiclaw-systemupdate",
-    "CRUD": "R"
+    "CRUD": "RU"
   },
   {
     "LiveFrom": "01/01/2017",


lewisbirks:
Why would it be trying to update risks? I thought you were just remove the sealed cmo and editing the linked hearing?

Jamesnelson117:
@lewisbirks when ran locally it looks like its insisting on removing two null fields within risks, here is old case data
{"risks": {"neglect": null, "sexualAbuse": "Yes", "physicalHarm": null, "emotionalHarm": "Yes", "sexualAbuseOccurrences": [], "emotionalHarmOccurrences": []},

new
{"risks": {"sexualAbuse": "Yes", "emotionalHarm": "Yes", "sexualAbuseOccurrences": [], "emotionalHarmOccurrences": []},

lewisbirks:
@Jamesnelson117 wonder why we haven't had this issue before then 🤔

tomaszpowroznik:
yeah, this change should not be needed. Looks that data migration could be written in wrong way. 
The method  
public Object migrate(Map<String, Object> data)
is supposed to return event data only and not entire case (only delta)
As we use our own callback and rely on updating case from there then it should return empty map and not entire map of all properties

PR #1729, opened 2020-10-08
refactor: refactor of courtseal and draftbackground (FPLA-2258)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-2258

### Change description ###
Refactors c110a, notice of proceedings and notice of hearing to use 'DocmosisTemplateImages' model class. 
C110A template changes.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisTemplateImages.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmcts.reform.fpl.model.docmosis;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class DocmosisTemplateImages {
```
lewisbirks, line 8: Thoughts on extending this to cover all images in `uk/gov/hmcts/reform/fpl/enums/DocmosisImages.java`

lewisbirks, line 8: Thoughts on expanding this slightly, instead of having a builder we "delombok" that part and we then change the signatures so that they can be something like
`.withDraftWaterMark()` which then points to the draft watermark constant (which could then be removed from the enum and made a constant property in this class)

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisTemplateImages.java
```diff
@@ -0,0 +1,12 @@
...
+    private final String draftWaterMark;
+    private final String courtseal;
```
lewisbirks, line 10: Court seal is really two words, I can understand not wanting to change in the templates but we can use the json property here to specify the name whilst maintaining camel case in the java model

PR #1760, opened 2020-10-16
feat: hearing adjournment and re-listing (FPLA-1916)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-1916


joez-f (commented):
Looks OK, only couple of comments so far, nothing major - will look again in a bit (someone else should also look through if available @samuelzerozero @lewisbirks 😄)

service/src/main/java/uk/gov/hmcts/reform/fpl/service/IdentityService.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.reform.fpl.service;
+
+import org.springframework.stereotype.Service;
+
+import java.util.UUID;
+
+@Service
+public class IdentityService {
```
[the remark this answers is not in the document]
tomaszpowroznik: it is only to help with testing. Currently we dont have option to spy on static UUID.randomUUID and in many places we cannot just use equals to verify that element is what expected. This can be spied/mocked and in a test we can easily check element object (and verify that it got e.g brand new id)
see ManageHearingsServiceTest.shouldAdjournAndReListHearingWithoutDocumentReassignment

maybe the correct option is to use powermock to have option to spy on static members

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -479,13 +513,34 @@ public boolean hasFutureHearing(List<Element<HearingBooking>> hearingBookings) {
             .collect(toList());
     }
 
+    @JsonIgnore
+    public List<Element<HearingBooking>> getPastAndTodayHearings() {
+        return defaultIfNull(hearingDetails, new ArrayList<Element<HearingBooking>>()).stream()
+            .filter(hearingBooking -> !hearingBooking.getValue().getStartDate().toLocalDate().isAfter(LocalDate.now()))
```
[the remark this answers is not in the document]
tomaszpowroznik: done

ccd-definition/AuthorisationCaseField/CareSupervision/court-admin.json
```diff
@@ -1383,6 +1404,27 @@
...
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "adjournmentReason",
+    "UserRole": "caseworker-publiclaw-courtadmin",
+    "CRUD": "CRUD"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "hearingReListOption",
```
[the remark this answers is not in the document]
tomaszpowroznik: looks that you are right, it is single word, but deign team uses re-list, and jira and intellij mark relisted as typo

service/src/main/java/uk/gov/hmcts/reform/fpl/service/IdentityService.java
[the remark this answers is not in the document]
samuelzerozero: Using static methods is always fishy IMHO, and wrap them whenever possible. If there are cases anyway when we cannot get rid of them (the wrapped implementation for example) there's now support for that since mockito-inline 3.4.4

https://github.com/hmcts/fpl-ccd-configuration/blob/466e403464d4cb8c039b048a57d6a3ed7f81262b/service/src/test/java/uk/gov/hmcts/reform/fpl/model/order/generated/GeneratedOrderTest.java#L34

https://github.com/hmcts/fpl-ccd-configuration/blob/2d2d9d0682c9e03ac2e0caa102ce909fc2684863/service/build.gradle#L246

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsMapTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.jupiter.api.Test;
+import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
+
+import java.util.Map;
+
+import static java.util.Collections.emptyList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.hmcts.reform.fpl.utils.CaseDetailsMap.caseDetailsMap;
+
+class CaseDetailsMapTest {
```
samuelzerozero, line 12: Missing the case when an existing key/value is overridden with a non null/emptyList?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/IdentityServiceTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.Set;
+import java.util.UUID;
+
+import static java.util.stream.Collectors.toSet;
+import static java.util.stream.IntStream.rangeClosed;
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+
+class IdentityServiceTest {
```
samuelzerozero, line 12: could use the static mock here

service/src/test/java/uk/gov/hmcts/reform/fpl/service/EventServiceTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import static org.mockito.Mockito.verify;
+
+@ExtendWith(MockitoExtension.class)
+class EventServiceTest {
+
+    @Mock
+    private ApplicationEventPublisher applicationEventPublisher;
+
+    @InjectMocks
```
samuelzerozero, line 18: To distinguish the object that is tested in the class I'd call that testObject or underTest instead of eventService

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsMapTest.java
tomaszpowroznik: ManageHearingsController.handleAboutToSubmit

samuelzerozero (approved):

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsMapTest.java
samuelzerozero: Having to rely on a test of a class that uses the component is a bit difficult to trace and it breaks a bit the model of unit testing. If refactoring in the future, you removed ManageHearingsController the code branch would be uncovered. Not covering at unit level prevent you as well to make sure that you could actually use a mock() of that object. Happy to discuss the approach.

tomaszpowroznik: sorry, I misunderstand you, though that you were asking where new class is used. Missing test  added

service/src/main/java/uk/gov/hmcts/reform/fpl/model/HearingBooking.java
```diff
@@ -59,13 +65,19 @@ public boolean startsAfterToday() {
         return startDate.isAfter(ZonedDateTime.now(ZoneId.of("Europe/London")).toLocalDateTime());
```
samuelzerozero, line 65: That's probably a tech debt story, but for stuff like this we should have our own Static class with its own clock. In this way we can have tests with stopped clock rather than using now(). Probably unlikely, but if the tests run at midnight, that could expose a flake

samuelzerozero (dismissed):

service/src/main/java/uk/gov/hmcts/reform/fpl/model/HearingBooking.java
tomaszpowroznik: we have it, see FixedTimeConfiguration, here it really does not matter and midnight should not afect as this operates on datetime and not on a date

nraghu32 (approved):

PR #1762, opened 2020-10-16
NOJIRA: fix IllegalArgumentException: No enum constant uk.gov.hmcts.r…

samuelzerozero:
…eform.fpl.enums.GeneratedOrderType

- [ x] tests have been updated / new tests has been added (if needed)

### Change description ###

```
[ ] Yes
[ x] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/enums/GeneratedOrderType.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/enums/GeneratedOrderType.java
@@ -1,8 +1,11 @@
 package uk.gov.hmcts.reform.fpl.enums;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.microsoft.applicationinsights.boot.dependencies.apachecommons.lang3.EnumUtils;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
+
+import java.util.Optional;
 
 @Getter
 @RequiredArgsConstructor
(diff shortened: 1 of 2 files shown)


lewisbirks (dismissed):

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/GeneratedOrderType.java
```diff
@@ -21,7 +21,11 @@ public static GeneratedOrderType fromType(String type) {
         type = type.replaceAll("(Final|Interim|\\(C21\\))", "").strip();
         type = type.toUpperCase();
         type = type.replace(" ", "_");
-        return GeneratedOrderType.valueOf(type);
+        try {
```
tomaszpowroznik, line 24: could be
return ofNullable(EnumUtils.getEnum(GeneratedOrderType.class, type)).orElse(UPLOAD);
as this should not be based on exception

tomaszpowroznik (approved):

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/GeneratedOrderType.java
samuelzerozero: the implementation of EnumUtils is actually the catch of the exception
    public static <E extends Enum<E>> E getEnum(Class<E> enumClass, String enumName) {
        if (enumName == null) {
            return null;
        } else {
            try {
                return Enum.valueOf(enumClass, enumName);
            } catch (IllegalArgumentException var3) {
                return null;
            }
        }
    }

lewisbirks: Wasn't aware of that class, if we want to use that method it could just be `EnumUtils.getEnum(GeneratedOrderType.class, type, UPLOAD)`

samuelzerozero: reads better on your option anyway, I like it more. Pushed the change

tomaszpowroznik: and we all learnt new thing 👍

tomaszpowroznik (approved):

PR #1783, opened 2020-10-23
feat: Introduce go to next page function

tomaszpowroznik:


e2e/steps_file.js
```diff
@@ -288,6 +289,19 @@ module.exports = function () {
...
 
+    async goToNextPage(nextButtonLabel = 'Continue' ){
+      const currentUrl = await this.grabCurrentUrl();
+      for (let tryNumber = 1; tryNumber <= maxRetries; tryNumber++) {
+        this.click(nextButtonLabel);
+        if(await this.grabCurrentUrl() !== currentUrl){
+          break;
+        } else {
+          output.log(`Navigation attempt failed for ${tryNumber} time(s)`);
+          this.wait(1);
```
lewisbirks, line 300: Would increase this as we know that aat/some doc generation can be quite slow. Maybe 2s instead

lewisbirks: Or maybe make the max number of retries configurable so it can be increased for aat

PR #1805, opened 2020-10-29
chore: FPLA-2360 Remove SDO for SN20C50019

aliveni:
https://tools.hmcts.net/jira/browse/FPLA-2360

### Change description ###
Remove SDO for SN20C50019

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/MigrateCaseController.java
```diff
@@ -34,31 +30,16 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
         Map<String, Object> data = caseDetails.getData();
         CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
 
-        if ("SA20C50034".equals(caseData.getFamilyManCaseNumber()) && 1600893083735118L == caseDetails.getId()) {
-            data.put("hearingDetails", removeHearingLinkedToCmo(caseData.getHearingDetails(),
-                caseData.getSealedCMOs().get(0).getId()));
+        if ("SN20C50019".equals(caseData.getFamilyManCaseNumber()) && 1603717767912577L == caseDetails.getId()) {
+            data.remove("standardDirectionOrder");
+            data.remove("noticeOfProceedingsBundle");
```
Jamesnelson117, line 35: Hmm, do we definitely want to remove NOP's? As they were created via a separate event prior to creating SDO. I think we are right in removing them but it's worth checking with Gail / James to confirm this is the right thing to do.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/MigrateCaseController.java
```diff
@@ -34,31 +30,16 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
         Map<String, Object> data = caseDetails.getData();
         CaseData caseData = mapper.convertValue(caseDetails.getData(), CaseData.class);
 
-        if ("SA20C50034".equals(caseData.getFamilyManCaseNumber()) && 1600893083735118L == caseDetails.getId()) {
-            data.put("hearingDetails", removeHearingLinkedToCmo(caseData.getHearingDetails(),
-                caseData.getSealedCMOs().get(0).getId()));
+        if ("SN20C50019".equals(caseData.getFamilyManCaseNumber()) && 1603717767912577L == caseDetails.getId()) {
```
Jamesnelson117, line 33: Checking both is probably a bit unnecessary moving forward as we can guarantee that CCD ID is unique to case. But happy if you feel safer with both considering we are modifying prod data.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/MigrateCaseController.java
aliveni: updated to remove just the standardDirectionOrder from case data

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/MigrateCaseController.java
aliveni: Removed condition on FamilyManCaseNumber

nkoranne (commented):
LGTM

nkoranne (approved):

PR #1820, opened 2020-11-02
chore: Use gotToNextPage function

tomaszpowroznik:


toireasam (approved):
LGTM

karthikkannankainos (commented):
In the previous implementation, after continue is clicked, they wait for a location. But in this new implementation we are missing finding the locator. is this expected?

PR #1823, opened 2020-11-02
chore: remove perf test support as not used

tomaszpowroznik:


service/src/main/resources/application.yaml
```diff
@@ -359,7 +359,7 @@ robotics:
     recipient: FamilyPublicLaw+robotics@gmail.com
 
 scheduler:
-  enabled: false
```
samuelzerozero, line 362: is this wanted or a typo?

tomaszpowroznik: good spot, was testing something and this slip in

tomaszpowroznik: done

samuelzerozero (approved):

PR #1801, opened 2020-10-28
Fpla 2348 change of address

nkoranne:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/FPLA-2348

### Change description ###
Changed address for Portsmouth Magistrates Court and added new entry for Newcastle Hearing Centre

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ x] No
```

PR #1839, opened 2020-11-05
chore: FPLA-2365 Remove Newcastle Combined Court hearing venue

aliveni:
https://tools.hmcts.net/jira/browse/FPLA-2365

### Change description ###
Removed incorrect hearing venue "Newcastle Combined Court"

```
[ ] Yes
[x] No
```


lewisbirks (commented):
There is also the entry in the ccd fixed list. I would be careful about doing removing that though, and would first check if any cases in production use that entry and if there are we would need to consider data migration

service/src/main/resources/static_data/hearingVenues.json
```diff
@@ -2429,16 +2429,6 @@
       "PostCode": "NE1 6PA"
     }
   },
-  {
```
tomaszpowroznik, line 2432: is it used on prod? what will happen if used?

tomaszpowroznik: select distinct reference from case_data c cross join jsonb_array_elements(c.data->'hearingDetails') ids(hearings)
where (hearings->'value')::jsonb->>'venue'='310' and c.case_type_id='CARE_SUPERVISION_EPO'

aliveni: I'll check with devops if this venue is used in prod. If it is used in prod, the new address which was added in [PR-1801](https://github.com/hmcts/fpl-ccd-configuration/pull/1801/files#diff-89b57754487e02efa27dac5f33a2c477048435ea5cda3df2f0e73ababb29e2d1R2453) should be used. I'll check with James

aliveni: James confirmed that hearing venue id `310` address should be replaced with venue id `312`, and `312` can be removed as it is no longer needed.
I verified the prod to check if venue id `312` is in use (DevOps ticket - https://tools.hmcts.net/jira/browse/RDO-9858)

nkoranne (approved):
LGTM

PR #1863, opened 2020-11-10
chore: workaround for flaky e2e tests

tomaszpowroznik:


e2e/helpers/hooks_helper.js
```diff
@@ -13,7 +13,7 @@ module.exports = class HooksHelpers extends Helper {
     recorder.retry({
       retries: 10,
       minTimeout: 1000,
-      when: err => err.message.indexOf('Execution context was destroyed') > -1,
+      when: err => err.message.indexOf('Execution context was destroyed') > -1 || err.message.indexOf('Node is') > -1,
```
lewisbirks, line 16: So is this going to retry when the message contains these strings?

lewisbirks (dismissed):
Looks fine based on the problem you have described

e2e/helpers/hooks_helper.js
tomaszpowroznik: yes, there were 2 messages
Node is either not visible or not an HTMLElement
and one I dont remember something like Node id detached
will find exact error and fix this one

tomaszpowroznik: done

lewisbirks (approved):

samuelzerozero (approved):

PR #1875, opened 2020-11-12
fix: representedBy not persisting in respondent event (FPLA-2428)

lewisbirks:

[FPLA-2428](https://tools.hmcts.net/jira/browse/FPLA-2428)

### Change description ###

Update representedBy collection with data from case data before during persistence mid event.

#### Steps to reproduce
To reproduce this issue you would have to be in an environment that is not the one for this PR.
1. Create a case with at least one respondent with confidential details and be in at least the submitted state
2. Log in as admin
3. Associate respondent with representative
3.1. This association can be checked by connecting to the database and running the query
    ```
    select elem -> 'value' -> 'party' ->> 'firstName' as name,  elem -> 'value' ->> 'representedBy' as repBy
    from case_data, jsonb_array_elements(data -> 'respondents1') elem
    where reference = CCD_NUM;
    ```
    where `CCD_NUM` is the case number assigned by CCD
4. Try to go through the respondents event again. The error `No field found` should be visible when you try to submit.

```
[ ] Yes
[x] No
```


joez-f (dismissed):
LGTM

karthikkannankainos (approved):

PR #1885, opened 2020-11-13
feat: Re-list cancelled hearing (FPLA-2200)

tomaszpowroznik:


service/src/test/java/uk/gov/hmcts/reform/fpl/model/HearingBookingTest.java
```diff
@@ -146,4 +149,105 @@ void shouldReturnFalseForHearingWithoutStartDate() {
...
     }
+
+    @Nested
+    class Status {
+
+        @ParameterizedTest
+        @EnumSource(
+            value = HearingStatus.class,
+            names = {"ADJOURNED", "ADJOURNED_TO_BE_RE_LISTED", "ADJOURNED_AND_RE_LISTED"},
+            mode = EnumSource.Mode.INCLUDE)
```
lewisbirks, line 160: Don't need to specify mode here, it is the default option

service/src/test/java/uk/gov/hmcts/reform/fpl/model/HearingBookingTest.java
```diff
@@ -146,4 +149,105 @@ void shouldReturnFalseForHearingWithoutStartDate() {
...
+
+            assertThat(hearingBooking.isAdjourned()).isTrue();
+            assertThat(hearingBooking.isVacated()).isFalse();
+        }
+
+        @ParameterizedTest
+        @EnumSource(
+            value = uk.gov.hmcts.reform.fpl.enums.HearingStatus.class,
+            names = {"VACATED", "VACATED_TO_BE_RE_LISTED", "VACATED_AND_RE_LISTED"},
+            mode = EnumSource.Mode.INCLUDE)
```
lewisbirks, line 172: Don't need to specify mode here, it is the default option

service/src/test/java/uk/gov/hmcts/reform/fpl/model/HearingBookingTest.java
```diff
@@ -146,4 +149,105 @@ void shouldReturnFalseForHearingWithoutStartDate() {
...
+
+            assertThat(hearingBooking.isVacated()).isTrue();
+            assertThat(hearingBooking.isAdjourned()).isFalse();
+        }
+
+        @ParameterizedTest
+        @EnumSource(
+            value = HearingStatus.class,
+            names = {"ADJOURNED_TO_BE_RE_LISTED", "VACATED_TO_BE_RE_LISTED"},
+            mode = EnumSource.Mode.INCLUDE)
```
lewisbirks, line 184: Don't need to specify mode here, it is the default option

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -612,6 +605,7 @@ public DynamicList buildDynamicHearingList(UUID selected) {
     private final Object hearingDateList;
     private final Object pastAndTodayHearingDateList;
     private final Object futureAndTodayHearingDateList;
+    private final Object toReListHearingDateList;
```
lewisbirks, line 608: Thoughts on creating a `ManageHearingsEventData` class like I did with the `UploadCMOEventData`?

lewisbirks: Maybe (if you think that it is a worthwhile idea) better as a future tech debt, would be quite far out of the scope of this story

e2e/tests/hmctsAdministersCaseAfterSubmission_test.js
```diff
@@ -349,6 +349,50 @@ Scenario('HMCTS admin vacates and re-lists a hearing', async ({I, caseViewPage,
   I.seeInTab(['Further evidence documents 2', 'Hearing'], 'Case management hearing, 1 January 2060');
 });
 
+Scenario('HMCTS admin re-lists previously cancelled hearing', async ({I, caseViewPage, manageHearingsEventPage}) => {
```
lewisbirks, line 352: vv minor but the hearing is not previously cancelled as you are cancelling it as part of the test

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageHearingsService.java
```diff
@@ -250,6 +281,21 @@ public void addOrUpdate(Element<HearingBooking> hearingBooking, CaseData caseDat
         }
     }
 
+    public Object getHearingsDynamicList(CaseData caseData) {
```
lewisbirks, line 284: Might it be sensible to combine this with the `getSelectedHearingId` method, it seems that it will only be used when you want the id anyway.
Something like
```java
    public UUID getSelectedHearingId(CaseData caseData) {
        Object dynamicList;
        switch (caseData.getHearingOption()) {
            case VACATE_HEARING:
                dynamicList = caseData.getFutureAndTodayHearingDateList();
                break;
            case ADJOURN_HEARING:
                dynamicList =  caseData.getPastAndTodayHearingDateList();
                break;
            case RE_LIST_HEARING:
                dynamicList =  caseData.getToReListHearingDateList();
                break;
            case EDIT_HEARING:
                dynamicList =  caseData.getHearingDateList();
                break;
            default:
                return null;
        }
        return ElementUtils.getDynamicListSelectedValue(dynamicList, mapper);
    }
```

lewisbirks (commented):
Looked over some of it, some support stuff has come up though, will try and get around to looking at the rest later

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ManageHearingsControllerAboutToSubmitTest.java
```diff
@@ -276,7 +312,7 @@ void shouldAdjournHearing(HearingReListOption adjournmentOption) {
         Element<HearingBooking> expectedAdjournedHearing = element(
             pastHearingToBeAdjourned.getId(),
             pastHearingToBeAdjourned.getValue().toBuilder()
-                .status(HearingStatus.ADJOURNED)
+                .status(adjournmentOption == RE_LIST_LATER ? ADJOURNED_TO_BE_RE_LISTED : ADJOURNED)
```
samuelzerozero, line 315: I'd avoid to add if statements in tests. You can pass the expected value as a test parameter

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ManageHearingsControllerAboutToSubmitTest.java
```diff
@@ -308,7 +344,7 @@ void shouldVacateHearing(HearingReListOption adjournmentOption) {
         Element<HearingBooking> expectedVacatedHearing = element(
             futureHearingToBeAdjourned.getId(),
             futureHearingToBeAdjourned.getValue().toBuilder()
-                .status(HearingStatus.VACATED)
+                .status(adjournmentOption == RE_LIST_LATER ? VACATED_TO_BE_RE_LISTED : VACATED)
```
samuelzerozero, line 347: I'd avoid to add if statements in tests. You can pass the expected value as a test parameter

samuelzerozero (approved):

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ManageHearingsControllerAboutToSubmitTest.java
tomaszpowroznik: agree, did it this way as test is already parametrised and I dont really like @MethodSource to provide multiple params, happy to change

samuelzerozero (approved):

PR #1888, opened 2020-11-16
ci: bump chart versions

lewisbirks:
### Change description ###

Bumping chart versions to see if that resolves the nil pointer issue

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/fpl-case-service/Chart.yaml
+++ b/charts/fpl-case-service/Chart.yaml
@@ -1,7 +1,7 @@
 name: fpl-case-service
 apiVersion: v2
 home: https://github.com/hmcts/fpl-ccd-configuration
-version: 1.12.7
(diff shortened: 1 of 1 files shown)


charts/fpl-case-service/Chart.yaml
```diff
@@ -1,17 +1,17 @@
...
 maintainers:
   - name: HMCTS Family Public Law team
     email: fpl-developers@HMCTS.NET
 dependencies:
   - name: java
-    version: 3.0.0
+    version: 3.4.4
     repository: '@hmctspublic'
   - name: ccd
-    version: 4.3.4
```
lewisbirks, line 14: Do we want to revert to the patch selector?

tomaszpowroznik (approved):

PR #1917, opened 2020-11-23
support: Prepare definitions for data migration

tomaszpowroznik:


ccd-definition/CaseEvent/CareSupervision/MultiState.json
```diff
@@ -7,7 +7,6 @@
     "PreConditionState(s)": "*",
     "PostConditionState": "*",
     "SecurityClassification": "Public",
-    "CallBackURLAboutToSubmitEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/migrate-case/about-to-submit",
```
lewisbirks, line 10: Why is this being removed?

lewisbirks: Ahh, I think I see. You are doing the migration in the other PR and as such don't want to do the callback here.

tomaszpowroznik: yes, dont want to call any callback. As it will go through all cases it may have performance impact

lewisbirks (approved):

PR #1935, opened 2020-11-26
fix: Back to first page when refresh happened on second page of manage documents event (FPLA-2451)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2451


joez-f (approved):
👍

aliveni (approved):

PR #1942, opened 2020-11-30
chore: update xui version

tomaszpowroznik:
Due to performance reasons, some tests has been split into multiple files. When e.g PARALLEL_CHUNKS = 10 is defined, it does not mean that there will be 10 workers competing for test to be executed. Instead, list of our test files is sorted alphabetically and split into 10 buckets, so that each bucket contains same (more or less) number of files. Then number of workers equals number of buckets is spin up. Once worker finish with its bucket, it does not help with other buckets. this lead to situation where some buckets have 5 tests in total and some have 25. Goal of splitting files it to have more even distribution among workers (avoid files with big number of tests within)


charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -61,7 +61,15 @@ xui-webapp:
       SERVICES_CCD_DATA_STORE_API: http://${SERVICE_NAME}-ccd-data-store-api
       SERVICES_TERMS_AND_CONDITIONS: http://xui-terms-and-conditions-${SERVICE_FQDN}.service.core-compute-preview.internal
       JURISDICTIONS: PUBLICLAW
-
+      LAUNCH_DARKLY_CLIENT_ID: 5e21bc1a8fd1c908894bf7e7
```
tomaszpowroznik, line 64: id of fpl LD account, so we can control ES toggle, alternatively can be non existing id

charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -61,7 +61,15 @@ xui-webapp:
...
+      FEATURE_REDIS_ENABLED: "false"
+      REDISCLOUD_URL: http://dummyrediscloudurl
+    keyVaults:
```
tomaszpowroznik, line 67: overrides default XUI chart mapping, as original pull LD id from vault and overrides what is in env vars

e2e/pages/caseView.page.js
```diff
@@ -30,9 +30,8 @@ module.exports = {
   caseTitle: '.case-title .markdown',
 
   async goToNewActions(actionSelected) {
-    I.waitForElement(this.actionsDropdown);
```
tomaszpowroznik, line 33: this throws exception if for some reason component is not found within default timeout (20s)

samuelzerozero (dismissed):

e2e/tests/courtAdminMangesHearings_test.js
```diff
@@ -0,0 +1,307 @@
...
+BeforeSuite(async ({I}) => {
+  caseId = await I.submitNewCaseWithData(mandatoryWithMultipleChildren);
+  submittedAt = new Date();
+
+  await I.navigateToCaseDetailsAs(config.hmctsAdminUser, caseId);
+});
+
+Before(async ({I}) => await I.navigateToCaseDetailsAs(config.hmctsAdminUser, caseId));
+
+Scenario('HMCTS admin enters FamilyMan case number', async ({I, caseViewPage, enterFamilyManCaseNumberEventPage}) => {
```
lewisbirks, line 25: Do we want this to be part of the before all?

tomaszpowroznik: done

samuelzerozero (approved):

PR #1967, opened 2020-12-07
chore: Remove old respondents filed

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2501
https://github.com/hmcts/fpl-ccd-data-migration-tool/pull/34


lewisbirks:
Have you checked for `children`?

tomaszpowroznik:
> Have you checked for `children`? Might need to remove that one as well

thanks Lewis, was not aware about this field

lewisbirks (commented):
Technically the test data for the old fields is wrong but as we are only asserting that the entries are removed it doesn't matter.

lewisbirks (approved):

PR #1971, opened 2020-12-08
fix: replace document download link for case access users (FPLA-2503)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-2503

### Change description ###

For users that have case access, document links in notifications should be a direct link to that document in the case
For users without access, they should still see the gov notify download link

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseManagementOrderIssuedEventHandler.java
```diff
@@ -36,20 +37,21 @@
...
         CaseManagementOrder issuedCmo = event.getCmo();
 
-        //TODO Document is downloaded 5 times, this could impact performance. Investigate in FPLA-2061
         sendToLocalAuthority(caseData, issuedCmo);
         sendToCafcass(caseData, issuedCmo);
         sendToRepresentatives(caseData, issuedCmo, DIGITAL_SERVICE);
         sendToRepresentatives(caseData, issuedCmo, EMAIL);
         issuedOrderAdminNotificationHandler.notifyAdmin(caseData,
-            documentDownloadService.downloadDocument(issuedCmo.getOrder().getBinaryUrl()), CMO);
+            concatUrlAndMostRecentUploadedDocumentPath(xuiBaseUrl, issuedCmo.getOrder().getBinaryUrl()), CMO);
```
joez-f, line 54: technically this changes how things work currently look for admin (but they should be seeing the link to document in case, not the download - needs business to verify)

joez-f: business said that's correct, so I will also make the change for C110a for admin

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NoticeOfPlacementOrderUploadedEventHandler.java
```diff
@@ -46,15 +49,15 @@ public void notifyParties(NoticeOfPlacementOrderUploadedEvent noticeOfPlacementE
         notificationService.sendEmail(
             NOTICE_OF_PLACEMENT_ORDER_UPLOADED_TEMPLATE, emails, notifyData, caseData.getId().toString());
 
-        issuedOrderAdminNotificationHandler.notifyAdmin(caseData,
-            noticeOfPlacementEvent.getDocumentContents(), NOTICE_OF_PLACEMENT_ORDER);
+        issuedOrderAdminNotificationHandler.notifyAdmin(
+            caseData, orderDocument.getBinaryUrl(), NOTICE_OF_PLACEMENT_ORDER);
```
tomaszpowroznik, line 53: minor, for consistence with e.g changes below it would be better to pass  orderDocument. It is type safe and better prepare for future changes. If e.g we would like do display name of doc in emai, then when passing entire DocReference, no changes to interface are needed

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/GeneratedOrderEventHandler.java
```diff
@@ -38,17 +38,16 @@
...
-    private final FeatureToggleService featureToggleService;
 
     @EventListener
     public void notifyParties(final GeneratedOrderEvent orderEvent) {
         final CaseData caseData = orderEvent.getCaseData();
-        final byte[] documentContents = orderEvent.getDocumentContents();
+        final DocumentReference orderDocument = orderEvent.getOrderDocument();
 
-        issuedOrderAdminNotificationHandler.notifyAdmin(caseData, orderEvent.getDocumentContents(), GENERATED_ORDER);
+        issuedOrderAdminNotificationHandler.notifyAdmin(caseData, orderDocument.getBinaryUrl(), GENERATED_ORDER);
```
tomaszpowroznik, line 47: inconsistency, it would be better to pass entire orderDocument, same as in sendNotificationToEmailServedRepresentatives(caseData, orderDocument);

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/OrderIssuedNotificationTestHelper.java
```diff
@@ -41,22 +41,17 @@ private OrderIssuedNotificationTestHelper() {
             "orderType", orderType.toLowerCase(),
             "callout", withCallout ? callout : "",
             "courtName", EXAMPLE_COURT,
-            "documentLink", jsonFileObject.toMap(),
+            "documentLink", jsonFileObject,
```
tomaszpowroznik, line 44: hmm, something smells here. In OrderIssuedNotifyData there is comment
```
private final Object documentLink; //This will be String or Map<String, Object> depending on case access/no access
```
and now expected object contains JsonObject and not map

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/NoticeOfPlacementOrderUploadedEventHandler.java
joez-f: I agree, thanks, will pass whole document

tomaszpowroznik (approved):

lewisbirks (approved):

PR #1975, opened 2020-12-08
migration: Remove invalid hearings (FPLA-2491)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2491
https://tools.hmcts.net/jira/browse/RDO-10240
https://github.com/hmcts/fpl-ccd-data-migration-tool/pull/35


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -48,28 +49,45 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
...
 
-            if (!expectedHearingDate.equals(hearingDate)) {
-                throw new IllegalArgumentException(format("Invalid hearing date %s", hearingDate));
+            if (!HearingType.ISSUE_RESOLUTION.equals(hearing2.getValue().getType())) {
+                throw new IllegalArgumentException(
+                    format("Invalid hearing type %s", hearing2.getValue().getType()));
             }
 
-            hearings.remove(hearingIndex);
+            hearings.remove(hearing2);
```
lewisbirks, line 89: If they are duplicate will this remove the expected one? By being duplicate it will remove based on `equals` method but this will be the same for 2 hearings. You are getting the 4th and 5th hearing to assert for removal but from the test it seems that it determines that 2 earlier ones are removed. Maybe it would be safer to remove on index here.

lewisbirks: ahhh no ignore me, the element means that it will be unique

lewisbirks (approved):

PR #1980, opened 2020-12-09
feat: FPLA-2461 save CMO and SDO orders in word format

aliveni:
https://tools.hmcts.net/jira/browse/FPLA-2461

### Change description ###
Store last uploaded SDO and CMO documents in word format before the documents are converted to pdf files.

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/StandardDirectionsOrderControllerAboutToSubmitTest.java
```diff
@@ -395,6 +397,11 @@ private StandardDirectionOrder expectedOrder() {
         return StandardDirectionOrder.builder()
             .directions(fullyPopulatedDirections())
             .orderStatus(SEALED)
+            .orderDoc(DocumentReference.builder()
```
lewisbirks, line 400:
```suggestion
            .lastUploadedOrder(DocumentReference.builder()
```

ccd-definition/ComplexTypes/CareSupervision/CaseManagementOrder/CaseManagementOrder.json
```diff
@@ -7,6 +7,15 @@
     "ElementLabel": "Order",
     "SecurityClassification": "Public"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "ID": "UploadedCaseManagementOrder",
+    "ListElementCode": "lastUploadedOrder",
```
lewisbirks, line 13: Would wordVersionOfOrder or something like that be better, I know that it will be the last uploaded order but that name doesn't seem to fit right in my head

aliveni: As discussed on the Slack channel, I would like to leave the name as `lastUploadedOrder` which is clear that the order was uploaded by the user instead of a generated document

lewisbirks (approved):

PR #1989, opened 2020-12-11
fix: Fix document upload

tomaszpowroznik:
Steps to reproduce
1. As LA Upload cmo with supporting documents
2. As hmcts user upload further evidence document for same hearing as in step 1

if document added in step 1 is the only one supporting document added for hearing, then step 2 will skip page when document can be added, and proceeds to summary page


joez-f (approved):
look good

PR #1996, opened 2020-12-14
ci: use ElasticSearch in preview environment

tomaszpowroznik:
deploy ES stack per PR, and configure ccd-data-store, ccd-definition-store and xui to use it

### DIFF
--- a/charts/fpl-case-service/values.preview.template.yaml
+++ b/charts/fpl-case-service/values.preview.template.yaml
@@ -61,7 +61,7 @@ xui-webapp:
       SERVICES_CCD_DATA_STORE_API: http://${SERVICE_NAME}-ccd-data-store-api
       SERVICES_TERMS_AND_CONDITIONS: http://xui-terms-and-conditions-${SERVICE_FQDN}.service.core-compute-preview.internal
       JURISDICTIONS: PUBLICLAW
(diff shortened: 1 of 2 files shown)


samuelzerozero (approved):

charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -61,7 +61,7 @@ xui-webapp:
       SERVICES_CCD_DATA_STORE_API: http://${SERVICE_NAME}-ccd-data-store-api
       SERVICES_TERMS_AND_CONDITIONS: http://xui-terms-and-conditions-${SERVICE_FQDN}.service.core-compute-preview.internal
       JURISDICTIONS: PUBLICLAW
-      LAUNCH_DARKLY_CLIENT_ID: dummy
+      LAUNCH_DARKLY_CLIENT_ID: 5de6610b23ce5408280f2268
```
lewisbirks, line 64: Do we want this to be public?

tomaszpowroznik: ideally not, this is xui id, exposed by xui publicly 
https://github.com/hmcts/rpx-xui-

webapp/blob/27b2eb59e045a1f72796b079712fe099089c0c19/charts/xui-webapp/values.preview.template.yaml#L9

PR #1997, opened 2020-12-14
FPLA-2379: Data migration for LA docs to new collection

samuelzerozero:
https://tools.hmcts.net/jira/browse/FPLA-2379

### Change description ###
Migration


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadDocumentsController.java
```diff
@@ -66,6 +66,7 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
             data.put("applicationDocumentsToFollowReason",
                 migratedData.getOrDefault("applicationDocumentsToFollowReason", null)
             );
+            data.put("courtBundleList", migratedData.getOrDefault("courtBundleList", null));
```
joez-f, line 69: can we do `caseDetails.getData().putAll(uploadDocumentsMigrationService.transformFromOldCaseData(caseData));` instead of data.put.. data.put... data.put..., and handle the null logic in the service i.e `        data.put("courtBundleList", !courtBundleList.isEmpty() ? courtBundleList : null);
`
I missed this because of how we are handling these maps

samuelzerozero: 👍

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicationDocumentsService.java
```diff
@@ -50,7 +50,8 @@
                 Optional<Element<ApplicationDocument>> documentBefore = findElement(document.getId(),
                     previousDocuments);
 
-                if (documentBefore.isPresent()) {
+                // in the old flow, we allowed other documents with just title and no file
+                if (documentBefore.map(doc -> doc.getValue().hasDocument()).orElse(false)) {
```
samuelzerozero, line 54: The old UI, allow Other documents to be added without file attached. In the new collection the file is mandatory. In that case, upon migration, the call would thrown an exception on handleExisitingDocuments()

joez-f (approved):
Looks good to me

PR #1999, opened 2020-12-14
fix: Prevent documents duplications when uploading cmo (FPLA-2512)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2512

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/cmo/UploadCMOService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/cmo/UploadCMOService.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.reform.fpl.service.cmo;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.Iterables;
 import lombok.RequiredArgsConstructor;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
@@ -25,6 +26,7 @@ import uk.gov.hmcts.reform.fpl.service.time.Time;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/cmo/UploadCMOService.java
```diff
@@ -163,15 +163,23 @@ public UploadCMOEvent buildEventToPublish(CaseData caseData, CaseData caseDataBe
...
-                                  HearingBooking hearing, List<Element<SupportingEvidenceBundle>> supportingDocs) {
+                                  HearingBooking hearing, List<Element<SupportingEvidenceBundle>> cmpSupportingDocs) {
         Optional<Element<HearingFurtherEvidenceBundle>> bundle = findElement(selectedHearingId, evidenceBundles);
 
         if (bundle.isPresent()) {
-            bundle.get().getValue().getSupportingEvidenceBundle().addAll(supportingDocs);
+            List<Element<SupportingEvidenceBundle>> hearingDocs = bundle.get().getValue().getSupportingEvidenceBundle();
+
+            cmpSupportingDocs.forEach(element -> {
+                hearingDocs.stream()
```
lewisbirks, line 173: would it be better to only add those that aren't present? Or are you trying to keep them all grouped in the collection?

something like
```java
cmpSupportingDocs.stream()
                .filter(doc -> hearingDocs.stream().anyMatch(hearingDoc -> hearingDoc.getId().equals(doc.getId())))
                .forEach(hearingDocs::add);
```

tomaszpowroznik: If you add only new, then changes to existing supporting document wont be propagated to hearing bundle.

samuelzerozero (approved):

service/src/main/java/uk/gov/hmcts/reform/fpl/service/cmo/UploadCMOService.java
lewisbirks: Good point, if the replace one or update a field it probably won't change the element id

PR #2012, opened 2020-12-16
fix: show appropriate error when no C2s to support

joez-f:
### Change description ###

Show error message to user in manage documents flow when no C2s to attach supporting docs to

Same as Manage Docs for LA journey

```
[ ] Yes
[X] No
```


aliveni (approved):
LGTM

PR #2013, opened 2020-12-16
EXUI regression test: verify hidden fields maintain value in hearing event

joez-f:
### Change description ###

firstHearingFlag only populated in about-to-start
If EXUI have fixed regression, this value should be maintained throughout event
Easiest way to check this is to add hearing in future, then go into event again, vacate hearing + relist it

```
[ ] Yes
[X] No
```


joez-f:
looks like hidden field firstHearingFlag was maintained throughout event 👍

PR #2014, opened 2020-12-16
fix: show appropriate error when no hearings to relist

joez-f:
### Change description ###

Show user appropriate error message when they try to re-list a hearing but there are no hearings to re-list. Currently it just throws an exception and tells them something went wrong (we should catch these exceptions)

```
[ ] Yes
[X] No
```

image


aliveni (dismissed):
LGTM

tomaszpowroznik (dismissed):

tomaszpowroznik (approved):

PR #2023, opened 2020-12-21
fix: Change case state when sending case to gatekeeper in a transactional way (FPLA-2544)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2544


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperController.java
```diff
@@ -51,18 +52,25 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(@RequestBody
...
-            publishEvent(new PopulateStandardDirectionsEvent(callbackRequest));
-        }
         publishEvent(new NotifyGatekeepersEvent(caseData));
     }
 
     private List<Element<EmailAddress>> resetGateKeeperEmailCollection() {
-        return List.of(
-            element(EmailAddress.builder().email("").build())
-        );
+        return wrapElements(EmailAddress.builder().email("").build());
```
lewisbirks, line 74: Just noticed that this doesn't add an id, might be something that needs changing. Not sure how CCD would like it.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperController.java
```diff
@@ -51,18 +52,25 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(@RequestBody
         return respond(caseDetails, errors);
     }
 
+    @PostMapping("/about-to-submit")
+    public AboutToStartOrSubmitCallbackResponse handleAboutToSubmitEvent(@RequestBody CallbackRequest callbackRequest) {
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseData caseData = getCaseData(caseDetails);
+
+        if (SUBMITTED.equals(caseData.getState())) {
```
joez-f, line 60: I'd say this check + test for it isn't needed because event is only available in submitted state, but if we want to make extra sure it's probably fine

joez-f (approved):
looks good, maybe we can delete PopulateStandardDirectionsEvent too if we set the state manually in RemoveOrderController, since this is the only other place it is used

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperController.java
lewisbirks: Isn't a copy of the event available in the gatekeeping state as well

joez-f: yes, tomek mentioned it to me, I forgot about that part - makes sense to keep the test

PR #2026, opened 2020-12-22
impr: idam api retries (FPLA-2389)

lewisbirks:

[FPLA-2389](https://tools.hmcts.net/jira/browse/FPLA-2389)

### Change description ###

Added a retrier and error decoder to the idam API and added unit tests for both of these components.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/config/feign/errordecoder/IdamErrorDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+                response.status(),
+                response.reason(),
+                response.request().httpMethod(),
+                exception,
+                null,
+                response.request()
+            );
+        }
+
+        return null;
```
tomaszpowroznik, line 39: are you sure about this one? Default decode returns exception from line 22

lewisbirks: yeah, my bad would want to return the original exception as otherwise we would be getting NPEs in the code

service/src/main/java/uk/gov/hmcts/reform/fpl/config/feign/errordecoder/IdamErrorDecoder.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public Exception decode(String methodKey, Response response) {
+        Exception exception = defaultDecoder.decode(methodKey, response);
+
+        if (exception instanceof RetryableException) {
+            return exception;
+        }
+
+        if (HttpStatus.valueOf(response.status()).is5xxServerError()) {
+            return new RetryableException(
```
tomaszpowroznik, line 29: throwing RetrywableException can make change not transparent to existing code. If there is code that do try{}catch(SomeException ex), ex will be wrapped in RetryableException and catch wont be trigger.

there is option to configure propagation policy
    @Bean
    ExceptionPropagationPolicy propagationPolicy() {
        return ExceptionPropagationPolicy.UNWRAP;
    }
    
  Your change is about http 5xx family only so i guess it should be fine (dont expect to have catch(Feign.InternalServerError)), but worth checking.
  
In general it is preferred to make retries transparent so i would consider using ExceptionPropagationPolicy.UNWRAP

lewisbirks: Yeah, makes sense. Added

service/src/test/java/uk/gov/hmcts/reform/fpl/config/feign/IdamClientTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+class IdamClientTest {
+    private static final Map<String, Collection<String>> EMPTY_HEADERS = Map.of();
+    private static final String EMPTY_REASON = "";
+    private static final byte[] EMPTY_BODY = new byte[0];
+    private static final Request REQUEST = Request.create(GET, EMPTY, Map.of(), new byte[] {}, UTF_8, null);
+
+    private final Client httpClient = mock(Client.Default.class);
+    private final IdamApi idamApi = Feign.builder()
+        .errorDecoder(new IdamErrorDecoder())
```
tomaszpowroznik, line 38: if you add 
.decoder(new JacksonDecoder())

and change EMPTY_BODY = "{\"id\":\"2\"}".getBytes() 

then in your test you can prove that user is returned after retries

UserDetails u = idamClient.getUserDetails("details")

tomaszpowroznik (dismissed):

tomaszpowroznik (approved):

tomaszpowroznik (approved):

PR #2027, opened 2020-12-22
fix: Validate if additional documents have file attached (FPLA-2551)

tomaszpowroznik:


service/src/main/java/uk/gov/hmcts/reform/fpl/service/DocumentsValidatorService.java
```diff
@@ -41,6 +42,9 @@
                 if (isEmpty(document.getDocumentTitle())) {
                     validationErrors.add(String.format("You must give additional document %s a name.", i.get()));
                 }
+                if (isNull(document.getTypeOfDocument())) {
+                    validationErrors.add(String.format("You must upload a file for additional document %s.", i.get()));
```
lewisbirks, line 46: minor but an atomic int is overkill here. there is a class called IncrementalInteger which provides the same functionality without the thread safety overhead

lewisbirks (approved):

PR #2045, opened 2021-01-06
NOJIRA: fix smoke test adding retry

samuelzerozero:
*fix smoke test retry

### DIFF
--- a/e2e/tests/smoke_test.js
+++ b/e2e/tests/smoke_test.js
@@ -8,6 +8,6 @@ Scenario('Sign in as local authority and create a case', async ({I, caseListPage
   const caseName = `smoke test case (${new Date().toISOString()})`;
   const caseId = await I.logInAndCreateCase(config.swanseaLocalAuthorityUserOne, caseName);
   await I.navigateToCaseList();
-  await caseListPage.searchForCasesWithName(caseName, 'Open');
+  await I.retryUntilExists(() => caseListPage.searchForCasesWithName(caseName, 'Open'), `//ccd-search-result/table//tr[//a[contains(@href,'${caseId}')]]`);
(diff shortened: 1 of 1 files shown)


e2e/tests/smoke_test.js
```diff
@@ -8,6 +8,6 @@ Scenario('Sign in as local authority and create a case', async ({I, caseListPage
   const caseName = `smoke test case (${new Date().toISOString()})`;
   const caseId = await I.logInAndCreateCase(config.swanseaLocalAuthorityUserOne, caseName);
   await I.navigateToCaseList();
-  await caseListPage.searchForCasesWithName(caseName, 'Open');
+  await I.retryUntilExists(() => caseListPage.searchForCasesWithName(caseName, 'Open'), `//ccd-search-result/table//tr[//a[contains(@href,'${caseId}')]]`);
```
tomaszpowroznik, line 11: ideally retries should be inside caseListPage.searchForCasesWithName and selector should be in page object

tomaszpowroznik (approved):

e2e/tests/smoke_test.js
lewisbirks, line 11: I was thinking of instead changing the timeouts for the search stuff to be available

lewisbirks: my current changes have increased timeout on several waitForElements and it worked 4/5 times with the only failure being the `net::ERR_ABORTED` error

tomaszpowroznik: the problem is different, timeout wont help. It is not that it loads slowly.  The issue is that case is created, and not yet propagated from db->ES, and we try to search for it in ES. Thats why test needs to click search multiple time and repeat request until case is in ES

PR #2063, opened 2021-01-11
fix: use TabLabel instead of TabID in case urls (FPLA-2565)

joez-f:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/FPLA-2565

### Change description ###

fixed what EXUI broke, case url links to tabs now use TabLabel instead of TabID

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[X] No
```

PR #2086, opened 2021-01-14
impr: update c2 notifications (FPLA-2583)

SooryaSu:
[FPLA-2583](https://tools.hmcts.net/jira/browse/FPLA-2583)

### Change description ###
 * Judges no longer receive C2 notifications.
 * Updated the template used when sending C2 notifications to admins.

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/handlers/C2UploadedEventHandlerEmailTemplateTest.java
```diff
@@ -0,0 +1,96 @@
...
+            );
+
+        underTest.notifyAdmin(new C2UploadedEvent(caseData, c2DocumentBundle));
+
+        assertThat(response())
+            .hasSubject("C2 application received, " + respondentLastName)
+            .hasBody(emailContent()
+                .line("A C2 application has been received for the case:")
+                .line()
+                .line("^ " + callout)//callout
```
lewisbirks, line 80: You could have a specific method called `callout()` that could be chain called here and would prefix the parameter with `"^ "`, other tests will need it and it means that you just have to focus on what the text will be instead of the format, same for the header and list below. You could have a h1 method which prefixes whatever is passed with `#` and a list method that takes in a var args of string and prefixes each one with `* `.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/handlers/C2UploadedEventHandlerEmailTemplateTest.java
```diff
@@ -0,0 +1,93 @@
...
+    C2UploadedEventHandler.class,
+    NotificationService.class,
+    ObjectMapper.class
+})
+
+public class C2UploadedEventHandlerEmailTemplateTest extends EmailTemplateTest {
+    private final C2DocumentBundle c2DocumentBundle = C2DocumentBundle.builder().build();
+    private final String respondentLastName = "Smith";
+    private final String calloutText = "Smith, SACCCCCCCC5676576567";
+    private final String caseUrl = "null/case/" + JURISDICTION + "/" + CASE_TYPE + "/12345#C2Tab";
```
joez-f, line 39: Depending on whether my PR https://github.com/hmcts/fpl-ccd-configuration/pull/2063/files is merged first, this will change slightly, but I'll keep an eye on it and let you know if anything needs to change

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/handlers/C2UploadedEventHandlerEmailTemplateTest.java
```diff
@@ -0,0 +1,93 @@
...
+            UserInfo.builder().sub("hmcts-non-admin@test.com").roles(LOCAL_AUTHORITY.getRoles()).build());
+
+        given(c2UploadedEmailContentProvider.getNotifyData(caseData,
+            c2DocumentBundle.getDocument()))
+            .willReturn(
+                C2UploadedTemplate.builder()
+                    .callout(calloutText)
+                    .respondentLastName(respondentLastName)
+                    .caseUrl(caseUrl)
+                    .documentLink(DOCUMENT_LINK)
```
joez-f, line 69: so this document link is not actually a Map<String, Object> in this situation, it's a string - essentially it's another URL. We have two ways of 'sending a document': one is to attach it by doing what you're doing here with the JSONObject, the other way is to send a case link to that document (so another url, but one that links directly to the document in the case - and will require user to be logged in/signed in). We want to use the latter as much as possible because there were some GDPR concerns about using the attached document process (anyone with this link can download the document, it was frowned upon) - so wherever we can (i.e, if user has case access), we will give them a URL linking to the document - if they dont have case access, then we can use the JSONObject approach. You havent changed the main code but just need to update this part of the test to reflect what happens in the live template

joez-f: you'll see what I mean if you look into C2UploadedEventHandler -> C2UploadedEmailContentProvider, the getDocumentUrl method is just building a String (and you'll see above it the method for the other approach, which is to 'attach' the document)

joez-f (approved):
looks good now 👍

joez-f (approved):

PR #2087, opened 2021-01-14
impr: log LA domain exception as warning (FPLA-2579)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-2579

### Change description ###

add new abstract exception + handler so can log as warning instead of error (no alert from azure)

```
[ ] Yes
[X] No
```


lewisbirks (approved):

lewisbirks (changes requested):
There isn't a test that covers this scenario would probs be wise for there to be one

joez-f:
> There isn't a test that covers this scenario would probs be wise for there to be one

added a test now

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseInitiationController.java
```diff
@@ -34,7 +39,14 @@ public AboutToStartOrSubmitCallbackResponse handleMidEvent(
         @RequestBody CallbackRequest callbackrequest) {
         CaseDetails caseDetails = callbackrequest.getCaseDetails();
 
-        return respond(caseDetails, localAuthorityOnboardedValidationService.validateIfUserIsOnboarded());
+        List<String> errors = new ArrayList<>();
+        try {
+            errors = localAuthorityOnboardedValidationService.validateIfUserIsOnboarded();
```
tomaszpowroznik, line 44: it could be done in more generic way and more transparently, see ResourceExceptionHandler. You can define new abstract exception type, make UnknownLocalAuthorityDomainException extends it and then add handler in ResourceExceptionHandler for new exception type that does not log error but warning

joez-f: will have a look at that, thanks

lewisbirks:
@tomaszpowroznik @joez-f would we still want an error into the support channel if the email ended in gov uk, just not for private solicitors?

tomaszpowroznik (approved):

PR #2107, opened 2021-01-18
impr: minor updates for messages and tabs (FPLA-2602)

SooryaSu:
[FPLA-2602](https://tools.hmcts.net/jira/browse/FPLA-2602)

### Change description ###
 * Changes the name for the "Message judge or legal adviser" event.
 * Changes the order of the tabs

```
[ ] Yes
[x] No
```


SooryaSu:
I noticed a few of the tabs were sharing the TabDisplayOrder numbers so I've updated it so each is unique. I've made the changes with the assumption the Overview tab will moved to be the first tab as part of [FPLA-2170](https://tools.hmcts.net/jira/browse/FPLA-2170)

ccd-definition/CaseEvent/CareSupervision/MultiState.json
```diff
@@ -675,9 +675,9 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "ID": "messageJudgeOrLegalAdviser",
```
SooryaSu, line 677: Should I also change the ID for this?

lewisbirks: If you want to note that you will have to update any authorisation entry, any case event to field entry, and any case event to complex type entry that uses it. It might also be used in any labels that are links to start the event.

SooryaSu: The task only seems to be asking to rename it in the dropdown and no functionality has changed so I think it is probably fine to leave it with the current ID as it still describes the event

lewisbirks: seems sensible, would be a lot more additional work otherwise

ccd-definition/CaseEvent/CareSupervision/MultiState.json
```diff
@@ -675,9 +675,9 @@
...
-    "Name": "Message judge or legal adviser",
```
lewisbirks, line 678: You will need to update the e2e tests, they rely on the name in the dropdown to start the event.

lewisbirks (changes requested):

lewisbirks (approved):

PR #2118, opened 2021-01-20
support: fix gatekeeper state change 

Jamesnelson117:

### Change description ###

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/NotifyGatekeeperController.java
```diff
@@ -51,18 +52,29 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToStartEvent(@RequestBody
...
 
+    @PostMapping("/about-to-submit")
+    public AboutToStartOrSubmitCallbackResponse handleAboutToSubmitEvent(@RequestBody CallbackRequest callbackRequest) {
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseDetails caseDetailsBefore = callbackRequest.getCaseDetailsBefore();
+        CaseData caseData = getCaseData(caseDetails);
+        CaseData caseDataBefore = getCaseData(caseDetailsBefore);
+
+
+        if (SUBMITTED.equals(caseDataBefore.getState())) {
```
joez-f, line 63: what was wrong with the previous implementation? I know there was a bug when we used caseData.getState(), but why? about-to-submit hasn't completed yet, so shouldn't getState() return SUBMITTED? is it some async issue? @tomaszpowroznik

tomaszpowroznik, line 63: maybe safer is not to have this check at all. this callback is called only from sendToGatekeeper and not from notifyGatekeeper. and sendToGatekeeper in ccd def is configured to be in single state - submitted. 

On the other side having check is safer as if somebody change ccd def in future to be accessible from different state then restriction will be in place.

Jamesnelson117: Moved to a new controller 'SendToGatekeeperController' since it's not shared between both events.

tomaszpowroznik: @joez-f it was about ccd definition. It was changed for sendToGatekeeper to be 
PreConditionState: Submitted
PostConditionState: Gatekeeping.

in a callback you have case data as it will be when you save it, that why in controller state is Gatekeeping and not Submitted

tomaszpowroznik (approved):

PR #2139, opened 2021-01-22
impr: update CTSC email about CMO draft (FPLA-2518)

SooryaSu:
[FPLA-2518](https://tools.hmcts.net/jira/browse/FPLA-2518)

### Change description ###
The notification sent to the Admin when a CMO is uploaded will now include the hearing judge if there is one assigned to the hearing.

```
[ ] Yes
[x] No
```


SooryaSu:
Jenkins seems to be failing on the tests in AgreedCMOUploadedEventHandlerEmailTemplateTest but these tests are passing in my local dev

joez-f (commented):
I think we will need to have a 2nd template for the CTSC email or add a flag, because the email starts with "Dear ((judgeTitle)) ((judgeName))" but this doesn't make sense for CTSC. We could replace this line with a single variable that we populate on our end, and then hide it with a flag based on who we're sending the notification to. Or we can duplicate the template and just remove that line/replace with something else for CTSC

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/cmo/AgreedCMOUploadedEventHandler.java
```diff
@@ -29,11 +31,24 @@
...
+
+        if (isNotEmpty(hearing.getJudgeAndLegalAdvisor().getJudgeEmailAddress())) {
+            judge = hearing.getJudgeAndLegalAdvisor();
+        } else if (caseData.hasAllocatedJudgeEmail()) {
+            judge = caseData.getAllocatedJudge();
+        }
+
+        if (judge == null) {
+            return;
+        }
```
joez-f, line 46: we could use the getJudgeToNotify method below for consistency - or replace the getJudgeToNotify method with what you have here if you prefer it + update the sendNotificationToJudge, that works too

joez-f (approved):
looks good

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/cmo/AgreedCMOUploadedEventHandler.java
```diff
@@ -29,21 +29,26 @@
...
+        JudgeAndLegalAdvisor judgeAttendingHearing = event.getHearing().getJudgeAndLegalAdvisor();
 
-        CMOReadyToSealTemplate template = contentProvider.buildTemplate(
-            event.getHearing(),
-            caseData.getId(),
-            caseData.getAllocatedJudge(),
-            caseData.getAllRespondents(),
-            caseData.getFamilyManCaseNumber()
-        );
+        if (judgeAttendingHearing.getJudgeEmailAddress() != null || caseData.hasAllocatedJudgeEmail()) {
```
joez-f, line 34: If you have some time, some of this is untested (check sonar), so can add 1 more test to cover it

PR #2148, opened 2021-01-26
chore/E2E stability improvements

tomaszpowroznik:
https://build.platform.hmcts.net/job/HMCTS_Nightly_FPL/job/fpl-ccd-configuration/job/nightly-dev/206/

### DIFF
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -67,8 +67,8 @@ withNightlyPipeline(type, product, component) {
 
   loadVaultSecrets(integrationTestSecrets)
 
-  enableSecurityScan()
-  enableMutationTest()
+//  enableSecurityScan()
+//  enableMutationTest()
   enableFullFunctionalTest(120)
 
   before('mutationTest') {
@@ -85,7 +85,6 @@ withNightlyPipeline(type, product, component) {
   before('fullFunctionalTest') {
     env.PARALLEL_CHUNKS = params.FUNCTIONAL_TESTS_WORKERS
     env.MOCKED_PAYMENTS = false
-    env.TEST_RETRIES = 2
     env.ENV = 'aat'
   }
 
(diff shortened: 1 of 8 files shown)


Jenkinsfile_nightly
```diff
@@ -85,7 +85,6 @@ withNightlyPipeline(type, product, component) {
   before('fullFunctionalTest') {
     env.PARALLEL_CHUNKS = params.FUNCTIONAL_TESTS_WORKERS
     env.MOCKED_PAYMENTS = false
-    env.TEST_RETRIES = 2
```
tomaszpowroznik, line 88: retries makes it much harder to investigate failures, as our tests are not independednt, and retries mess with data

e2e/pages/caseView.page.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
   async goToNewActions(actionSelected) {
     const currentUrl = await I.grabCurrentUrl();
     await I.retryUntilExists(async () => {
-      if(await I.waitForSelector(this.actionsDropdown, 10) != null) {
+      if(await I.waitForSelector(this.actionsDropdown, 60) != null) {
```
tomaszpowroznik, line 37: give it a little more time to load entire page. Page refresh on aat is really slow (it loads a lot of js files plus ccd call, 10 sec is not enough)

e2e/helpers/puppeter_helper.js
```diff
@@ -43,8 +43,8 @@ module.exports = class PuppeteerHelpers extends Helper {
     }
   }
 
-  async waitForAnySelector(selectors) {
-    return this.waitForSelector([].concat(selectors).join(','));
+  async waitForAnySelector(selectors, maxWaitInSecond) {
```
lewisbirks, line 46: one minor thing to note about this and the base method that it uses, the only selectors that it properly works on is css selectors (see [here](https://github.com/puppeteer/puppeteer/blob/v5.5.0/docs/api.md#pagewaitforselectorselector-options) which links to [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors))

PR #2151, opened 2021-01-27
fix: handle null judgeName for magistrate in draft CMO email (FPLA-NA)

joez-f:
### Change description ###

If hearing judge is a magistrate with no name entered, the draft CMO email will fail to send

https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/DataModel/%7B%22eventId%22:%22a6a1a7d9-607e-11eb-ad1f-497820d12697%22,%22timestamp%22:%222021-01-27T09:04:21.198Z%22%7D/ComponentId/%7B%22Name%22:%22fpl-case-service-appinsights-prod%22,%22ResourceGroup%22:%22fpl-case-service-prod%22,%22SubscriptionId%22:%228999dec3-0104-4a27-94ee-6588559729d1%22%7D

```
[ ] Yes
[X] No
```


service/src/test/java/uk/gov/hmcts/reform/fpl/service/email/content/cmo/DraftCMOUploadedContentProviderTest.java
```diff
@@ -65,4 +65,38 @@ void shouldCreateTemplateWithExpectedParameters() {
...
+
+        DraftCMOUploadedTemplate expected = DraftCMOUploadedTemplate.builder()
+            .judgeName("")
+            .judgeTitle("Justice of the Peace")
+            .respondentLastName("Vlad")
+            .subjectLineWithHearingDate("Vlad, 123456, case management hearing, 20 February 2020")
+            .caseUrl(caseUrl(CASE_NUMBER.toString(), DRAFT_ORDERS))
+            .build();
+
+        assertThat(template).usingRecursiveComparison().isEqualTo(expected);
```
lewisbirks, line 100: don't need to use recursive comparison, the data annotation provides an implementation for equals and hashcode so that when .equals is called it will compare fields instead of reference

lewisbirks (approved):

PR #2182, opened 2021-02-01
fix: Retry e2e test on network issues

tomaszpowroznik:
https://build.platform.hmcts.net/blue/organizations/jenkins/HMCTS_FPL%2Ffpl-ccd-configuration/detail/PR-2151/11/pipeline


lewisbirks (approved):
think there might be another net error that occasionally comes up but I cannot remember which one

PR #2186, opened 2021-02-02
impr: add "Finished" tag to task list (FPLA-2504)

SooryaSu:
[FPLA-2504](https://tools.hmcts.net/jira/browse/FPLA-2504)

### Change description ###
Adds a finished tag to certain tasks once they have had all the questions in their section fully completed

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/FeatureToggleService.java
```diff
@@ -67,6 +67,10 @@ public boolean isEpoOrderTypeAndExclusionEnabled() {
         return ldClient.boolVariation("epo-order-type-and-exclusion", createLDUser(), false);
     }
 
+    public boolean isFinishedTagEnabled() {
```
SooryaSu, line 70: What are the next steps for adding this toggle to LaunchDarkly?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/HearingUrgencyChecker.java
```diff
@@ -26,11 +28,41 @@ public boolean isStarted(CaseData caseData) {
...
+            hearing.getTimeFrame(),
+            hearing.getType(),
+            hearing.getTypeGiveReason(),
+            hearing.getWithoutNotice(),
+            hearing.getReducedNotice(),
+            hearing.getRespondentsAware())) {
+            return false;
+        }
+
+        if (!hearing.getTimeFrame().equals("Within 18 days")
```
SooryaSu, line 52: When you select a time frame such as "Within 7 days" then the reason will be stored under "reason7Days" instead of "reason" in the case data JSON. Selecting "Same day" will have the reason stored under "Reason". 

This behaviour is kind of annoying for trying to check if the user has entered a reason. Should the behaviour be changed so the reason is always stored under "Reason"?

SooryaSu:
Looks like I need to make some changes to the E2E tests, do I need to worry about toggles when updating the tests?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/TaskListRenderer.java
```diff
@@ -149,7 +150,11 @@ public String render(List<Task> allTasks) {
                 lines.add(renderLink(task) + renderImage("in-progress.png", "In progress"));
                 break;
             case COMPLETED:
-                lines.add(renderLink(task) + renderImage("information-added.png", "Information added"));
+                if (useFinishedTag(task)) {
```
samuelzerozero, line 153: Since this method is private and very deep into the code, you can push the whole code into some other class, like TaskListCompletedMarkGenerator(), so that can be unit tested in isolation in TaskListCompletedMarkGeneratorTest instead and mock the component on TaskListRenderTest.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/AllocationProposalChecker.java
```diff
@@ -21,6 +22,14 @@
     public boolean isStarted(CaseData caseData) {
         final Allocation allocationProposal = caseData.getAllocationProposal();
         return isNotEmpty(allocationProposal)
-                && anyNonEmpty(allocationProposal.getProposal(), allocationProposal.getProposalReason());
+            && anyNonEmpty(allocationProposal.getProposal(), allocationProposal.getProposalReason());
+    }
+
+    @Override
```
samuelzerozero, line 28: worth following the same pattern as per AllocationProposalCheckerStartTest (or group the tests with nested) ? In a unit test normally you want to test 1 method only. That would be valid for all the other Checkers

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/CourtServiceChecker.java
```diff
@@ -27,15 +28,44 @@ public boolean isStarted(CaseData caseData) {
...
+            hearingPreferences.getDisabilityAssistance(),
+            hearingPreferences.getExtraSecurityMeasures(),
+            hearingPreferences.getSomethingElse())) {
+            return false;
+        }
+
+        if (("Yes").equals(hearingPreferences.getWelsh())
+            && isEmpty(hearingPreferences.getWelshDetails())) {
+            return false;
+        } else if (("Yes").equals(hearingPreferences.getInterpreter())
```
toireasam, line 54: We have enum YesOrNo.java that could maybe be used here and other places instead of the string "Yes".. we use it like YES.getValue()..

nraghu32 (approved):

PR #2197, opened 2021-02-04
feat: Make DynamiList more generic so it can works with UUID and Strings as a code (FPLA-2604)

tomaszpowroznik:
needed by https://tools.hmcts.net/jira/browse/FPLA-2604


lewisbirks:
Would it be better to use generics? Then we could have the value be anything

PR #2224, opened 2021-02-09
fix: supporting evidence no longer overwritten (FPLA-2679)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-2679

### Change description ###

hearing + c2 supporting evidence can be overwritten by admin/LA (overwrites existing documents from other user), this is a fix for that

list is sorted by date uploaded - documents coming from CMO did not have dateTimeUploaded set so I have set it now

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageDocumentService.java
```diff
@@ -182,19 +182,22 @@
...
+                        new ArrayList<>(element.getValue().getSupportingEvidenceBundle());
+
+                    updateExistingEvidenceWithChanges(existingEvidence, modifiedEvidence);
+
+                    element.getValue().setSupportingEvidenceBundle(existingEvidence);
                 }
-                updateEvidenceBundles.add(element);
             }
-            return updateEvidenceBundles;
+            return hearingFurtherEvidenceBundle;
```
joez-f, line 195: might need to sort list otherwise itll keep swapping it around

joez-f: list now sorted by dateTimeUploaded. if null, put it at the bottom

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageDocumentService.java
```diff
@@ -240,16 +246,21 @@
...
-        List<Element<SupportingEvidenceBundle>> updatedCorrespondenceDocuments =
+        List<Element<SupportingEvidenceBundle>> modifiedEvidence =
             setDateTimeUploadedOnSupportingEvidence(caseData.getSupportingEvidenceDocumentsTemp(),
                 c2DocumentBundle.getSupportingEvidenceBundle());
 
-        List<Element<C2DocumentBundle>> updatedC2Bundles = new ArrayList<>();
-        for (Element<C2DocumentBundle> c2DocumentBundleElement : caseData.getC2DocumentBundle()) {
-            if (selected.equals(c2DocumentBundleElement.getId())) {
-                c2DocumentBundleElement.getValue().setSupportingEvidenceBundle(updatedCorrespondenceDocuments);
+        List<Element<C2DocumentBundle>> updatedC2Bundles = caseData.getC2DocumentBundle();
```
lewisbirks, line 253: vv minor: don't like the name updated here because it is a while in reading terms until it is actually updated

lewisbirks (commented):
you don't seem to have any test to confirm that the sorting works

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageDocumentService.java
joez-f: changed to c2Bundles

lewisbirks (approved):

PR #2227, opened 2021-02-10
chore: Fix HMCTS admin handles supplementary evidence to work on slower envs

tomaszpowroznik:
https://build.platform.hmcts.net/job/HMCTS_Nightly_FPL/job/fpl-ccd-configuration/job/master/574/console

### DIFF
--- a/e2e/pages/caseList.page.js
+++ b/e2e/pages/caseList.page.js
@@ -9,6 +9,7 @@ module.exports = {
     caseState: '#wb-case-state',
     evidenceHandled: '#evidenceHandled-Yes',
     evidenceNotHandled: '#evidenceHandled-No',
+    caseId: 'CCD Case Number',
     caseName: '#caseName',
     search: 'Apply',
     caseList: 'Case list',
@@ -23,10 +24,10 @@ module.exports = {
     I.click(this.fields.search);
   },
 
-  async searchForCasesWithHandledEvidences(submittedAt, state = 'Any') {
+  searchForCasesWithHandledEvidences(caseId, state = 'Any') {
     this.setInitialSearchFields(state);
     I.waitForElement(this.fields.evidenceHandled, 30);
-    await I.fillDate(submittedAt);
+    I.fillField(this.fields.caseId, caseId);
     I.click(this.fields.evidenceHandled);
     I.click(this.fields.search);
   },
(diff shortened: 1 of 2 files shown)


e2e/pages/caseList.page.js
```diff
@@ -53,7 +54,7 @@ module.exports = {
   },
 
   locateCase(caseId){
-    return locate(`//ccd-search-result/table//tr[//a[contains(@href,'${caseId}')]]`);
+    return `a[href$='${caseId}']`;
```
tomaszpowroznik, line 57: retryUntilExists does not work with xpath selectors as used I.waitForSelector does not support xpath

lewisbirks, line 57: Not 100% up to date on this notation but I assume it is still saying a link whose href contains the case number

e2e/pages/caseList.page.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
     caseState: '#wb-case-state',
     evidenceHandled: '#evidenceHandled-Yes',
     evidenceNotHandled: '#evidenceHandled-No',
+    caseId: 'CCD Case Number',
```
lewisbirks, line 12: The others seem to use IDs instead of the text, would it be better to be consistent?

e2e/pages/caseList.page.js
lewisbirks: Forgot to mention this the other day, it is why I like to use wait for element

tomaszpowroznik: the difference is that first is xpath second is css selector
waitForSelector does now work with xpath selectors

e2e/pages/caseList.page.js
tomaszpowroznik: this element has id in a format '[CASE_REFERENCE]' and selector like '#[CASE_REFERENCE]' is not valid and is not recognised by codecept

lewisbirks: fair enough

e2e/pages/caseList.page.js
lewisbirks: ahh it means href that ends in case id, makes sense

lewisbirks (approved):

PR #2238, opened 2021-02-12
feat: upload and manage confidential docs (FPLA-2417)

lewisbirks:

[FPLA-2417](https://tools.hmcts.net/jira/browse/FPLA-2417)

### Change description ###

Added confidential field to supporting evidence bundle that will be visible during the manage docs events when managing further evidence, c2 supporting docs, and correspondence.

Please go over the permissions with a very close eye as I have tied myself in knots multiple times trying to make sure it is right.

The result from the back end is that for:

- `correspondenceDocuments` and `correspondenceDocumentsLA`
- `furtherEvidenceDocuments` and `furtherEvidenceDocumentsLA`

both now have duplicate collections (suffixed with `NC`) that are contain only the non confidential documents.

The `hearingFurtherEvidenceDocuments` and `c2DocumentBundle` now contain further sub-collections from ccd's point of view however from the java point of view there is only a getter for each of them that generate the appropriate collection. The model ignores these fields when it is deserialised so they are remade each time the fields are added to the data map and serialised.

Once this story is merged a migration will be required to populate all the appropriate fields in existing cases so that when the toggle on occurs the users will be able to see all of the current documents in the system.

As of right now when testing you will experience a bug where if you upload further evidence for a hearing or c2 you will overwrite the existing collection (i.e. upload evidence as admin for hearing 1 then do the same as LA) but this should be fixed by #2224 

```
[ ] Yes
[x] No
```


ccd-definition/CaseField/CareSupervision/caseDocuments.json
```diff
@@ -0,0 +1,96 @@
+[
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "hearingFurtherEvidenceDocuments",
+    "FieldType": "Collection",
+    "FieldTypeParameter": "HearingFurtherEvidenceBundle",
+    "Label": "Further evidence documents for hearings",
+    "SecurityClassification": "Public"
```
joez-f, line 9: could add some comments to these fields for more clarity, as it is quite confusing for anyone not familiar

e2e/tests/manageDocuments_test.js
```diff
@@ -0,0 +1,227 @@
...
+  await manageDocumentsEventPage.uploadSupportingEvidenceDocument(supportingEvidenceDocuments[1]);
+  await I.completeEvent('Save and continue');
+  I.seeEventSubmissionConfirmation(config.applicationActions.manageDocumentsLA);
+
+  caseViewPage.selectTab(caseViewPage.tabs.c2);
+
+  I.seeInTab(['C2 Application 1', 'C2 supporting documents 1', 'Document name'], 'Email to say evidence will be late');
+  I.seeInTab(['C2 Application 1', 'C2 supporting documents 1', 'Notes'], 'Evidence will be late');
+  I.seeInTab(['C2 Application 1', 'C2 supporting documents 1', 'File'], 'mockFile.txt');
+  I.seeTextInTab(['C2 Application 1', 'C2 supporting documents 1', ''], 'Confidential');
```
joez-f, line 180: is 'Uploaded by' and 'Date and time uploaded' also shown for C2 supporting? should be I think since it is SupportingEvidenceBundle - if its there lets check it with I.seeTextInTab

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/documents/ManageDocumentsControllerAboutToSubmitTest.java
```diff
@@ -74,106 +77,129 @@ void shouldPopulateHearingFurtherDocumentsCollection() {
         UUID hearingId = UUID.randomUUID();
         HearingBooking hearingBooking = buildFinalHearingBooking();
 
-        Map<String, Object> data = new HashMap<>(Map.of(
-            "hearingDetails", List.of(element(hearingId, hearingBooking)),
-            "manageDocumentsHearingList", DynamicList.builder()
+        CaseData caseData = CaseData.builder()
```
joez-f, line 80: much better 👍

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/documents/ManageDocumentsControllerMidEventTest.java
```diff
@@ -0,0 +1,194 @@
...
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.asDynamicList;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
+import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.wrapElements;
+
+@ActiveProfiles("integration-test")
+@WebMvcTest(ManageDocumentsController.class)
+@OverrideAutoConfiguration(enabled = true)
+public class ManageDocumentsControllerMidEventTest extends AbstractControllerTest {
+
+    private static final String USER_ROLES = "caseworker-publiclaw-courtadmin";
```
joez-f, line 41: could have each role as a static field in AbstractControllerTest instead of specifying the roles in each test, can do it later and update other tests too

lewisbirks: I can do it later, will do along with updating other tests

nraghu32 (approved):

PR #2299, opened 2021-02-24
fix: issue with deleting further evidence docs (INC5167376)

toireasam:
https://tools.hmcts.net/jira/browse/FPLA-2746

### Change description ###
Issue was raised where a user could not delete a further evidence document. 

Issue seems to be in the manage docs event for HMCTS staff.
The old pageShowCondition "supportingEvidenceDocumentsTemp!=\"\"" causes issues when the document is removed and doesn't allow this continue or previous buttons to work. I've changed this page show over to '"manageDocument.type!=\"\"",' as this will redirect to first page when refreshing while also allowing documents to be removed without any issues.

Have also added code to remove the hearing section from documents tab when no documents exist for that hearing anymore

```
[ ] Yes
[ X] No
```


lewisbirks (dismissed):

ccd-definition/CaseTypeTab/CareSupervision/9_Documents.json
```diff
@@ -40,7 +40,7 @@
     "TabDisplayOrder": 9,
     "CaseFieldID": "documentTabFurtherEvidenceHeading",
     "TabFieldDisplayOrder": 4,
-    "FieldShowCondition": "hearingFurtherEvidenceDocuments=\"*\" OR furtherEvidenceDocuments=\"*\" OR furtherEvidenceDocumentsLA=\"*\""
+    "FieldShowCondition": "hearingFurtherEvidenceDocuments!=\"\" OR furtherEvidenceDocuments!=\"\" OR furtherEvidenceDocumentsLA=\"*\""
```
lewisbirks, line 43: Do you need to make the same change to `furtherEvidenceDocumentsLA`?

ccd-definition/CaseTypeTab/CareSupervision/9_Documents.json
```diff
@@ -50,7 +50,8 @@
     "TabLabel": "Documents",
     "TabDisplayOrder": 9,
     "CaseFieldID": "hearingFurtherEvidenceDocuments",
-    "TabFieldDisplayOrder": 5
+    "TabFieldDisplayOrder": 5,
+    "FieldShowCondition": "hearingFurtherEvidenceDocuments!=\"\""
```
lewisbirks, line 54: Isn't this condition the default implementation of CCD?

lewisbirks: Actually I remember, if the list you send is empty then it can show. Maybe it would be best to send  null instead of an empty list 🤔

ccd-definition/CaseTypeTab/CareSupervision/9_Documents.json
toireasam: removed this as discussed when using the caseDetailsMap.putIfNotEmpty this removes need for change here

ccd-definition/CaseTypeTab/CareSupervision/9_Documents.json
toireasam: discussed and removed

lewisbirks (dismissed):

lewisbirks (dismissed):

lewisbirks (approved):

PR #2301, opened 2021-02-24
fix: migration (FPLA-2470)

lewisbirks:

[FPLA-2470](https://tools.hmcts.net/jira/browse/FPLA-2740)

### Change description ###

Allow for more case notes to have been added since data examination

```
[ ] Yes
[ ] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -112,9 +112,9 @@ private void removeFirstDraftCaseManagementOrder(CaseDetails caseDetails) {
     private void removeFirstCaseNotes(CaseDetails caseDetails) {
         CaseData caseData = getCaseData(caseDetails);
 
-        if (isEmpty(caseData.getCaseNotes()) || caseData.getCaseNotes().size() != 4) {
+        if (isEmpty(caseData.getCaseNotes()) || caseData.getCaseNotes().size() < 4) {
```
toireasam, line 115: why do we care if 4 case notes are there? do we not just want to check that at least one case note is present to ensure our caseNote.remove(0) condition will work?

aliveni: wouldn't it remove the latest case note when there are 5 or more?

PR #2314, opened 2021-03-01
chore: update terraform

lewisbirks:
### Change description ###

Update the cnp key vault for terraform see [slack](https://hmcts-reform.slack.com/archives/CA4F2MAFR/p1614607248021900).

Updated keys that could have the interpolation replaced. (can revert if it doesn't work)

```
[ ] Yes
[x] No
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ docker/wiremock/mappings/user-by-email/*
 
 # report outputs
 functional-output/
+
+# terraform modules
+infrastructure/.terraform/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -49,3 +49,6 @@ docker/wiremock/mappings/user-by-email/*
 
 # report outputs
 functional-output/
+
+# terraform modules
+infrastructure/.terraform/
```
tomaszpowroznik, line 54: probably not needed, unless you run terraform on local

tomaszpowroznik (approved):

.gitignore
lewisbirks: I did 😛

PR #2333, opened 2021-03-04
feat: Mandatory email for Applicant Solicitor (FPLA-2824)

SooryaSu:
https://tools.hmcts.net/jira/browse/FPLA-2824

### Change description ###
Makes the solicitor's email a mandatory field when completing the applicant's details section

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantMidEventControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantMidEventControllerTest.java
@@ -64,7 +64,7 @@ class ApplicantMidEventControllerTest extends AbstractCallbackTest {
                         .email(EmailAddress.builder().build())
                         .build())
                     .build()),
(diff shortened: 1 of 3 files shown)


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/ApplicantMidEventControllerTest.java
```diff
@@ -64,7 +64,7 @@ void shouldReturnNoErrorsWhenThereIsNewApplicantAndPbaNumberIsNull() {
                         .email(EmailAddress.builder().build())
                         .build())
                     .build()),
-                "solicitor", Solicitor.builder().build()))
+                "solicitor", Solicitor.builder().email("email@example.com").build()))
```
lewisbirks, line 67: Could we update these tests so that instead of using a map we are using a case data object. The post methods accept case data and it has better type safety over a map which accepts any object.

lewisbirks (dismissed):

lewisbirks (approved):

PR #2361, opened 2021-03-11
chore: Handle ExUI loading spinner if present on the Case List page in the e2e tests

Rob-Doherty:

n/a

### Change description ###

This change introduces a new `waitForSpinnerToFinish()` browser helper to handle the new ExUI loading spinner that can appear while the case list is being generated (see screenshot below).

```
[ ] Yes
[x] No
```

### DIFF
--- a/e2e/helpers/browser_helper.js
+++ b/e2e/helpers/browser_helper.js
@@ -65,7 +65,7 @@ module.exports = class BrowserHelpers extends Helper {
     return this.waitForSelector([].concat(selectors).join(','), maxWaitInSecond);
   }
 
-  async canSee(selector){
+  async canSee(selector) {
     const helper = this.getHelper();
     try {
       const numVisible = await helper.grabNumberOfVisibleElements(selector);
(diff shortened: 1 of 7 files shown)


e2e/helpers/browser_helper.js
```diff
@@ -100,6 +100,31 @@ module.exports = class BrowserHelpers extends Helper {
...
+      // short wait to allow spinner to appear
+      await helper.wait(3);
+    }
+    for (let i = 0; i <= waitTimeout; i++) {
+      spinner = await this.hasSelector(spinnerElement);
+      if (!spinner) {
+        break;
+      }
+      if (i < waitTimeout) {
+        console.log('ExUI loading spinner found, waiting...');
```
lewisbirks, line 120: The preferred way of logging in codecept is using their output object -> https://codecept.io/internal-api/#output

lewisbirks: You can see an example of usage in the steps_file.js

e2e/pages/caseList.page.js
```diff
@@ -44,16 +44,17 @@ module.exports = {
...
     I.waitForVisible(this.fields.caseName, 30);
     I.fillField(this.fields.caseName, caseName);
     I.click(this.fields.search);
     await I.runAccessibilityTest();
   },
 
-  setInitialSearchFields(state='Any') {
+  async setInitialSearchFields(state='Any') {
     // wait for initial filters to load
+    await I.waitForSpinnerToFinish();
```
lewisbirks, line 57: Does the spinning not occur after the search? Or does it do spinning as well whilst populating the search fields as well?

e2e/helpers/browser_helper.js
Rob-Doherty: good shout, I'll change this 👍

e2e/pages/caseList.page.js
Rob-Doherty: The spinning that I saw interfering with the tests wasn't when it was waiting for a case to load, but when the test was using a select dropdown which was already on that page _behind_ a loading spinner. If the test step was waiting for a case to load it seemed to work, I'm thinking because the spinner disappears once the search is complete and thats when the searched case is present.
But it wouldn't be much effort to go with you idea of using a hook to check it's presence before any step

lewisbirks: Ahh it loads the case list when you first go to the page as well using the default filters... ignore this 😄

lewisbirks: The hook might be a nicer idea, that way if they introduce it anywhere else then we can automatically wait for it

lewisbirks: I guess we would need something that is constant across all pages that we can use to check, maybe the case list/create case/sign out links in the banner?

Rob-Doherty:
closing due to neater solution here: https://github.com/hmcts/fpl-ccd-configuration/pull/2363

PR #2363, opened 2021-03-11
chore: Fix spinner issue

tomaszpowroznik:


e2e/helpers/hooks_helper.js
```diff
@@ -22,6 +22,11 @@ module.exports = class HooksHelpers extends Helper {
     });
   }
 
+  _beforeStep() {
+    const helper = this.helpers['Puppeteer'] || this.helpers['WebDriver'];
+    return helper.waitForInvisible('xuilib-loading-spinner', 20);
```
lewisbirks, line 27: Does it need to be attached to the return?

lewisbirks (commented):
Looks like a simplified version of #2361 which should cover all future scenarios if they add it anywhere else

lewisbirks (approved):

e2e/helpers/hooks_helper.js
tomaszpowroznik: yes, it returns promise, and framework that executes _beforeStep must get promise, otherwise it will be waiting in background and wont help with anything

Rob-Doherty (commented):
looks good 👍

Rob-Doherty (approved):

e2e/helpers/hooks_helper.js
lewisbirks: ahh missed that sentence in the helpers readme https://codecept.io/helpers/#hooks

PR #2376, opened 2021-03-15
Test

tomaszpowroznik:


Jenkinsfile_CNP
```diff
@@ -118,6 +118,17 @@ withPipeline(type, product, component) {
 
   after('checkout') {
     sh """ git submodule update --init """
+
+    environment {
+      DEFINITION_UPDATED = sh(script: "git show -- ccd-definition | wc -l", returnStdout: true).trim()
```
lewisbirks, line 123:
```sh
CHANGED=`git diff --name-only $GIT_PREVIOUS_COMMIT $GIT_COMMIT ccd-definition`

if [ -z "$CHANGED" ]; then
    # No changes detected
else
    # Directory changed
fi
```
Thoughts on this, the two vars should be provided by the got Jenkins plugin

PR #2387, opened 2021-03-18
fix: Port gov notify email validation (FPLA-2888)

tomaszpowroznik:


service/src/main/java/uk/gov/hmcts/reform/fpl/service/ValidateEmailService.java
```diff
@@ -1,20 +1,37 @@
...
-import javax.mail.internet.InternetAddress;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
+import static java.lang.String.format;
+import static java.net.IDN.toASCII;
+import static java.net.IDN.toUnicode;
+import static java.util.regex.Pattern.CASE_INSENSITIVE;
+import static java.util.regex.Pattern.compile;
 import static org.apache.logging.log4j.util.Strings.isEmpty;
```
lewisbirks, line 18: not sure we intended to use the logging version of this method

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ValidateEmailServiceTest.java
```diff
@@ -13,10 +12,11 @@
 import java.util.Optional;
 import java.util.stream.Stream;
 
+import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
 
 @ExtendWith(SpringExtension.class)
-@ContextConfiguration(classes = { ValidateEmailService.class })
+@ContextConfiguration(classes = {ValidateEmailService.class})
```
lewisbirks, line 19: Don't think we need to have the spring extension + this, there aren't any dependencies in the service

tomaszpowroznik: done

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ValidateEmailService.java
```diff
@@ -52,15 +69,74 @@ private boolean emailIsOptional(String key) {
...
-            internetAddress.validate();
-            return true;
-        } catch (AddressException addressException) {
+            hostname = toASCII(toUnicode(hostname));
+        } catch (Exception e) {
+            log.warn("Email hostname can not be converted to ascii");
+            return false;
+        }
+
+        final String[] hostParts = hostname.split("\\.");
```
lewisbirks, line 115: Would it be nicer to use `Pattern.quote(".")` instead of the raw string, might make it more readable but don't really mind, if you do it might be worthwhile precomputing it like you do with the patterns.

lewisbirks: Also it is advised to use a limit of -1 to get consistent output of the method when no explicit limit is required, see https://errorprone.info/bugpattern/StringSplitter

lewisbirks (approved):

PR #2388, opened 2021-03-18
Dev/fpla 2809 Respondent Solicitor access to Tabs

imran-hmcts:
https://tools.hmcts.net/jira/browse/FPLA-2828
https://tools.hmcts.net/jira/browse/FPLA-2809

### Change description ###
Part of https://tools.hmcts.net/jira/browse/FPLA-2828 which is done as a part of this PR.

This PR is to focus on structural conversion from old to new json format for roles. This is in prep for giving relevant permissions for respondent solicitors as part of Notice of Change epic. 

```
[ ] Yes
[x] No
```


ccd-definition/AuthorisationCaseField/CareSupervision/solicitor.json
```diff
@@ -3,8 +3,14 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "caseHistory",
-    "UserRole": "[SOLICITOR]",
-    "CRUD": "CRU"
+    "AccessControl": [
```
tomaszpowroznik, line 6: just 
```
"UserRoles": ["[SOLICITOR]"],
"CRUD": "CRU"

should do the job

but anyway i though that it is about adding new roles as well
```
"UserRoles": ["[SOLICITOR1]", "[SOLICITOR2]",...."[SOLICITOR10]"],
    "CRUD": "CRU"

ccd-definition/AuthorisationCaseField/CareSupervision/solicitor.json
```diff
@@ -3,7 +3,7 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "caseHistory",
-    "UserRole": "[SOLICITOR]",
+    "UserRole": ["[SOLICITOR]"],
```
tomaszpowroznik, line 6: UserRole -> UserRoles

ccd-definition/AuthorisationCaseField/CareSupervision/solicitor.json
imran-hmcts: "UserRoles": ["[SOLICITOR]"] implementation is done.

One thing we discussed and tested is that numbers are not allowed in role identifier. [SOLICITOR1], [SOLICITOR_1] etc won't work as '_' and '1' are not allowed. The validation accepts alpha characters only.

The proposed change is to use [SOLICITORONE], [SOLICITORTWO].... unless there are other ideas.

ccd-definition/AuthorisationCaseField/CareSupervision/solicitor.json
imran-hmcts: Done. Missed a commit push but should be there now.

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/CaseRole.java
```diff
@@ -6,6 +6,16 @@
...
+    // Disabling enum naming convention check (Java S115) as ALL CAPS makes this unreadable
+    solicitorA, //NOSONAR
+    solicitorB, //NOSONAR
+    solicitorC, //NOSONAR
+    solicitorD, //NOSONAR
+    solicitorE, //NOSONAR
+    solicitorF, //NOSONAR
+    solicitorG, //NOSONAR
+    solicitorH, //NOSONAR
+    solicitorI, //NOSONAR
```
tomaszpowroznik, line 18: this looks a bit weird, for now we dont store enum directly in case data, but use 

public String formattedName() {
        return formattedName;
    }
    public static CaseRole from(String name) {
        return CaseRole.valueOf(name.replaceAll("[\\[\\]]", ""));
    }
    
    so maybe these names can be normalised, and 2 methods can be adopted to produce correct result

service/src/main/java/uk/gov/hmcts/reform/fpl/enums/CaseRole.java
```diff
@@ -6,21 +6,34 @@
...
 
-    private String formattedName;
+    private final String formattedName;
 
     CaseRole() {
         this.formattedName = String.format("[%s]", name());
     }
 
+    CaseRole(String formattedName) {
+        this.formattedName = formattedName;
```
tomaszpowroznik, line 29: this is not ok, all roles should be formatted as [...] so e.g [solicitorA] and not solicitorA

tomaszpowroznik (approved):

PR #2400, opened 2021-03-22
fix: Allow application document generation when child DoB is in future (FPLA-2896)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2896

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/utils/AgeDisplayFormatHelper.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/utils/AgeDisplayFormatHelper.java
@@ -3,7 +3,6 @@ package uk.gov.hmcts.reform.fpl.utils;
 import java.time.LocalDate;
 import java.time.Period;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 public class AgeDisplayFormatHelper {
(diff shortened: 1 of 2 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/utils/AgeDisplayFormatHelper.java
```diff
@@ -13,27 +12,35 @@ private AgeDisplayFormatHelper() {
...
-        int years = period.getYears();
-        int months = period.getMonths();
-        int days = period.getDays();
+        final Period period = Period.between(dateOfBirth, LocalDate.now());
+        final int years = period.getYears();
+        final int months = period.getMonths();
+        final int days = period.getDays();
 
+        if (period.isNegative()) {
+            return "0 years old";
```
lewisbirks, line 22: Does the last return not cover this?

lewisbirks (changes requested):

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/AgeDisplayFormatHelper.java
tomaszpowroznik: last return says '0 days old' not '0 years old' and is used when child born today

lewisbirks: Could it not be updated to just return the same? Or do we want the explicit difference?

tomaszpowroznik: this is what was agreed with Sara today on standup.

lewisbirks: sure thing

lewisbirks (approved):

PR #2456, opened 2021-04-06
feat: Allow identify local authority based on organisation id (FPLA-2955)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2955

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityService.java
@@ -67,7 +67,7 @@ public class LocalAuthorityService {
     private Optional<String> getLocalAuthorityFromOrganisation() {
         return organisationService.findOrganisation()
             .map(Organisation::getOrganisationIdentifier)
-            .map(orgId -> idsConfig.getLocalAuthorityCode(orgId).orElse(null));
+            .flatMap(idsConfig::getLocalAuthorityCode);
     }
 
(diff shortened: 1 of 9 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/LocalAuthorityService.java
```diff
@@ -31,13 +32,14 @@
...
     private final LocalAuthorityNameLookupConfiguration namesConfig;
+    private final OrganisationService organisationService;
 
     public Optional<String> getLocalAuthorityCode() {
         UserInfo userInfo = idam.getUserInfo(requestData.authorisation());
         String email = userInfo.getSub();
         String domain = extractEmailDomain(email);
 
-        return codesConfig.getLocalAuthorityCode(domain);
+        return codesConfig.getLocalAuthorityCode(domain).or(this::getLocalAuthorityFromOrganisation);
```
lewisbirks, line 42: Would be it be better to rely on PRD as the source of truth first then fall back to our local mapping? Or is this way better due to less network requests or was what business wanted?

tomaszpowroznik: PRD for now has been decided by business to be only fallback solution played in situation when we cannot do correct mapping. Main reason is that relying on PRD reverse order of onboarding process. Currently we can onboard LA that did not register in MO. If we treat PRD as main source of data then org should be first onboarded in PRD before we onboard on our side

lewisbirks (approved):
Looks good

lewisbirks (approved):

PR #2503, opened 2021-04-14
NOJIRA: fix broken func test reports

samuelzerozero:
NOJIRA: fix broken func test reports

Had to downgrade the plugin to resolve the conflict with kotlin and exclude a dependency for the new contract test libraries.

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -4,6 +4,9 @@ buildscript {
   }
   dependencies {
     classpath("net.serenity-bdd:serenity-gradle-plugin:2.4.4")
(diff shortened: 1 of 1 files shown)


service/build.gradle
```diff
@@ -4,6 +4,9 @@ buildscript {
   }
   dependencies {
     classpath("net.serenity-bdd:serenity-gradle-plugin:2.4.4")
+    classpath("gradle.plugin.au.com.dius.pact.provider:gradle:4.2.3") {
```
lewisbirks, line 7: wanna try test the dep bump as well?

samuelzerozero: couldn't resolve/download the plugin with the point version

PR #2508, opened 2021-04-14
FPLA-2842: Include CCD number on all relevant orders

samuelzerozero:
# JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/FPLA-2842

### Change description ###
https://tools.hmcts.net/jira/browse/FPLA-2842

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerMidEventTest.java
```diff
@@ -637,6 +637,7 @@ private void generateEpoValues(CaseData.CaseDataBuilder builder) {
         private CaseData.CaseDataBuilder generateDefaultValues(CaseData.CaseDataBuilder builder) {
             builder.caseLocalAuthority(LOCAL_AUTHORITY_1_CODE);
             builder.familyManCaseNumber("SACCCCCCCC5676576567");
+            builder.id(1234123412341234L);
```
lewisbirks, line 640: Could probs be chained

lewisbirks: and inlined into the return

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsHelper.java
```diff
@@ -2,23 +2,22 @@
...
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 import uk.gov.hmcts.reform.fpl.enums.State;
 
 import static uk.gov.hmcts.reform.fpl.enums.State.GATEKEEPING;
 import static uk.gov.hmcts.reform.fpl.enums.State.OPEN;
 import static uk.gov.hmcts.reform.fpl.enums.State.RETURNED;
 
+@Component
```
lewisbirks, line 13: Any particular reason for making it a component rather than leaving it as a static helper. I guess it does make testing easier as you can then more easily mock.

lewisbirks (commented):
It isn't mentioned in the ticket but do you need to do the cover doc as well?

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsHelper.java
samuelzerozero: yes, static methods are mockable now, but it's rather clunky and need some powermock magic. I personally don't like static, and it should be abolished from the java language. there's no reason why an helper can't be a normal object. :D

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/GeneratedOrderControllerMidEventTest.java
samuelzerozero: inlined 👍

nraghu32 (approved):

PR #2512, opened 2021-04-15
impr: sort applications list in reverse chronological (FPLA-2981)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-2981

### Change description ###

Sort drop down list of applications that appears in several events by reverse chronological order (most recent on top)
Type of order takes first priority in the sort - this means date sorting happens within each type of application

```
[ ] Yes
[X] No
```


lewisbirks (dismissed):
Looks fine to me, there a couple of design decisions that I dislike such as the list creation being in case data and the date being stored as a string but 1 of those we can't update now and the other will probs be out of scope for this PR

lewisbirks (approved):

PR #2517, opened 2021-04-16
NOJIRA: fix oswap exceptions

samuelzerozero:


lewisbirks (commented):
Want to preemptively do the about to submit as well?

Rob-Doherty (approved):
New endpoints, so just were missing from the whitelist file. Risk flagged by OWASP as Low. OK to add.

lewisbirks:
and maybe the initial section as well which has a different callback structure. I don't know why it doesn't detect them all at once because I can almost guarantee the next scan will fail because of it

PR #2548, opened 2021-04-26
fix: hide complex type labels in tab (FPLA-2998)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-2998

### Change description ###

After EXUI changes labels are shown by default in tabs (previously hidden by default), this PR configures some labels to be hidden in tab by default and still show in event

```
[ ] Yes
[X] No
```


ccd-definition/ComplexTypes/CareSupervision/Orders.json
```diff
@@ -107,7 +107,9 @@
     "ListElementCode": "directionLabel",
     "FieldType": "Label",
     "ElementLabel": "## Directions",
-    "SecurityClassification": "Public"
+    "SecurityClassification": "Public",
+    "FieldShowCondition": "directionLabel=\"HIDE_LABEL\"",
+    "Comment": "Needs showing in event, but requires configuring a CaseEventToComplexType for ordersNeeded"
```
joez-f, line 112: this will remove the header from the event, but will hide it from the tab - thoughts? is it that important in event?

ccd-definition/CaseEventToComplexTypes/enterApplicant/Solicitor.json
```diff
@@ -5,7 +5,8 @@
     "CaseEventID": "enterApplicant",
     "CaseFieldID": "solicitor",
     "ListElementCode": "solicitorLabel",
-    "DisplayContext": "OPTIONAL"
+    "DisplayContext": "OPTIONAL",
```
lewisbirks, line 8: READONLY?

ccd-definition/ComplexTypes/CareSupervision/Orders.json
lewisbirks: I would say we should maintain old functionality so even though it is more work it might be worth while making a CEtCT file for it

joez-f: yea will revert for now and handle separately

lewisbirks (dismissed):

lewisbirks (approved):

lewisbirks (approved):

PR #2553, opened 2021-04-27
feat: update respondent solicitor access (FPLA-2840)

tomaszpowroznik:


service/src/test/java/uk/gov/hmcts/reform/fpl/service/NoticeOfChangeServiceTest.java
```diff
@@ -1,29 +1,40 @@
...
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
+import static org.mockito.quality.Strictness.LENIENT;
+import static uk.gov.hmcts.reform.ccd.model.Organisation.organisation;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
 
 @ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = LENIENT)
```
lewisbirks, line 37: not sure the linient is required, you don't appear to be doing any stubbings outside of your test methods so if it is still complaining then you can probably just delete that stubbing

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -10,20 +15,30 @@
...
 
+@ExtendWith(MockitoExtension.class)
 class RespondentServiceTest {
 
-    private final RespondentService service = new RespondentService();
+    @Spy
+    private Time time = new FixedTime();
+
+    @InjectMocks
+    private RespondentService service;
```
lewisbirks, line 41: seeing as it is only 1 parameter I don't think there is any harm in doing
```java
private RespondentService service = new RespondentService(time);
```
and removing the annotations at the same time

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -260,4 +275,250 @@ private RespondentSolicitor buildRespondentSolicitor(String email) {
...
+
+        @Test
+        void shouldReturnEmptyListOfChangesWhenRespondentsNotPresent() {
+            List<ChangeOrganisationRequest> changes = service.getRepresentationChanges(null, null);
+
+            assertThat(changes).isEmpty();
+        }
+
+        @Test
+        void shouldReturnEmptyListOfChangesWhenRespondentsAreEmpty() {
```
lewisbirks, line 289: could be combined with the previous to use a null and empty source parameterised test if you wanted

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
```diff
@@ -10,20 +15,30 @@
...
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.wrapElements;
+import static uk.gov.hmcts.reform.fpl.utils.TestDataHelper.caseRoleDynamicList;
 
+@ExtendWith(MockitoExtension.class)
 class RespondentServiceTest {
 
-    private final RespondentService service = new RespondentService();
+    @Spy
+    private Time time = new FixedTime();
```
lewisbirks, line 38: `private Time time = new FixedTimeConfiguration.stoppedTime();`?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/SystemUserService.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.fpl.config.SystemUpdateUserConfiguration;
+import uk.gov.hmcts.reform.idam.client.IdamClient;
+
+import static uk.gov.hmcts.reform.fpl.config.CacheConfiguration.REQUEST_SCOPED_CACHE_MANAGER;
+import static uk.gov.hmcts.reform.fpl.config.CacheConfiguration.SYS_USER_CACHE;
+
+@Service
+@RequiredArgsConstructor(onConstructor_ = {@Autowired})
+public class SystemUserService {
```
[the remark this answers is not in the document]
tomaszpowroznik: can do it in separate PR, together with introducing cacheing for other user related requesst

service/src/test/java/uk/gov/hmcts/reform/fpl/utils/FixedTime.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.hmcts.reform.fpl.utils;
+
+import uk.gov.hmcts.reform.fpl.service.time.Time;
+
+import java.time.LocalDateTime;
+
+public class FixedTime implements Time {
```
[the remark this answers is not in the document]
tomaszpowroznik: FixedTimeConfiguration.stoppedTime() gives a lambda that cannot be spied do to the final nature. Changed FixedTimeConfiguration to use FixedTime internally

service/src/test/java/uk/gov/hmcts/reform/fpl/service/RespondentServiceTest.java
tomaszpowroznik: but then it is not really extendable, if new dependencies are added then you need to updated constructor (which is sensitive even to order in which dependencies are declared)

e2e/tests/noticeOfChange_test.js
```diff
@@ -54,7 +54,7 @@ Scenario('Hmcts admin removes respondent solicitor', async ({I, caseViewPage, en
   I.see('No cases found.');
 });
 
-const assertRepresentative = async (I, user, organisation) => {
+const assertRepresentative = (I, user, organisation) => {
```
lewisbirks, line 57: method calls still have awaits

lewisbirks (approved):

e2e/tests/accessSegregation_test.js
```diff
@@ -11,9 +11,9 @@ Scenario('Different user in the same local authority can see case created', asyn
   I.see(caseId);
 });
 
-Scenario('Different user in a different local authority cannot see case created', async ({I}) => {
-  await I.navigateToCaseDetailsAs(config.hillingdonLocalAuthorityUserTwo, caseId);
-  I.see('No cases found.');
+Scenario('Different user in a different local authority cannot see case created', async ({I, caseListPage}) => {
+  await I.signIn(config.hillingdonLocalAuthorityUserTwo);
+  caseListPage.verifyCaseIsNotAccessible(caseId);
```
Rob-Doherty, line 16: this is an async function, so could do with an `await`
```suggestion
  await caseListPage.verifyCaseIsNotAccessible(caseId);
```
and same in the other places you've used this refactor

service/src/test/java/uk/gov/hmcts/reform/fpl/service/respondent/RespondentAfterSubmissionValidatorTest.java
```diff
@@ -1,23 +1,36 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.mock;
-import static uk.gov.hmcts.reform.fpl.enums.YesNo.NO;
+import static org.mockito.quality.Strictness.LENIENT;
 import static uk.gov.hmcts.reform.fpl.enums.YesNo.YES;
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.element;
 
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = LENIENT)
```
Rob-Doherty, line 33: how come you're adding `LENIENT` strictness here? is it definitely needed at the class level? would it make sense to limit it to any individual methods that might need it, while retaining the benefits of stricter stubbing in general?

tomaszpowroznik: the problem is that it should be applicable on most of test methods, especially ones that are parametrised. If we have parametrised test has 4 different combinations like in out cases toggle and user role, one one of combination does not use mock then you need to add @MockitoSettings(strictness = LENIENT) on a method or write separate test method.

E.g problem is with if(toggle.isNoCEnabled && userService.isAdmin). when toggle.isNoCEnabled is false then userService.isAdmin is not executed at all and we have unnecesary stubbing warning.
What if we change order  
if(userService.isAdmin && toggle.isNoCEnabled)
if you dont have @MockitoSettings(strictness = LENIENT)  than this will fail as now if userService is false then  isNoCEnabled in not executed at all. But these conditions are logically same and test should pass, but it wont due to unnecessary stubbing. Thats is why i dont like this warning, even if code is same logically your test will start failing.

Dont want to put annotation on method also, because test should be as clean as possible, and focus on scenario and not distract you with not important things

tomaszpowroznik: regarding example with changing order, it may be different when performance is important factor, if e.g toggle is fast but user.isAdmin is slow, then it is better to have toggle condition first, but performance consideration should not have massive impact on functional tests

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/AbstractCallbackTest.java
```diff
@@ -71,16 +71,22 @@ protected AboutToStartOrSubmitCallbackResponse postAboutToStartEvent(CaseData ca
         return postAboutToStartEvent(asCaseDetails(caseData), SC_OK, userRoles);
     }
 
-    protected AboutToStartOrSubmitCallbackResponse postMidEvent(byte[] data, int expectedStatus) {
-        return postEvent(String.format("/callback/%s/mid-event", eventName), data, expectedStatus);
+    protected AboutToStartOrSubmitCallbackResponse postMidEventWithUserRole(CallbackRequest callbackRequest,
```
lewisbirks, line 74: Not sure this needed the name change, the var args means that you could call this method without roles

tomaszpowroznik: for mid event there is overloaded

postMidEvent(callbackrequest, String additionalPath)

if i have 

postMidEvent(callbackrequest, String...roles)

then invocation like postMidEvent(callbackerror,"test") is ambiguous. So 2 options, 
have different method name or change param type to enum

lewisbirks: missed that, makes sense

lewisbirks (approved):

Rob-Doherty (approved):
changes make sense and have been tested. automation coverage is good. ready to release.

PR #2557, opened 2021-04-27
chore: give possiblity to fetch user data for the e2e test purpose, so that user data can be asserted in env agnostic way

tomaszpowroznik:
Usage:

const apiHelper = require('../helpers/api_helper.js');
const userDetails = await apiHelper.getUser(config.privateSolicitorOne);

userDetails looks like:
```
{
  "id": "3e93ba97-ba39-4229-80fd-a14561053eb6",
  "email": "solicitor1@solicitors.uk",
  "forename": "solicitor1@solicitors.uk",
  "surname": "External",
  "roles": [
    "caseworker-publiclaw",
    "pui-case-manager",
    "caseworker",
    "caseworker-publiclaw-solicitor"
  ],
  "fullName": "solicitor1@solicitors.uk External"
}
```

### DIFF
--- a/e2e/helpers/api_helper.js
+++ b/e2e/helpers/api_helper.js
@@ -88,6 +88,17 @@ const createCase = async (user, caseName) => {
   return await response.json();
 };
 
+const getUser = async (user) => {
+  const authToken = await getAuthToken();
+  const url = `${config.fplServiceUrl}/testing-support/user`;
+  const headers = {
+    'Content-Type': 'application/json',
+    'Authorization': `Bearer ${authToken}`,
+  };
+  let response = await post(url, user, headers);
+  return await response.json();
+};
+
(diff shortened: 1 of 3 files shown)


Rob-Doherty (approved):
I like it, should reduce issues with cross-environment test user details.

e2e/helpers/api_helper.js
```diff
@@ -88,6 +88,17 @@ const createCase = async (user, caseName) => {
   return await response.json();
 };
 
+const getUser = async (user) => {
+  const authToken = await getAuthToken(config.systemUpdateUser);
```
lewisbirks, line 92: `config.systemUpdateUser` is the default value so you don't have to pass it

tomaszpowroznik: done

PR #2537, opened 2021-04-22
feat: noc event to add or replace respondent (FPLA-2838)

toireasam:
**Before creating a pull request make sure that:**

- [x] PR's title includes JIRA ticket number and follow [Conventional Commits](https://www.conventionalcommits.org/) specification
- [x] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)

### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/FPLA-2838

### Change description ###
Extended NoC approval event to update respondent solicitor details on `respondent1` with new user details (firstname, lastname, email) and organisation details.

https://user-images.githubusercontent.com/4429628/115738324-db964100-a384-11eb-869a-36246482866c.mov

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #2564, opened 2021-04-28
chore: Check case last event to avoid concurrent data modification by e2e tests

tomaszpowroznik:


service/src/main/java/uk/gov/hmcts/reform/fpl/testingsupport/controllers/TestingSupportController.java
```diff
@@ -84,9 +87,18 @@ public void populateCase(@PathVariable("caseId") Long caseId, @RequestBody Map<S
         }
     }
 
+    @GetMapping("/testing-support/case/{caseId}/lastEvent")
+    public AuditEvent getLastEvent(@PathVariable("caseId") String caseId) {
+        return coreCaseDataApiV2.getAuditEvents(requestData.authorisation(), authToken.generate(), true, caseId)
```
lewisbirks, line 92: Is there a reason we are saying the request is experimental here? In #2537 we saying the request isn't experimental

lewisbirks (commented):
Looks good, I completely forgot that we could do this now that we know about the case audit api. Do we want to extend this functionality to some other tests as well? The create order ones in the slow hmcts admin test all have reties due to send letter stuff as well as, the sdo and draft order stuff.

tomaszpowroznik:
> Looks good, I completely forgot that we could do this now that we know about the case audit api. Do we want to extend this functionality to some other tests as well? The create order ones in the slow hmcts admin test all have reties due to send letter stuff as well as, the sdo and draft order stuff.

Lets trial this approach with Hearing administration scenarios, then we can update rest of test

service/src/main/java/uk/gov/hmcts/reform/fpl/testingsupport/controllers/TestingSupportController.java
tomaszpowroznik: changed to false to match other invocation. But from ccd pov value does not matter, it is just required marker header

PR #2565, opened 2021-04-28
chore: migration to flatten respondents (FPLA-2961)

Jamesnelson117:

https://tools.hmcts.net/jira/browse/FPLA-2961

Migration tool PR
https://github.com/hmcts/fpl-ccd-data-migration-tool/pull/77

### Change description ###
Migration to support upcoming notice of change release.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -85,6 +95,39 @@ private void run2982(CaseDetails caseDetails) {
         caseDetails.getData().put("additionalApplicationsBundle", fixedAdditionalApplicationsBundle);
     }
 
+    private void run2961(CaseDetails caseDetails) {
+        CaseData caseData = getCaseData(caseDetails);
+
+        if (isMigratableState(caseData.getState()) && !containsNoCFields(caseData)) {
+            if (caseData.getRespondents1().size() > 10) {
```
lewisbirks, line 102: If we are going to error out on these cases then you probs want to add it to the filter? That way we won't get any preventable errors in the logs

Jamesnelson117: I ran a query after this comment to see how many cases we have with responents.size() > 10. Results shown we do not have any so this we should not get any. Shall we just remove?

tomaszpowroznik: i think in theory we should migrated such case and have possibility to at least having NoC working for first 10 respondents (we can log warn). But if you checked that it is not a case that is fine

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -85,6 +95,39 @@ private void run2982(CaseDetails caseDetails) {
...
+            }
+
+            Map<String, Object> data = caseDetails.getData();
+
+            data.putAll(respondentPolicyService.generateForSubmission(caseDetails));
+        }
+    }
+
+    private boolean isMigratableState(State state) {
+        List<State> supportedStates = List.of(
```
tomaszpowroznik, line 114: perfectly misleading name :)

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
tomaszpowroznik: I would rather thow exception when isMigratableState(caseData.getState()) && !containsNoCFields(caseData) is not met as this apparently is an error situation as these conditions are check in migration tool

tomaszpowroznik (approved):

PR #2617, opened 2021-05-11
fix: incorrect callback for create-order closed state (FPLA-3041)

joez-f:
### Change description ###

Callback for createOrder-CLOSED mid event is wrong and doesn't match post mapping from controller

should probably add an e2e as well (but let's quick fix for now)

```
[ ] Yes
[X] No
```


SooryaSu (approved):
LGTM

PR #2618, opened 2021-05-11
chore: New api tests (FPLA-2910)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2910


service/src/functionalTest/java/uk/gov/hmcts/reform/fpl/StatelessCallbackApiTest.java
```diff
@@ -1,58 +1,28 @@
...
 
-    @Autowired
-    private AuthenticationService authenticationService;
-
-    @Before
-    public void setUp() {
-        RestAssured.baseURI = testConfiguration.getFplUrl();
-        RestAssured.useRelaxedHTTPSValidation();
-    }
-
```
samuelzerozero, line 55: as discussed, that can be flattened and inline in some stateless abstract class, with the only parameter the input/output to simplify (or something like that)

service/src/main/java/uk/gov/hmcts/reform/fnp/client/PaymentApi.java
```diff
@@ -1,22 +1,33 @@
...
 public interface PaymentApi {
+
     @PostMapping("/credit-account-payments")
     CreditAccountPaymentRequest createCreditAccountPayment(
         @RequestHeader(AUTHORIZATION) String authorisation,
         @RequestHeader(SERVICE_AUTHORIZATION) String serviceAuthorization,
         @RequestBody CreditAccountPaymentRequest creditAccountPaymentRequest
     );
 
+    @GetMapping("/cases/{caseId}/payments")
```
samuelzerozero, line 26: I know it's only for testing, at some point we need to look at the added contract tests we had for other stories. I'll create a ticket in the backlog

service/src/main/java/uk/gov/hmcts/reform/fpl/model/event/MessageJudgeEventData.java
```diff
@@ -22,14 +23,15 @@
...
-        return new String[] {
+        return new String[]{
             "hasAdditionalApplications", "isMessageRegardingAdditionalApplications",
             "additionalApplicationsDynamicList", "relatedDocumentsLabel",
             "nextHearingLabel", "judicialMessageMetaData", "judicialMessageNote", "judicialMessageDynamicList",
             "messageJudgeOption", "judicialMessageReply", "hasJudicialMessages"
         };
     }
 
+    @JsonIgnore
```
samuelzerozero, line 34: was that a bug?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -702,6 +702,7 @@ public boolean hasFutureHearing(List<Element<HearingBooking>> hearingBookings) {
     }
 
     private final DocumentReference submittedForm;
+    private final DocumentReference draftApplicationDocument;
```
samuelzerozero, line 705: interesting. was it never referred?

samuelzerozero (approved):

service/src/main/java/uk/gov/hmcts/reform/fpl/model/event/MessageJudgeEventData.java
tomaszpowroznik: not a bug, but there is no such filed in ccd def, basically MessageJudgeEventData is never put into caseData. 
but in a test when empty CaseData is created and converted to CaseDetails then this filed appears, which is not intended

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
tomaszpowroznik: no, never, it is just put to caseDetails map, but never read back

service/src/main/java/uk/gov/hmcts/reform/fpl/model/event/MessageJudgeEventData.java
samuelzerozero: ok good to know. a bit counterintuitive with @JsonUnwrapped, good to have @jsonIgnore

samuelzerozero (approved):

PR #2619, opened 2021-05-11
impr:  legal team manager application submisson (FPLA-2865)

SooryaSu:
[FPLA-2865](https://tools.hmcts.net/jira/browse/FPLA-2865)

### Change description ###
Adds a new field to record the "legal team manager" for the applicant section of an application, the value is then used in the statement of truth screen and generated C110A document.

```
[ ] Yes
[x] No
```


SooryaSu:
Tweaked a parameter name in the docmosis template from **userFullName** to **signeeName** since it doesn't make much sense if we might use this new field in some cases

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisCaseSubmission.java
```diff
@@ -26,7 +26,7 @@
     private final Allocation allocation;
     private final DocmosisHearingPreferences hearingPreferences;
     private final DocmosisInternationalElement internationalElement;
-    private final String userFullName;
+    private final String signeeName;
```
tomaszpowroznik, line 29: this change requires a bit of coordination how things are released. Docmosis template has been changed to use new filed, but the problem is that release path is different for application code and docmosis.
If code is deployed first and sends signeeName then doc will be generated without name
Same situation when docmosis is deployed first and expects signeeName to be sent but old code sends userFullName
Option is to make template backward compatible for time being. It can use signeeName, but if not present then can use userFullName. Other option is to create new template e.g FL-PLW-APP-ENG-00095v2.doc which uses signeeName, point this code to new template, and release template first.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionController.java
```diff
@@ -119,17 +126,15 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmitEvent(@RequestBod
             data.put("dateAndTimeSubmitted", DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(zonedDateTime));
             data.put("dateSubmitted", DateTimeFormatter.ISO_LOCAL_DATE.format(zonedDateTime));
             data.put("sendToCtsc", setSendToCtsc(data.get("caseLocalAuthority").toString()).getValue());
-            data.put("submittedForm", ImmutableMap.<String, String>builder()
-                .put("document_url", document.links.self.href)
-                .put("document_binary_url", document.links.binary.href)
-                .put("document_filename", document.originalDocumentName)
-                .build());
+            data.put("submittedForm", buildFromDocument(document));
```
tomaszpowroznik, line 129: 👍 +1

service/src/main/java/uk/gov/hmcts/reform/fpl/service/docmosis/CaseSubmissionGenerationService.java
```diff
@@ -266,6 +266,13 @@ private String getThresholdDetails(final Grounds grounds) {
             ? grounds.getThresholdDetails() : DEFAULT_STRING;
     }
 
+    private String getSigneeName(final List<Element<Applicant>> applicants) {
```
tomaszpowroznik, line 269: maybe you can move method to caseSubmissionService and then reuse it here and in controller

service/src/test/java/uk/gov/hmcts/reform/fpl/service/docmosis/CaseSubmissionGenerationServiceTest.java
```diff
@@ -125,6 +125,51 @@ void shouldReturnExpectedCaseNumberInDocmosisCaseSubmissionWhenCaseNumberGiven()
...
+                        .legalTeamManager("legal team manager")
+                        .build())
+                    .build()))
+                .build();
+
+            DocmosisCaseSubmission caseSubmission = templateDataGenerationService.getTemplateData(updatedCaseData);
+            assertThat(caseSubmission.getSigneeName()).isEqualTo("legal team manager");
+        }
+
+        @Test
```
tomaszpowroznik, line 144: you can merge 2 tests into 1 using parametrised test
```
 @ParameterizedTest
        @NullAndEmptySource
        void shouldReturnCurrentUserWhenLegalTeamManagerIsEmptyOrNotPresent(String legalTeamManager) {
            CaseData updatedCaseData = givenCaseData.toBuilder()
                .applicants(wrapElements(Applicant.builder()
                    .party(ApplicantParty.builder()
                        .legalTeamManager(legalTeamManager)
                        .build())
                    .build()))
                .build();

            DocmosisCaseSubmission caseSubmission = templateDataGenerationService.getTemplateData(updatedCaseData);
            assertThat(caseSubmission.getSigneeName()).isEqualTo("Professor");
        }
```

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/CaseSubmissionControllerAboutToStartTest.java
```diff
@@ -64,8 +67,9 @@ void mocking() {
 
     @Test
     void shouldAddConsentLabelToCaseDetails() {
-        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(CaseDetails.builder()
-            .data(of("caseName", "title"))
+        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseData().toBuilder()
```
tomaszpowroznik, line 70: tbh, not big fan of this approach, it suggest that certain things happens for RETURNED state, but it is just technical trick to not trigger certain code. I think it would be better to
add private  method 
```
    private static FeesData feesData(long amount) {
        return FeesData.builder()
            .totalAmount(BigDecimal.valueOf(amount))
            .build();
    }
```

then do:
```
 @Test
    void shouldAddConsentLabelToCaseDetails() {
        given(feeService.getFeesDataForOrders(any())).willReturn(feesData(10));

        AboutToStartOrSubmitCallbackResponse callbackResponse = postAboutToStartEvent(caseData());

        assertThat(callbackResponse.getData())
            .containsEntry("submissionConsentLabel",
                "I, Emma Taylor, believe that the facts stated in this application are true.");
    }
```

you can use new method in other places in this class

service/src/main/java/uk/gov/hmcts/reform/fpl/model/docmosis/DocmosisCaseSubmission.java
SooryaSu: Reverted to userFullName

service/src/main/java/uk/gov/hmcts/reform/fpl/service/casesubmission/CaseSubmissionService.java
```diff
@@ -19,7 +26,15 @@
...
     private final DocmosisDocumentGeneratorService docmosisDocumentGeneratorService;
     private final UploadDocumentService uploadDocumentService;
-    private final CaseSubmissionGenerationService documentGenerationService;
+    private final IdamClient idamClient;
+    private final RequestData requestData;
+
+    private CaseSubmissionGenerationService documentGenerationService;
+
+    @Autowired
+    public void setCaseSubmissionGenerationService(CaseSubmissionGenerationService documentGenerationService) {
```
SooryaSu, line 35: This seems to solve an error message about a circular dependancy between caseSubmissionService and caseSubmissionGenerationService.

Error I was getting before this change below:

image

tomaszpowroznik: was not aware that it requires circular dependencies. You can move getSigneeName to CaseSubmissionGenerationService to avoid it. 

If you want to have it you can use
    @Autowired
    private CaseSubmissionGenerationService documentGenerationService;
    
 setter should not be needed

tomaszpowroznik (approved):

tomaszpowroznik (approved):

PR #2631, opened 2021-05-13
chore: Provide custom implementation from grabAttributes as WebDerive one throws exception when attr not present

tomaszpowroznik:
To support cross browser tests:

from WebDriver.js
```
async grabAttributeFrom(locator, attr) {
    const attrs = await this.grabAttributeFromAll(locator, attr);
    assertElementExists(attrs, locator);
    if (attrs.length > 1) {
      this.debugSection('GrabAttribute', `Using first element out of ${attrs.length}`);
    }
    return attrs[0];
  }
```
and
```
function assertElementExists(res, locator, prefix, suffix) {
  if (!res || res.length === 0) {
    throw new ElementNotFound(locator, prefix, suffix);
  }
}

```

if element is found but attr not present then it throws ElementNotFound exception

### DIFF
--- a/e2e/actors/base.js
+++ b/e2e/actors/base.js
@@ -22,7 +22,7 @@ let currentUser = {};
 
 module.exports = {
   async signIn(user) {
-    if (currentUser !== user) {
+    if (!(this.isPuppeteer() &&  (currentUser === user))) {
       output.debug(`Logging in as ${user.email}`);
       currentUser = {}; // reset in case the login fails
 
--- a/e2e/helpers/browser_helper.js
+++ b/e2e/helpers/browser_helper.js
@@ -7,9 +7,13 @@ module.exports = class BrowserHelpers extends Helper {
     return this.helpers['Puppeteer'] || this.helpers['WebDriver'];
   }
 
+  isPuppeteer(){
+    return this.helpers['Puppeteer'];
+  }
+
   clickBrowserBack() {
     const helper = this.getHelper();
-    if (this.helpers['Puppeteer']) {
+    if (this.isPuppeteer()) {
       return helper.page.goBack();
     } else {
       return helper.browser.back();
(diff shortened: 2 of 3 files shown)


e2e/actors/base.js
```diff
@@ -22,7 +22,7 @@ let currentUser = {};
 
 module.exports = {
   async signIn(user) {
-    if (currentUser !== user) {
+    if (this.isPuppeteer() && currentUser !== user) {
```
tomaszpowroznik, line 25: when run in SauceLab via WebDriver, each test is started in new browser, so this optimalisation cause test failures

e2e/helpers/browser_helper.js
```diff
@@ -100,6 +104,28 @@ module.exports = class BrowserHelpers extends Helper {
...
+    if(this.isPuppeteer()){
+      getAttribute = async (element, attr) =>  (await element.getProperty(attr)).jsonValue();
+    } else {
+      getAttribute = async (element, attr) => await element.getAttribute(attr);
+    }
+
+    const texts = elements.map(async element => getAttribute(element, attr));
+
+    if (texts.length > 1) {
+      throw new Error(`More than one element found for locator ${locator}`);
```
lewisbirks, line 121: It seems that both the [puppeteer](https://github.com/codeceptjs/CodeceptJS/blob/feadbd830c2d85c2f62f65700ebc60638900c58e/lib/helper/Puppeteer.js#L1765) and [webdriver](https://github.com/codeceptjs/CodeceptJS/blob/feadbd830c2d85c2f62f65700ebc60638900c58e/lib/helper/WebDriver.js#L1271) implementation just warn that the first value is being used in the debug log. Do we want to maintain that functionality instead of throwing an error?

e2e/helpers/browser_helper.js
```diff
@@ -100,6 +104,28 @@ module.exports = class BrowserHelpers extends Helper {
...
+    } else if (texts.length === 1) {
+      return texts[0];
+    } else {
+      return undefined;
```
lewisbirks, line 125: I think that this can be reduced to just `return texts[0]`, if the length is 0 then the result is undefined

tomaszpowroznik: done

e2e/helpers/browser_helper.js
tomaszpowroznik: i think that for us it is an error situation, in case multiple elements matches we should prepare specific selector that identifies exact element

lewisbirks (approved):

PR #2634, opened 2021-05-14
Chore/move docmosis tornado docker

samuelzerozero:


tomaszpowroznik (approved):

tomaszpowroznik:
change assumes that fpla-docker is run from particular location (fpla-docker points to fpl-case-service directory using relative path)

PR #2642, opened 2021-05-17
chore: Scroll to tab that is about to be selected, to allow webdriver to select tabs that are initially not in view

tomaszpowroznik:


Jamesnelson117 (approved):
LGTM

PR #2650, opened 2021-05-18
fix: labels in allocation decision tab view

lewisbirks:
### Change description ###

As part of [FPLA-2759](https://tools.hmcts.net/jira/browse/FPLA-2759) it was noticed that the labels that are members of the allocation decision complex type were showing in the tab view went they were not supposed to.
This addresses that issue and also updates the e2e tests to assert that the field was populated during the tests.

#### Before
image

#### After
image

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #2661, opened 2021-05-20
feat: urgent hearing order (FPLA-2759)

lewisbirks:

[FPLA-2759](https://tools.hmcts.net/jira/browse/FPLA-2759)

### Change description ###

Added a futher route to the SDO flow to allow the user to upload an urgent hearing order.
This change also allows the draftSDO event to be accessible from the case management state to allow for users to create and SDO after creating an urgent hearing order.

There are 6 new notifications as part of this change.

### Toggle info

- Event definition toggle controlled by `MultiState-nonprod.json`, toggle makes event available in new state and provides a CYA.
- Event fields definition toggle controlled by `SDORoutes-nonprod.json`, fields are still present in the draftSDO case event to fields file but the route option won't be present so they cannot be accessed.

Number of files changed should reduce slightly when #2650 is merged

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/NotifyTemplates.java
```diff
@@ -44,8 +44,14 @@ private NotifyTemplates() {
     public static final String AMENDED_APPLICATION_RETURNED_CAFCASS_TEMPLATE = "d2e44b62-436c-4bb6-b3b9-afa4f2bca7b5";
     public static final String NOTICE_OF_NEW_HEARING = "c5525d1b-6431-489d-be0e-51de3f163583";
     public static final String SDO_AND_NOP_ISSUED_CTSC = "5ddf3927-f102-4c81-900a-57d8facac1d0";
+    public static final String URGENT_AND_NOP_ISSUED_CTSC = "309c7e39-b1e9-4a30-9d50-fa00d8d87e1c";
+    public static final String SDO_ISSUED_CTSC = "ea694754-d05e-46e2-85e5-f36dd25008fd";
     public static final String SDO_AND_NOP_ISSUED_LA = "6d0316f4-f58f-4ae8-ae25-a344854ac494";
+    public static final String URGENT_AND_NOP_ISSUED_LA = "3314bb72-f4b2-4c59-b67e-83825c44a030";
+    public static final String SDO_ISSUED_LA = "66806269-561c-4ea4-bced-701f0c126e4d";
     public static final String SDO_AND_NOP_ISSUED_CAFCASS = "e6292279-cc27-42d3-92ca-f75da3e18681";
```
joez-f, line 52: personally these should just be one template with a condition to show the NoP content or not, as we basically duplicate 3 templates to include very minor wording changes - but apparently it was problematic with UX to come up with the right wording for a combined template

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/StandardDirectionsOrderController.java
```diff
@@ -187,119 +227,127 @@ public CallbackResponse handleUploadMidEvent(@RequestBody CallbackRequest reques
...
-            caseData,
-            caseDataBefore.getStandardDirectionOrder()
-        );
+        StandardDirectionOrder sdo = sdoService.buildTemporarySDO(caseData, caseDataBefore.getStandardDirectionOrder());
 
-        caseDetails.getData().put(STANDARD_DIRECTION_ORDER_KEY, order);
+        caseDetails.getData().put(STANDARD_DIRECTION_ORDER_KEY, sdo);
+        // work around as I could not get the page hiding when using the [STATE] metadata field
+        caseDetails.getData().put(
+            "showNoticeOfProceedings", YesNo.from(GATEKEEPING == caseData.getState())
```
joez-f, line 235: as mentioned could look into re-using existing 'state' field instead of new case field

ccd-definition/AuthorisationCaseField/CareSupervision/gatekeeper.json
```diff
@@ -2539,5 +2539,61 @@
     "CaseFieldID": "showFurtherEvidenceTab",
     "UserRole": "caseworker-publiclaw-gatekeeper",
     "CRUD": "CRU"
+  },
```
joez-f, line 2542: afaik gatekeeper role is removed from all users, do we even need to update this? in fact can we just delete the file (in a separate PR). guess idam would have be queried first

ccd-definition/CaseEvent/CareSupervision/Gatekeeping-prod.json
```diff
@@ -0,0 +1,18 @@
+[
```
joez-f, line 1: how does this toggle work - so, existing event is available in gatekeeping, but with the additional urgent order - then after toggle, event is available in case management too?

joez-f (commented):
looks generally OK from first check over
one question for future: should superuser be able to remove the urgent order via 'remove order' tool?
also, definition toggle might need looking at

will have another skim through later

ccd-definition/CaseEvent/CareSupervision/MultiState-nonprod.json
```diff
@@ -0,0 +1,19 @@
...
+    "Description": "Add the gatekeeping order",
+    "DisplayOrder": 9,
+    "PreConditionState(s)": "Gatekeeping;PREPARE_FOR_HEARING",
+    "PostConditionState": "*",
+    "SecurityClassification": "Public",
+    "CallBackURLAboutToStartEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/draft-standard-directions/about-to-start",
+    "CallBackURLAboutToSubmitEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/draft-standard-directions/about-to-submit",
+    "CallBackURLSubmittedEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/draft-standard-directions/submitted",
+    "EndButtonLabel": "Save and continue",
+    "ShowEventNotes": "Y",
```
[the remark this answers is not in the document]
lewisbirks: AC 4 requested that it have the CYA, event notes are in the current flow so I kept them as otherwise it would be altering the other flows. If we decide we don't want them in the end I think it can be removed in the rework rather than do them here

ccd-definition/CaseEventToFields/CareSupervision/draftSDO.json
```diff
@@ -325,19 +364,99 @@
...
+    "PageColumnNumber": 1,
+    "PageShowCondition": "sdoRouter!=\"URGENT\"",
+    "ShowSummaryChangeOption": "N"
   },
   {
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseEventID": "draftSDO",
-    "CaseFieldID": "noticeOfProceedings",
+    "CaseFieldID": "showNoticeOfProceedings",
```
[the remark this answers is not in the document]
lewisbirks: I could but I didn't want to have unintended consequences and this allows us to have a field which we can use to add other conditions in the future if required

service/src/main/java/uk/gov/hmcts/reform/fpl/NotifyTemplates.java
lewisbirks: yeah we tried but it was decided that it would be more work to try and make them one template for each user, and then we would want to replace the existing templates as well which would require more work

ccd-definition/AuthorisationCaseField/CareSupervision/gatekeeper.json
lewisbirks: ahh, forgot about that, was just keeping things consistent and was only testing with the gatekeeper-only user. I guess it would be a future clean up

ccd-definition/CaseEvent/CareSupervision/Gatekeeping-prod.json
lewisbirks: So there are 2 toggles for this story, the first is this one which controls the state and the CYA changes, the second is with the fixed list, the new option in the fixed list is toggled so that it won't be selectable by the user. This then allows me to not need to toggle all the other code as the sdoRouter will never have the urgent value selected until the toggled fixed lists are merged

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/StandardDirectionsOrderController.java
lewisbirks: see previous

ccd-definition/CaseEvent/CareSupervision/Gatekeeping-prod.json
joez-f: OK, I trust you to do the toggles properly!

nraghu32 (approved):

PR #2663, opened 2021-05-20
feat: Store applicant name in additional applications (FPLA-2401)

aliveni:
https://tools.hmcts.net/jira/browse/FPLA-2401

### Change description ###
Store applicant name in c2 applications and other applications

```
[ ] Yes
[x] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/UploadAdditionalApplicationsControllerAboutToStartTest.java
```diff
@@ -0,0 +1,70 @@
...
+                .build())
+            .build();
+
+        AboutToStartOrSubmitCallbackResponse response = postAboutToStartEvent(caseData);
+
+        DynamicList actualDynamicList = mapper.convertValue(
+            response.getData().get("applicantsList"), DynamicList.class
+        );
+
+        Assertions.assertThat(actualDynamicList.getListItems())
```
lewisbirks, line 59: Would be nice to assert against the full dynamic list here, code and label, although that may be a struggle with the others. Do you think it would be a bad test if we just didn't populate firstOther 🤭that way we could directly specify all the codes when creating the elements

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadAdditionalApplicationsController.java
```diff
@@ -52,6 +53,17 @@
...
     private final UploadAdditionalApplicationsService uploadAdditionalApplicationsService;
     private final ApplicationsFeeCalculator applicationsFeeCalculator;
+    private final ApplicantsListGenerator applicantsListGenerator;
+
+    @PostMapping("/about-to-start")
+    public AboutToStartOrSubmitCallbackResponse handleAboutToStart(@RequestBody CallbackRequest callbackRequest) {
+        CaseDetails caseDetails = callbackRequest.getCaseDetails();
+        CaseData caseData = getCaseData(caseDetails);
+
+        caseDetails.getData().put("applicantsList", applicantsListGenerator.buildApplicantsList(caseData));
```
lewisbirks, line 63: Although I would assume that it is working can you double check that you don't need to regenerate the dynamic list on mid event calls. There used to be an issue where the dynamic list would be sent back to us as just the selected item code instead of the actual list and as such would need regenerating otherwise it would fail validation in CCD. Going by the fact you are not doing this I am assuming it is now working and as such I think that the data type in case data can be changed from Object to DynamicList

service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/ApplicantsListGenerator.java
```diff
@@ -0,0 +1,70 @@
...
+
+        return dynamicLists.asDynamicList(
+            applicantsFullNames,
+            InterlocutoryApplicant::getCode,
+            InterlocutoryApplicant::getName);
+    }
+
+    private List<InterlocutoryApplicant> buildOthersElements(List<Element<Other>> others) {
+        IncrementalInteger i = new IncrementalInteger(1);
+        return others.stream()
```
lewisbirks, line 53: any reason for a stream rather than a for i loop?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
```diff
@@ -56,7 +69,26 @@ public AdditionalApplicationsBundle buildAdditionalApplicationsBundle(CaseData c
             .build();
     }
 
-    private C2DocumentBundle buildC2DocumentBundle(CaseData caseData, String uploadedBy, LocalDateTime uploadedTime) {
+    private String getSelectedApplicantName(Object applicantsList, String otherApplicant) {
+        DynamicList dynamicList = mapper.convertValue(applicantsList, DynamicList.class);
```
lewisbirks, line 73: Dynamic list service should be able to give you this functionality

service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
```diff
@@ -56,7 +69,26 @@ public AdditionalApplicationsBundle buildAdditionalApplicationsBundle(CaseData c
...
+            DynamicListElement selectedElement = dynamicList.getValue();
+
+            if (isNotEmpty(selectedElement)) {
+                if (APPLICANT_SOMEONE_ELSE.equals(selectedElement.getCode())) {
+                    return otherApplicant;
+                }
+                return selectedElement.getLabel();
+            }
+        }
+        return EMPTY;
```
lewisbirks, line 85: Would it be nicer to return an empty optional instead of an empty string here?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
```diff
@@ -11,40 +12,52 @@
...
 
+    public static final String APPLICANT_SOMEONE_ELSE = "someoneelse";
+
     private final Time time;
     private final DocumentUploadHelper documentUploadHelper;
+    private final ObjectMapper mapper;
 
     public AdditionalApplicationsBundle buildAdditionalApplicationsBundle(CaseData caseData) {
+        final String applicantName = getSelectedApplicantName(
+            caseData.getApplicantsList(), caseData.getOtherApplicant());
```
lewisbirks, line 46: minor formatting niggle of mine but can the final `);` be on a new line. I find it much easier to read. Same with the following below as well
```java
            otherApplicationsBundle = buildOtherApplicationsBundle(
                caseData, applicantName, uploadedBy, currentDateTime);
```

service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
lewisbirks: although looking at what you are doing the you can probs just declare the dynamic list type in case data

lewisbirks (commented):
Looks good, couple of comments for things that could change

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/UploadAdditionalApplicationsControllerAboutToStartTest.java
aliveni: yeah, I wrote the test to assert with both code and label but `firstOther` entry was failing. so I changed it to assert just labels. I think asserting both code and label without `otehrs` is better. I'll update the test

service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
aliveni: updated to return Optional.empty()

service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
```diff
@@ -11,40 +11,54 @@
...
         if (caseData.getAdditionalApplicationType().contains(AdditionalApplicationType.C2_ORDER)) {
-            c2DocumentBundle = buildC2DocumentBundle(caseData, uploadedBy, currentDateTime);
+            c2DocumentBundle = buildC2DocumentBundle(
+                caseData, applicantName.orElse(EMPTY), uploadedBy, currentDateTime);
         }
 
         if (caseData.getAdditionalApplicationType().contains(AdditionalApplicationType.OTHER_ORDER)) {
-            otherApplicationsBundle = buildOtherApplicationsBundle(caseData, uploadedBy, currentDateTime);
+            otherApplicationsBundle = buildOtherApplicationsBundle(
+                caseData, applicantName.orElse(EMPTY), uploadedBy, currentDateTime);
```
lewisbirks, line 61: When adding the applicant name as an empty string in the case that the getSelectedApplicantName cannot find it does that mean we will have a blank field in the tab?

aliveni: The applicant name field shouldn't be empty as it is a mandatory field. If it is empty, the applicant field will not be displayed on the tab

aliveni: @lewisbirks Updated the code to throw an exception when applicant name is empty

lewisbirks (dismissed):

PR #2666, opened 2021-05-21
feat: Remove managing organisation from a case (FPLA-2633)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-2633


ccd-definition/CaseEvent/CareSupervision/MultiState.json
```diff
@@ -720,5 +720,23 @@
...
+    "PreConditionState(s)": "Submitted;Gatekeeping;PREPARE_FOR_HEARING;FINAL_HEARING",
+    "PostConditionState": "*",
+    "SecurityClassification": "Public",
+    "ShowSummary": "Y",
+    "ShowEventNotes": "N",
+    "EndButtonLabel": "Save and continue",
+    "CallBackURLAboutToStartEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/remove-managing-organisation/about-to-start",
+    "CallBackURLAboutToSubmitEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/remove-managing-organisation/about-to-submit",
+    "CallBackURLSubmittedEvent": "${CCD_DEF_CASE_SERVICE_BASE_URL}/callback/remove-managing-organisation/submitted",
+    "EventEnablingCondition": "outsourcingPolicy.Organisation.OrganisationID != \"\""
```
samuelzerozero, line 740: is this field not subject to this bug https://tools.hmcts.net/jira/browse/CCD-1123?

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ManagingOrganisationServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+    private Time time;
+
+    @Mock
+    private OrganisationService organisationService;
+
+    @InjectMocks
+    private ManagingOrganisationService underTest;
+
+    @Test
+    void shouldReturnManagingOrganisation() {
```
samuelzerozero, line 48: could use the @Nested class for each method

samuelzerozero (approved):

service/src/test/java/uk/gov/hmcts/reform/fpl/service/ManagingOrganisationServiceTest.java
tomaszpowroznik: @Nested does not work well with @MethodSource. @MethodSource requires static method on a class. Cannot add static method to nested non static class. Option is to add
@TestInstance(TestInstance.Lifecycle.PER_CLASS):
but then cannot reuse existing static method in different nested classes

ccd-definition/CaseEvent/CareSupervision/MultiState.json
tomaszpowroznik: yes it is, but in this particular setup it works. If outsourcingPolicy does not exists, CCD evaluation fails giving false value to entire expression, which is what we want here. If outsourcingPolicy exists then it is evaluated as expected. 
The problem is if conditions is like:
```
outsourcingPolicy.Organisation.OrganisationID **==**  \"\""
```
then if outsourcingPolicy is missing, CCD due to a bug, evaluates it as false which is not expected in such scenario

lewisbirks (approved):

ccd-definition/CaseEvent/CareSupervision/MultiState.json
samuelzerozero: 👍

nraghu32 (approved):

PR #2667, opened 2021-05-21
fix: case state validator change (FPLA-3052)

toireasam:
https://tools.hmcts.net/jira/browse/FPLA-3052

### Change description ###
change hasFinalFields method to only return true when is final order issue type is null or has final order issued is yes. previously this returned true when has final order issued was no.

```
[ ] Yes
[X ] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/validation/validators/IsStateMigratableValidator.java
```diff
@@ -23,6 +23,6 @@ public boolean isValid(CaseData caseData, ConstraintValidatorContext constraintV
     }
 
     private boolean hasFinalFields(Child child) {
-        return child.getFinalOrderIssuedType() != null || child.getFinalOrderIssued() != null;
+        return child.getFinalOrderIssuedType() != null || "Yes".equals(child.getFinalOrderIssued());
```
lewisbirks, line 26: Has it been checked that `child.getFinalOrderIssuedType() != null` is a valid condition still?

lewisbirks (approved):

service/src/main/java/uk/gov/hmcts/reform/fpl/validation/validators/IsStateMigratableValidator.java
toireasam: yup discussed / tested and believe this condition is still valid / works fine

PR #2670, opened 2021-05-21
fix: remove asterisks from fields that are not mandatory

SooryaSu:
N/A

### Change description ###
Due to an oversight, asterisks were added to the date of birth, gender and address field labels for the Resondent's details although these fields are not mandatory.

This change will remove those asterisks so we do not indicate those fields are mandatory

```
[ ] Yes
[x] No
```


joez-f (approved):

tomaszpowroznik:
tested - looks ok

PR #2673, opened 2021-05-24
chore: bump up gradle (spring upgrade preparation)

tomaszpowroznik:
https://build.platform.hmcts.net/blue/organizations/jenkins/HMCTS_FPL%2Ffpl-ccd-configuration/detail/PR-2665/4/pipeline

Screenshot 2021-05-24 at 14 38 37

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -34,6 +34,9 @@ apply plugin: 'au.com.dius.pact'
 group = 'uk.gov.hmcts.reform'
 version = '1.0.0'
 
+def mirrorHost = System.env.NONPROD_SUBSCRIPTION_NAME == "sandbox" ? "artifactory.sandbox" : "artifactory";
+def mirrorUrl = "https://${mirrorHost}.platform.hmcts.net/artifactory/maven-remotes"
+
 sourceSets {
   functionalTest {
     java {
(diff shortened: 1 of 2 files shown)


service/build.gradle
```diff
@@ -236,14 +236,7 @@ sonarqube {
...
   mavenCentral()
-  // TODO: that's the only dependency to move (confirm with ops when that's migrated) https://tools.hmcts.net/jira/browse/FPLA-2978
-  maven {
-    url "https://dl.bintray.com/hmcts/hmcts-maven"
-  }
-  //maven {
-  //  url "https://jitpack.io"
-  //}
-
+  jcenter()
```
lewisbirks, line 239: @samuelzerozero @tomaszpowroznik I thought we removed jcenter with its deprecation

lewisbirks (commented):
Doesn't look like the wrapper jar itself has updated, to do so you can run
```
./gradlew wrapper --gradle-version 6.9
```

service/build.gradle
tomaszpowroznik: true, but i'm confused, as we still have 
```
maven {
    url "https://dl.bintray.com/hmcts/hmcts-maven"
  }
```
do we know where artefacts has been ported?

btw, jcenter will serve artefacts till 1st Feb 2022

tomaszpowroznik: and i think that all of this works without jcenter only because artefacts are cached in our locals and in jenkins agents.  When i cleared my cache and build then plenty of artefacts are missing

samuelzerozero (approved):

PR #2699, opened 2021-05-28
chore(submodules):Using relative URL for submodule

jvictorcf:
### Change description ###
The idea here is to make it easier to make changes to the submodule easier by using the same URL pattern used for the parent project.
This way, pushing changes will work whether you've used ssh or https to clone the parent project.

For example, I use ssh and therefore don't want to type my password to push changes to fpla-docker.

```
[ ] Yes
[X] No
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "fpla-docker"]
 	path = fpla-docker
-	url = https://github.com/hmcts/fpla-docker
+	url = ../fpla-docker.git
 	branch = master
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "fpla-docker"]
 	path = fpla-docker
-	url = ../fpla-docker.git
+	url = ../../hmcts/fpla-docker.git
 	branch = master


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "fpla-docker"]
 	path = fpla-docker
-	url = https://github.com/hmcts/fpla-docker
+	url = ../fpla-docker.git
```
jvictorcf, line 3: I'd be happy to use "../../hmcts/fpla-docker.git" (for clarity) instead, if the team prefers.

lewisbirks: I think it doesn't hurt to have a bit of clarity with this, so do submodule urls work the same way as file paths?

tomaszpowroznik: if build is green then we are ok, jenkins pipeline rely on submodule and scripts there

tomaszpowroznik (approved):

.gitmodules
jvictorcf: Cool. I change it.

tomaszpowroznik (approved):

PR #2724, opened 2021-06-08
chore: only update Saucelabs at the end of a test run with the overall test result

Rob-Doherty:

https://tools.hmcts.net/jira/browse/FPLA-3049

### Change description ###

Only update Saucelabs at the end of a test run with the overall test result (i.e. any failed test marks the test). 

This is to avoid the current situation where one Saucelabs job is used for all test scenarios, but Codecept updates Saucelabs per scenario. This can result in a mis-match where Saucelabs shows the test run as green if the last scenario passes, even though a previous scenario might have failed.

Also includes:
- update Mac crossbrowser tests to run on Mac OS 11
- fix "Cannot read property 'map' of undefined" error in dump_browser_logs_helper.js
- `waitForElement()` before continuing in `checkTaskStatus()`

image

https://build.platform.hmcts.net/job/HMCTS_Nightly_FPL/job/fpl-ccd-configuration/job/PR-2724/

```
[ ] Yes
[x] No
```

### DIFF
--- a/e2e/crossbrowser/supportedBrowsers.js
+++ b/e2e/crossbrowser/supportedBrowsers.js
@@ -1,4 +1,4 @@
-const LATEST_MAC = 'macOS 10.15';
+const LATEST_MAC = 'macOS 11.00';
 const LATEST_WINDOWS = 'Windows 10';
 
 const supportedBrowsers = {
@@ -25,7 +25,7 @@ const supportedBrowsers = {
   safari: {
     safari_mac_latest: {
       browserName: 'safari',
-      platformName: 'macOS 11.00',
+      platformName: LATEST_MAC,
       browserVersion: 'latest',
       'sauce:options': {
         name: 'FPLA: MAC_SAFARI',
@@ -41,7 +41,7 @@ const supportedBrowsers = {
       browserVersion: 'latest',
       'sauce:options': {
         name: 'FPLA: WIN_CHROME_LATEST',
-        screenResolution: '1400x1050',
+        screenResolution: '1600x1200',
       },
     },
     chrome_mac_latest: {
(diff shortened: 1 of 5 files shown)


e2e/helpers/SauceLabsReportingHelper.js
```diff
@@ -9,21 +9,31 @@ function updateSauceLabsResult(result, sessionId) {
...
-module.exports = function() {
-
-  // Setting test success on SauceLabs
-  event.dispatcher.on(event.test.passed, () => {
-
-    const sessionId = container.helpers('WebDriver').browser.sessionId;
-    exec(updateSauceLabsResult('true', sessionId));
-
+module.exports = function () {
+  let failedTests = [];
```
lewisbirks, line 13: would it be easier to use as [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead?

Rob-Doherty: yep, waaaaay easier 🤦🏻‍♂️ 
will change

e2e/crossbrowser/supportedBrowsers.js
```diff
@@ -41,7 +41,7 @@ const supportedBrowsers = {
       browserVersion: 'latest',
       'sauce:options': {
         name: 'FPLA: WIN_CHROME_LATEST',
-        screenResolution: '1400x1050',
+        screenResolution: '1600x1200',
```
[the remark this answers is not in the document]
Rob-Doherty: I increased the size to provide more info for debugging because we can see more of the screen in the Saucelabs videos, etc.

lewisbirks (approved):

e2e/helpers/SauceLabsReportingHelper.js
Rob-Doherty: done

PR #2730, opened 2021-06-09
FPLA-1986: Docmosis generation testing

samuelzerozero:
FPLA-1986: Docmosis generation testing

This is automating the integration between java and docmosis. There's a similar test in the API testing, but for a more thorough business logic testing (rather than a smoke test), there's a feeling that API testing is not the easiest place (trying to avoid the reverse pyramid testing trouble)

The way it works is to hook for generated orders (the one generated by "Manage orders" event) via the most generic component. 

The test automatically loops over all the Order enums and generate the appropriate mock case data based on the QuestionBlocks configured for the order.

The test test against the stripped code (as per the api test) and stores the pdf generated in the artefacts https://build.platform.hmcts.net/job/HMCTS_FPL/job/fpl-ccd-configuration/job/PR-2730/15/artifact/build/docmosis-generated/

When a new order gets added to the enum list, the developer need to provide a correspondent TXT file with the expected content, plus to implement the mocked data correspondent to the added question (if added)

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/docmosis/AllOrdersDocmosisTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/docmosis/AllOrdersDocmosisTest.java
@@ -31,11 +31,13 @@ import uk.gov.hmcts.reform.fpl.service.orders.generator.C23EPODocumentParameterG
 import uk.gov.hmcts.reform.fpl.service.orders.generator.C32CareOrderDocumentParameterGenerator;
 import uk.gov.hmcts.reform.fpl.service.orders.generator.C33InterimCareOrderDocumentParameterGenerator;
 import uk.gov.hmcts.reform.fpl.service.orders.generator.C35aSupervisionOrderDocumentParameterGenerator;
+import uk.gov.hmcts.reform.fpl.service.orders.generator.C35bISODocumentParameterGenerator;
 import uk.gov.hmcts.reform.fpl.service.orders.generator.C47AAppointmentOfAChildrensGuardianParameterGenerator;
 import uk.gov.hmcts.reform.fpl.service.orders.generator.DocmosisCommonElementDecorator;
 import uk.gov.hmcts.reform.fpl.service.orders.generator.DocumentMerger;
 import uk.gov.hmcts.reform.fpl.service.orders.generator.OrderDocumentGenerator;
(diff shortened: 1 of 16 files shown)


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/docmosis/AllOrdersDocmosisTest.java
```diff
@@ -0,0 +1,143 @@
...
+import static org.mockito.ArgumentMatchers.anyMap;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.when;
+import static uk.gov.hmcts.reform.fpl.docmosis.DocmosisHelper.extractPdfContent;
+import static uk.gov.hmcts.reform.fpl.docmosis.DocmosisHelper.remove;
+import static uk.gov.hmcts.reform.fpl.utils.ResourceReader.readString;
+import static uk.gov.hmcts.reform.fpl.utils.TestDataHelper.testDocument;
+
+@SpringBootTest(classes = {
```
lewisbirks, line 56: I would move `ObjectMapper.class`, `DocmosisDocumentGeneratorService.class`, `RestTemplate.class` and any other common classes to the base class like we do with the notification tests, then to access them in this class you just need to use `@ContextConfiguration` instead of `@SpringBootTest`

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/docmosis/AllOrdersDocmosisTest.java
```diff
@@ -0,0 +1,143 @@
...
+            .isEqualToNormalizingWhitespace(readString(format("order-generation/%s.txt", order.name())));
+    }
+
+    private static Stream<Arguments> allOrders() {
+        return Stream.of(Order.values())
+            .filter(order -> !EXCLUDED_ORDERS.contains(order))
+            .map(Arguments::of);
+    }
+
+    public String getPdfContent(CaseData caseData, String... ignores) {
```
lewisbirks, line 127: this would also probs be worth while in the abstract class along with the results captor

samuelzerozero: the mocking is specific for the class we are testing, not sure should go into the abstract superclass

PR #2743, opened 2021-06-11
Chore: Add screenshots to Mochawesome e2e reports, update Codecept and fix a failing nightly e2e test

Rob-Doherty:

n/a

### Change description ###

1. Add screenshots to Mochawesome e2e reports
2. Make Codecept wait for unique /case-details page `#next-step` element, to fix failing "Local authority deletes application" test in nightly
3. Update Codecept to latest 3.0.7 version

### Example of screenshots in report:
image

```
[ ] Yes
[x] No
```

### DIFF
--- a/e2e/actors/base.js
+++ b/e2e/actors/base.js
@@ -184,7 +184,7 @@ module.exports = {
         await this.goToPage(`${baseUrl}/cases/case-details/${caseId}`);
       }, signedInSelector);
     }
-    await this.waitForSelector('.ccd-dropdown');
+    await this.waitForSelector('#next-step');
   },
 
   async navigateToCaseDetailsAs(user, caseId) {
--- a/e2e/actors/base.js
+++ b/e2e/actors/base.js
@@ -184,7 +184,7 @@ module.exports = {
         await this.goToPage(`${baseUrl}/cases/case-details/${caseId}`);
       }, signedInSelector);
     }
-    await this.waitForSelector('#next-step');
+    this.waitForElement('#next-step');
   },
 
   async navigateToCaseDetailsAs(user, caseId) {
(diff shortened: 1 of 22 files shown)


e2e/actors/base.js
```diff
@@ -184,7 +184,7 @@ module.exports = {
         await this.goToPage(`${baseUrl}/cases/case-details/${caseId}`);
       }, signedInSelector);
     }
-    await this.waitForSelector('.ccd-dropdown');
+    await this.waitForSelector('#next-step');
```
lewisbirks, line 187: Not sure where this wait for selector method is defined on my phone but can we instead use wait for element and remove the async nature?

Rob-Doherty: `waitForSelector()` looks to be a FPLA defined method, but reading the code it looks like it's just doing the equivalent of a `waitForElement()` anyway, just without throwing the error (possibly to allow the `retryUntilExists()` code upstream to handle it instead of Codecept)
```
  async waitForSelector(locator, sec) {
    const helper = this.getHelper();

    try {
      if (this.isPuppeteer()) {
        const waitTimeout = sec ? sec * 1000 : helper.options.waitForTimeout;
        const context = await helper._getContext();
        return await context.waitForSelector(locator, {timeout: waitTimeout});
      } else {
        const waitTimeout = sec ? sec : helper.options.waitForTimeout;
        return await helper.waitForElement(locator, waitTimeout);
      }
    } catch (error) {
      return undefined;
    }
  }
```
So looks like it should be fine to change - I'll do that now

Rob-Doherty: done

lewisbirks (approved):

PR #2763, opened 2021-06-17
Chore: Improve nightly pipeline stability

Rob-Doherty:

n/a

### Change description ###

Multiple tweaks and hardening for the nightly pipeline:
- configure max 2 retries for the nightly cross-browser tests
- increased Saucelabs max test duration to 50mins, to allow for additional retries (and reduced the `delayAfter` to speed up cross-browser tests in general)
- make `navigateToCaseDetails()` wait for a less generic element, to make sure page has loaded more fully. Improves retry logic.
- reduced a long wait in `goToNewActions()`, as this blows out test run times when those steps can end up retrying up to 5 attempts. This is currently getting caught on a [ExUI page load issue](https://tools.hmcts.net/jira/browse/EUI-4210), and refreshing would actually resolve this issue, so no point waiting too long. If the element hasn't appeared in 30 secs it probably won't. 
- added an extra case to `goToNewActions()`'s refresh logic to handle a situation where IE11 can not return a URL if the page hasn't loaded correctly (again suffering from the [ExUI page load issue](https://tools.hmcts.net/jira/browse/EUI-4210))
- moved steps out of `Before()` function for `localAuthoritySubmitsApplication_test.js` to allow Codecept to retry them if they fail - unfortunately Codecept does not appear to support retrying of either `Before()` or `BeforeSuite()` code. (it's more rare to have failures at the `BeforeSuite()` stage)

```
[ ] Yes
[x] No
```

image

### DIFF
--- a/e2e/pages/caseView.page.js
+++ b/e2e/pages/caseView.page.js
@@ -44,7 +44,7 @@ module.exports = {
         I.click(this.goButton);
       } else {
         const newUrl = await I.grabCurrentUrl();
-        if(newUrl === currentUrl){
+        if(newUrl === currentUrl || !newUrl.includes('http')){
           output.print('Page refresh');
           I.refreshPage();
         }
--- a/e2e/pages/caseView.page.js
+++ b/e2e/pages/caseView.page.js
@@ -38,7 +38,7 @@ module.exports = {
   async goToNewActions(actionSelected) {
     const currentUrl = await I.grabCurrentUrl();
     await I.retryUntilExists(async () => {
-      if(await I.waitForSelector(this.actionsDropdown, 60) != null) {
+      if(await I.waitForSelector(this.actionsDropdown, 30) != null) {
         await I.scrollToElement(this.actionsDropdown);
         I.selectOption(this.actionsDropdown, actionSelected);
         I.click(this.goButton);
--- a/saucelabs.conf.js
+++ b/saucelabs.conf.js
@@ -78,7 +78,7 @@ const setupConfig = {
     },
     autoDelay: {
       enabled: true,
-      delayAfter: 2000,
+      delayAfter: 1000,
     },
     screenshotOnFail: {
       enabled: true,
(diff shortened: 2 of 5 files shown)


saucelabs.conf.js
```diff
@@ -77,7 +78,7 @@ const setupConfig = {
     },
     autoDelay: {
       enabled: true,
-      delayAfter: 2000,
+      delayAfter: 1000,
```
lewisbirks, line 81: we have a helper named stepsListener might that work for you instead of this?

e2e/pages/caseView.page.js
```diff
@@ -38,13 +38,13 @@ module.exports = {
...
     await I.retryUntilExists(async () => {
-      if(await I.waitForSelector(this.actionsDropdown, 60) != null) {
+      if(await I.waitForSelector(this.actionsDropdown, 30) != null) {
         await I.scrollToElement(this.actionsDropdown);
         I.selectOption(this.actionsDropdown, actionSelected);
         I.click(this.goButton);
       } else {
         const newUrl = await I.grabCurrentUrl();
-        if(newUrl === currentUrl){
+        if(newUrl === currentUrl || !newUrl.includes('http')){
```
lewisbirks, line 47: guess this is that IE issue you were talking about

saucelabs.conf.js
Rob-Doherty: the `stepListener.js` helper might be useful, but I think that might make sense to investigate in a separate ticket as the delays are far shorter (default 200ms) than what's been set here. So I'd want to run a fair number of test cycles, and knowing how fragile the crossbrwoser pipeline can be I wouldn't be surprised if it led to further tweaking of the timings.

Definitely worth investigating, but given the benefit the other improvements on this branch I reckon it might be best for a separate PR

e2e/pages/caseView.page.js
Rob-Doherty: yeah, that's it, hopefully it'll help - I've not actually been able to hit the same ExUI flakey issue to prove it, but it's not caused any failures either 🙂

lewisbirks (approved):

PR #2782, opened 2021-06-22
bug: nested complex field mandatory even when hidden (EUI-4213)

joez-f:

https://tools.hmcts.net/jira/browse/EUI-4213

### Steps ###

1. log in hmcts-admin@example.com | Password12
2. on any case, go to (TEST) Mandatory hidden field event
3. Use yes/no options and see that continue button is greyed out when selecting no, due to a bug with nested complex type controlled by COMPLEX DisplayContext and MANDATORY set in CaseEventToComplexType

```
[ ] Yes
[ ] No
```


joez-f:
XUI fix has been made

PR #2788, opened 2021-06-24
fix: Make OrderCreationServiceTest not dependant on test order

tomaszpowroznik:


service/src/test/java/uk/gov/hmcts/reform/fpl/service/orders/OrderCreationServiceTest.java
```diff
@@ -43,7 +43,7 @@
     private final OrderCreationService underTest = new OrderCreationService(
         documentGenerator, uploadedOrderDocumentGenerator, uploadService
     );
-    private static final CaseData CASE_DATA = CaseData.builder()
+    private final CaseData CASE_DATA = CaseData.builder()
```
tomaszpowroznik, line 46: class fields (static) should be non mutable. If test mutate such instance, then next one see these changes. When test execution order changes (not guaranteed by junit) then it may lead to unexpected failures

PR #2790, opened 2021-06-24
fix: Fix OrderCreationServiceTests

tomaszpowroznik:


service/src/test/java/uk/gov/hmcts/reform/fpl/service/orders/OrderCreationServiceTest.java
```diff
@@ -35,6 +33,8 @@
     private static final String DRAFT_FILE_NAME = "Preview order.pdf";
     private static final String MEDIA_TYPE = "mock/media_type";
 
+    private final Order order = mock(Order.class);
```
lewisbirks, line 36: makes sense, if we have a static mock then it will remember the mocks from previous tests

lewisbirks: I think this might be an issue in other places as well, I feel like I have probs done this, do you want to look around for them or shall I?

lewisbirks (approved):

service/src/test/java/uk/gov/hmcts/reform/fpl/service/orders/OrderCreationServiceTest.java
tomaszpowroznik: if you have time, then go ahead :)

lewisbirks: to get all constant mocks we can search using the regex `private static final [\w_ ]+ = mock`

PR #2792, opened 2021-06-25
Chore: Move e2e Before() and BeforeSuite() steps into Scenarios

Rob-Doherty:

https://tools.hmcts.net/jira/browse/FPLA-3172

### Change description ###

Move e2e Before() and BeforeSuite() steps into Scenarios because Codecept doesn't retry failures in either stage, so known flakiness in AAT can't be worked around, which results in build failures unrelated to FPLA code.

Also includes a change to reduce the cross-browser test `WAIT_FOR_TIMEOUT` to help avoid hitting Saucelabs `maxDuration` threshold, since steps get retried and so can end up waiting for minutes.

image

```
[ ] Yes
[x] No
```

### DIFF
--- a/e2e/tests/noticeOfChange_test.js
+++ b/e2e/tests/noticeOfChange_test.js
@@ -11,16 +11,6 @@ let caseId;
 
 Feature('Notice of change');
 
-BeforeSuite(async ({I}) => {
-  caseId = await I.submitNewCaseWithData(mandatoryWithMultipleRespondents);
-  solicitor1.details = await apiHelper.getUser(solicitor1);
-  solicitor1.details.organisation = 'Private solicitors';
-  solicitor2.details = await apiHelper.getUser(solicitor2);
-  solicitor2.details.organisation = 'London Borough Hillingdon';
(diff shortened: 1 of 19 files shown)


e2e/tests/noticeOfChange_test.js
```diff
@@ -21,7 +21,24 @@ BeforeSuite(async ({I}) => {
   solicitor3.details.organisation = 'Wiltshire County Council';
 });
 
+async function setupScenario(I) {
```
tomaszpowroznik, line 24: forget to remove BeforeSuite

tomaszpowroznik:
In genera just wondering if it can be kind of semi automated by convention, e.g having listener like

event.dispatcher.on(event.test.started, function (test) {

that check if suite have method called setupScenario and if yes, execute it, but that can be done later

e2e/tests/noticeOfChange_test.js
Rob-Doherty: good spot!

tomaszpowroznik (approved):

PR #2810, opened 2021-06-30
Fix: Nightly security scan stage

Rob-Doherty:

n/a

### Change description ###

The Security Scan step in the nightly pipeline fails since the new ZAP version was updated to Docker, apparently due to the ordering of some of the calls. To solve this it was suggested to move the `zap-cli` calls above the copy calls as there's an issue where the env vars can be lost after something like a copy or mkdir call.

```
[ ] Yes
[x] No
```

### DIFF
--- a/security.sh
+++ b/security.sh
@@ -3,11 +3,13 @@
 export LC_ALL=C.UTF-8
 export LANG=C.UTF-8
 export PYTHONDONTWRITEBYTECODE=1
-
 zap-api-scan.py -t ${URL_FOR_SECURITY_SCAN}/v2/api-docs -f openapi -S -d -u ${SECURITY_RULES} -P 1001 -l FAIL --hook=zap_hooks.py
-cat zap.out
 curl --fail http://0.0.0.0:1001/OTHER/core/other/jsonreport/?formMethod=GET --output report.json
+
+export LC_ALL=C.UTF-8
+export LANG=C.UTF-8
+export PYTHONDONTWRITEBYTECODE=1
 zap-cli --zap-url http://0.0.0.0 -p 1001 report -o api-report.html -f html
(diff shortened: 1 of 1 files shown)


samuelzerozero (dismissed):

security.sh
```diff
@@ -3,11 +3,13 @@
 export LC_ALL=C.UTF-8
 export LANG=C.UTF-8
 export PYTHONDONTWRITEBYTECODE=1
-
 zap-api-scan.py -t ${URL_FOR_SECURITY_SCAN}/v2/api-docs -f openapi -S -d -u ${SECURITY_RULES} -P 1001 -l FAIL --hook=zap_hooks.py
-cat zap.out
```
tomaszpowroznik, line 8: why it is removed, does it tell us what went wrong?

security.sh
```diff
@@ -3,11 +3,13 @@
...
 curl --fail http://0.0.0.0:1001/OTHER/core/other/jsonreport/?formMethod=GET --output report.json
+
+export LC_ALL=C.UTF-8
```
tomaszpowroznik, line 9: same on lines 3-5

tomaszpowroznik:
it is not clear what this PR tries to solve, maybe small comment?

security.sh
Rob-Doherty: replaced

security.sh
Rob-Doherty: I was led to believe that the script needed the duplication due to a recent ZAP docker update, and my testing seemed to bear that out too - but on re-trying this morning I'm getting different results.
I'd seen some other Jenkins issues late yesterday, so that might've skewed my results, as I can now see the script working as expected.
I've changed this to just log those env vars for future debugging

Rob-Doherty:
@tomaszpowroznik I've re-written the description, hopefully that makes it clearer

samuelzerozero (approved):

PR #2822, opened 2021-07-01
fix: flip condition that is causing NPE after EXUI release (FPLA-3193)

joez-f:

https://tools.hmcts.net/jira/browse/FPLA-3193

### Change description ###

PreviousHearingVenue complex type was presumably null before XUI release 22.1, now it is an empty object which is causing our if statement to throw NPE

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageHearingsService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageHearingsService.java
@@ -249,7 +249,7 @@ public class ManageHearingsService {
 
     public void findAndSetPreviousVenueId(CaseData caseData) {
         if (isNotEmpty(caseData.getHearingDetails()) && caseData.getPreviousHearingVenue() != null
-            && caseData.getPreviousHearingVenue().getUsePreviousVenue().equals(YES.getValue())) {
+            && YES.getValue().equals(caseData.getPreviousHearingVenue().getUsePreviousVenue())) {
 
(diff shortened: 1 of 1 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageHearingsService.java
```diff
@@ -249,7 +249,7 @@ public HearingBooking getHearingBooking(UUID id, List<Element<HearingBooking>> h
 
     public void findAndSetPreviousVenueId(CaseData caseData) {
         if (isNotEmpty(caseData.getHearingDetails()) && caseData.getPreviousHearingVenue() != null
-            && caseData.getPreviousHearingVenue().getUsePreviousVenue().equals(YES.getValue())) {
+            && YES.getValue().equals(caseData.getPreviousHearingVenue().getUsePreviousVenue())) {
```
lewisbirks, line 252: Could `caseData.getPreviousHearingVenue()` have a chance to be null?

joez-f: well if it is then the condition will exit before this part, we check that it's not null

lewisbirks: Completely missed that, too early 😅

lewisbirks (approved):

joez-f:
closing, fix is in other PR

PR #2827, opened 2021-07-02
Close cookie dialog

tomaszpowroznik:


e2e/actors/base.js
```diff
@@ -73,6 +74,13 @@ module.exports = {
     }
   },
 
+  async acceptCookies() {
+    const acceptCookieSelector = '//button[@name="cookies"]';
+    if (await this.hasSelector(acceptCookieSelector)) {
```
Rob-Doherty, line 79: would it be worth using a check with a short timeout, to handle scenarios where the page/banner might take a moment to load, like on AAT in the nightly pipeline?

Maybe something like
```
    if (await this.waitForSelector(acceptCookieSelector, 3)) {
...
```

tomaszpowroznik: that would slow down things, even if cookie is not there

Rob-Doherty (approved):
👍

PR #2852, opened 2021-07-08
fix: Cannot add hearing in a past (FPLA-3214)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-3214

Steps to reproduce:

1. Add hearing in future as first hearing in a case
2. Vacate this hearing (without relisting)
3. Create new hearing, should be able to provide date in past, but system complains date must be in future

Issue description:
Service allowed date in past only when new hearing is added (not edited or relisted). To determine if this is new hearing filed 'hearingOption' is used. This field contains radio button selection (ADD|EDIT|VACATE|ADJOURN|RELIST). This filed was supposed to be cleaned each time new event is started (in about to start callback) but with recent xui changes filed is preserved (bug) which cause that system sees current action as VACATE and not ADD, and does not allow date to be in past.

As a solution flag should be reset after event submission and not at the start of new event, but this approach wont solve cases that already have it saved. To solve it other field - is first hearing is used to determine if action is about adding new hearing.

### DIFF
--- a/ccd-definition/CaseEventToFields/CareSupervision/manageHearings.json
+++ b/ccd-definition/CaseEventToFields/CareSupervision/manageHearings.json
@@ -16,18 +16,6 @@
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseEventID": "manageHearings",
-    "CaseFieldID": "firstHearingFlag",
-    "PageFieldDisplayOrder": 1,
-    "DisplayContext": "READONLY",
-    "PageID": "SelectHearingOption",
-    "ShowSummaryChangeOption": "Y",
-    "PageDisplayOrder": 1,
-    "FieldShowCondition": "hearingOption=\"DO_NOT_SHOW\""
-  },
-  {
-    "LiveFrom": "01/01/2017",
-    "CaseTypeID": "CARE_SUPERVISION_EPO",
-    "CaseEventID": "manageHearings",
     "CaseFieldID": "hasFutureHearingDateFlag",
     "PageFieldDisplayOrder": 1,
     "DisplayContext": "READONLY",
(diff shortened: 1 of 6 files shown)


ccd-definition/CaseEventToFields/CareSupervision/manageHearings.json
```diff
@@ -371,6 +359,18 @@
     "FieldShowCondition": "hearingStartDate=\"DO_NOT_SHOW\"",
     "PageDisplayOrder": 5
   },
+  {
+    "LiveFrom": "01/01/2017",
```
tomaszpowroznik, line 363: moved to next page. XUI works in a weird way, if there is hidden filed and it is used on a page further in a flow then its value is gone. To keep this filed for longer, hidden field is moved a bit

lewisbirks (approved):

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/ManageHearingsController.java
```diff
@@ -356,7 +357,9 @@ private static JudgeAndLegalAdvisor setAllocatedJudgeLabel(Judge allocatedJudge)
     }
 
     private boolean isAddingNewHearing(CaseData caseData) {
-        return isEmpty(caseData.getHearingOption()) || NEW_HEARING.equals(caseData.getHearingOption());
+        return isEmpty(caseData.getHearingOption())
+            || NEW_HEARING.equals(caseData.getHearingOption())
+            || YesNo.fromString(caseData.getFirstHearingFlag()) == YES;
```
Rob-Doherty, line 362: this new logic needs a unit test(s) to cover it

ccd-definition/CaseEventToFields/CareSupervision/manageHearings.json
Rob-Doherty: is there an e2e that needs updating because of this?

PR #2886, opened 2021-07-16
Chore: Improve nightly e2e: allow more nightly retries, wait for task status and scroll to task being checked for easier visual debugging

Rob-Doherty:

https://tools.hmcts.net/jira/browse/FPLA-3172

### Change description ###

1. Increased nightly pipeline test retries, to work around some flakey AAT issues in FPLA's dependencies. 
2. Wait for task status rather than checking as soon as the task element appears, because there can be a delay with displaying the task's status which was causing flakiness.
3. Scroll down to the task status being checked, so to help with debugging failures at this step visually.

```
[ ] Yes
[x] No
```


e2e/pages/caseView.page.js
```diff
@@ -71,12 +71,13 @@ module.exports = {
   },
 
   checkTaskStatus(task, status) {
-    I.waitForElement(locate(`//p/a[text()="${task}"]`), 10);
+    const taskElement = `//p/a[text()="${task}"]`;
+    I.waitForElement(locate(taskElement), 10);
+    I.scrollIntoView(taskElement);
     if(status) {
-      I.seeElement(locate(`//p/a[text()="${task}"]/../img`).withAttr({title: status}));
+      I.waitForElement(locate(`${taskElement}/../img`).withAttr({title: status}), 10);
```
lewisbirks, line 78: both have the `/../img`, could just be a new variable if you want

e2e/pages/caseView.page.js
```diff
@@ -71,12 +71,13 @@ module.exports = {
   },
 
   checkTaskStatus(task, status) {
-    I.waitForElement(locate(`//p/a[text()="${task}"]`), 10);
+    const taskElement = `//p/a[text()="${task}"]`;
```
lewisbirks, line 74: I know that you didn't write this but if you wanted to have codecept build the locator for you you could write this as
```js
locate('p').withChild(locate('a').withText(task));
```
see https://codecept.io/locators/ for more on how that works

lewisbirks (approved):

e2e/pages/caseView.page.js
Rob-Doherty: yeah I was thinking the same, but it turns out that the `/../img/` part is related to the status icon, so for the `I.dontSeeElement()` step which is checking that the _status_ isn't displayed we still want to know that the _task_ is present - which happens on line 75.
otherwise the check could pass on a totally different page or if the page hadn't finished loading

e2e/pages/caseView.page.js
Rob-Doherty: yeah that's a fair point - is there a team preference on that format? personally I prefer using the raw xpath just because I feel I've got more control and less magic happening, but I'd noticed the Codecept locators and happy to follow that pattern if it's preferred?

Rob-Doherty: it looks to be about split in the e2e code, so i'll merge it in and if we want to go with one or other pattern i'm happy to update in a separate PR

PR #2890, opened 2021-07-19
fix: child solcitior e2e tests

lewisbirks:
### Change description ###

Address lookup occasionally couldn't find the manual link when using the locator so changing to just search for the text value instead.

Occasionally `const numOfElements = await I.grabNumberOfVisibleElements(selectedTable);` would return 0 when it is supposed to return 1. Tried adding additional waits to make sure that it is going to be present when we want it to be.

```
[ ] Yes
[x] No
```


Rob-Doherty (approved):
looks good to me 👍

PR #2924, opened 2021-07-23
NOJIRA: Make Manage orders test less failure prone (now retryable)

samuelzerozero:


e2e/tests/adminManagesOrders_test.js
```diff
@@ -81,12 +83,15 @@ Scenario('Create 32b discharge of care order', async ({ I, caseViewPage, manageO
...
     orderIndex: 1,
     orderType: 'Discharge of care order (C32B)',
-    approvalDate: today,
+    approvalDate: approvalDate,
   });
 });
 
 Scenario('Create EPO order', async ({ I, caseViewPage, manageOrdersEventPage }) => {
-  await setupScenario(I, caseViewPage);
+  const newCaseId = await I.submitNewCaseWithData(caseData);
```
lewisbirks, line 91: What is the reason for submitting a new case here?

e2e/tests/adminManagesOrders_test.js
```diff
@@ -118,7 +123,10 @@ Scenario('Create EPO order', async ({ I, caseViewPage, manageOrdersEventPage })
 });
 
 Scenario('Create EPO Prevent removal order', async ({ I, caseViewPage, manageOrdersEventPage }) => {
-  await setupScenario(I, caseViewPage);
+  const newCaseId = await I.submitNewCaseWithData(caseData);
```
lewisbirks, line 126: what is the reason for submitting a new case here?

e2e/tests/adminManagesOrders_test.js
```diff
@@ -81,12 +83,15 @@ Scenario('Create 32b discharge of care order', async ({ I, caseViewPage, manageO
   assertOrder(I, caseViewPage, {
     orderIndex: 1,
     orderType: 'Discharge of care order (C32B)',
-    approvalDate: today,
+    approvalDate: approvalDate,
   });
 });
 
 Scenario('Create EPO order', async ({ I, caseViewPage, manageOrdersEventPage }) => {
```
lewisbirks, line 90: this one still uses today as the approval date time, want to change that to approvalDate as well?

e2e/tests/adminManagesOrders_test.js
```diff
@@ -118,7 +123,10 @@ Scenario('Create EPO order', async ({ I, caseViewPage, manageOrdersEventPage })
 });
 
 Scenario('Create EPO Prevent removal order', async ({ I, caseViewPage, manageOrdersEventPage }) => {
```
lewisbirks, line 125: this one still uses today as the approval date time, want to change that to approvalDate as well?

samuelzerozero: For both EPO had to keep the same dates, as this is constraint to the present and 8 days margin, so I can't use the sequence. In this way I put it in their own case, so don't need too much magic

e2e/tests/adminManagesOrders_test.js
samuelzerozero: For both EPO had to keep the same dates, as this is constraint to the present and 8 days margin, so I can't use the sequence. In this way I put it in their own case, so don't need too much magic

e2e/tests/adminManagesOrders_test.js
samuelzerozero: For both EPO had to keep the same dates, as this is constraint to the present and 8 days margin, so I can't use the sequence. In this way I put it in their own case, so don't need too much magic

e2e/tests/adminManagesOrders_test.js
samuelzerozero: For both EPO had to keep the same dates, as this is constraint to the present and 8 days margin, so I can't use the sequence. In this way I put it in their own case, so don't need too much magic

PR #2925, opened 2021-07-23
feat: C2 notification should show next hearing (not first) (FPLA-3194)

SooryaSu:
[FPLA-3194](https://tools.hmcts.net/jira/browse/FPLA-3194)

### Change description ###
The C2 notification sent to CTSC will now show the next hearing rather than the first hearing in a case.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/utils/EmailNotificationHelper.java
```diff
@@ -81,6 +82,12 @@ public static String buildCallout(final CaseData caseData) {
             caseData.getFirstHearing().orElse(null));
     }
 
+    public static String buildCalloutWithNextHearing(final CaseData caseData) {
```
lewisbirks, line 85: instead of having a static method could it be instanced, and then we can autowire time in and use that to compare instead of LocalDateTime.now()

lewisbirks: actually there is a method that accommodates this already, `buildCalloutWithNextHearing` that accepts `CaseData` and a `LocalDataTime`

SooryaSu: A small problem with using that function is that it adds a "^" which will result in the email having a "^^" resulting in a double callout due to the template already having a "^"

lewisbirks (approved):

PR #2976, opened 2021-08-09
Fpla 2850 notification respondent sol upload

meganaslattery:

https://tools.hmcts.net/jira/browse/FPLA-2850

### Change description ###

Send notification when solicitor uploads a document

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/FurtherEvidenceUploadedEventHandler.java
```diff
@@ -50,16 +57,33 @@ public void handleDocumentUploadedEvent(final FurtherEvidenceUploadedEvent event
...
-                && !d.isConfidentialDocument());
+                && !d.isConfidentialDocument())
+            .map(d -> d.getName())
+            .collect(Collectors.toList());
+    }
+
+    private List<Element<SupportingEvidenceBundle>> getEvidenceBundle(CaseData caseData, boolean uploadedByLA) {
+        if (uploadedByLA) {
+            return caseData.getFurtherEvidenceDocumentsLA();
+        } else if (isSolicitor(caseData.getId())) {
```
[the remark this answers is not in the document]
lewisbirks: You could if you wanted have a small private enum in the class specifying the user type, something like
```java
    private enum UserType {
        LA, SOLICITOR, OTHER
    }
```
which can then be used as a parameter to the methods instead
```java
    private List<Element<SupportingEvidenceBundle>> getEvidenceBundle(CaseData caseData, UserType type) {
        if (UserType.LA == type) {
            return caseData.getFurtherEvidenceDocumentsLA();
        } else if (UserType.SOLICITOR == type) {
            return caseData.getFurtherEvidenceDocumentsSolicitor();
        } else {
            return caseData.getFurtherEvidenceDocuments();
        }
    }
```
where the parameter is calculated like so
```java
        UserType type = uploadedByLA ? UserType.LA
                                     : (isSolicitor(caseData.getId()) ? UserType.SOLICITOR : UserType.OTHER);
```

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/FurtherEvidenceUploadedEventHandler.java
```diff
@@ -10,38 +10,45 @@
...
+        List<Element<SupportingEvidenceBundle>> newBundle = getEvidenceBundle(caseData, uploadedByLA);
+        List<Element<SupportingEvidenceBundle>> oldBundle = getEvidenceBundle(caseDataBefore, uploadedByLA);
+        List<String> newNonConfidentialDocuments = getNewNonConfidentialDocuments(newBundle, oldBundle);
+
         final Set<String> recipients = new HashSet<>();
 
-        if (event.isUploadedByLA()) {
-            if (hasNewNonConfidentialDocuments(caseData.getFurtherEvidenceDocumentsLA(),
-                caseDataBefore.getFurtherEvidenceDocumentsLA())) {
+        if (uploadedByLA) {
```
lewisbirks, line 46: I would flip the ordering of these conditions, you are calling an if on `!newNonConfidentialDocuments.isEmpty()` twice so that could be the outer if and then have the `addAll` determined by `uploadedByLA`

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/FurtherEvidenceUploadedEventHandler.java
```diff
@@ -50,16 +57,33 @@ public void handleDocumentUploadedEvent(final FurtherEvidenceUploadedEvent event
...
-            furtherEvidenceNotificationService.sendNotification(caseData, recipients, uploader.getFullName());
+            furtherEvidenceNotificationService.sendNotification(caseData, recipients, uploader.getFullName(),
+                newNonConfidentialDocuments);
         }
     }
 
-    private static boolean hasNewNonConfidentialDocuments(List<Element<SupportingEvidenceBundle>> newEvidenceBundle,
-                                                          List<Element<SupportingEvidenceBundle>> oldEvidenceBundle) {
-        List<SupportingEvidenceBundle> oldEvidenceBundleUnwrapped = unwrapElements(oldEvidenceBundle);
+    private List<String> getNewNonConfidentialDocuments(List<Element<SupportingEvidenceBundle>> newEvidenceBundle,
```
lewisbirks, line 65: would rename this to `getNewNonConfidentialDocumentsNames` or something like that, we are not getting the actual docs

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FurtherEvidenceNotificationServiceTest.java
```diff
@@ -187,4 +188,11 @@ private CaseData caseData(Representative... representatives) {
             .representatives(wrapElements(representatives))
             .build();
     }
+
+    private static List<String> buildDocumentsNamesList() {
+        List<String> documents = new ArrayList<>();
```
lewisbirks, line 193: java 11 has a handy util function here, if you want to construct a list of pre-determined objects then there is the `List.of` method, similar exists for maps and sets. Note that the returned list is immutable, so it won't work when the list is going to be updated

lewisbirks (commented):
The ticket mentions a toggle, have you considered how you are going to do that? I don't see one unless I am missing the obvious

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/FurtherEvidenceUploadedEventHandler.java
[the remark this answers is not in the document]
meganaslattery: Thank you for your comments. They are really helpful!

I like the idea of using an enum here, but I'm not sure if it's necessary – we are only making one call to the ccd-access-data-store-api overall.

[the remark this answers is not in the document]
lewisbirks: We are making 2 as you call this method for both caseData and caseData before, but it is a vv minor improvement which could also be resolved with something like caching.

lewisbirks (approved):

lewisbirks (dismissed):

lewisbirks (approved):

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/documents/ManageDocumentsControllerSubmittedTest.java
```diff
@@ -25,36 +31,72 @@
...
+        when(idamClient.getUserDetails(any())).thenReturn(UserDetails.builder().build());
+        when(caseAccessDataStoreApi.getUserRoles(any(), any(), any(), any()))
+            .thenReturn(buildCaseAssignedUserRole("[SOLICITORA]"));
+
+        postSubmittedEvent(buildCallbackRequest(SOLICITOR_BUNDLE_NAME, true));
+        verifyNoInteractions(notificationClient);
+    }
+
+    @Test
+    void shouldPublishEventWithSolUserWhenUploadNotificationFeatureIsEnabled() throws NotificationClientException {
```
SooryaSu, line 94: Might be better for readability to have "SolUser" as "Solicitor"

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/handlers/FurtherEvidenceUploadedEventHandlerEmailTemplateTest.java
```diff
@@ -43,51 +50,29 @@
     private static final LocalDateTime HEARING_DATE = LocalDateTime.of(2021, 2, 22, 0, 0, 0).plusMonths(3);
     private static final String RESPONDENT_LAST_NAME = "Smith";
     private static final String CHILD_LAST_NAME = "Jones";
+    private static final UUID REPRESENTATIVE_UUID = UUID.randomUUID();
+    private static final CaseData caseData = buildCaseData();
+    private static final CaseData caseDataBefore = buildCaseDataBefore();
```
SooryaSu, line 55: These look like they should be CASE_DATA and CASE_DATA_BEFORE for consistency

SooryaSu (dismissed):

lewisbirks (dismissed):

service/src/test/java/uk/gov/hmcts/reform/fpl/service/FurtherEvidenceNotificationServiceTest.java
```diff
@@ -225,6 +273,14 @@ private CaseData caseData(Representative... representatives) {
             .build();
     }
 
+    private CaseData caseData(Child children, Respondent respondents) {
```
lewisbirks, line 276: what is the point of passing a child and respondent here?

lewisbirks: not sure that the method is even used

meganaslattery: Thank you for the review. It's because I have written a test which needs casedata to have children and respondent fields :)

lewisbirks (approved):

PR #2980, opened 2021-08-09
fix: Fix Others event to save non-empty others (FPLA-3303)

aliveni:
https://tools.hmcts.net/jira/browse/FPLA-3303

### Change description ###
- Update "others to be given notice" event to save `first other` only if details are entered.
- Update NoH Notifications to send "No address added to Other" email only if other's name is not empty.

```
[ ] Yes
[x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/Others.java
```diff
@@ -25,13 +26,13 @@ public static Others from(List<Element<Other>> allOthers) {
 
         final LinkedList<Element<Other>> others = new LinkedList<>(Optional.ofNullable(allOthers).orElse(emptyList()));
         return Others.builder()
-            .firstOther(ofNullable(others.pollFirst()).map(Element::getValue).orElse(null))
+            .firstOther(ofNullable(others.pollFirst()).map(Element::getValue).filter(not(Other::isEmpty)).orElse(null))
```
lewisbirks, line 29: thoughts on something like this?
```java
        Element<Other> first;
        do {
            first = others.pollFirst();
        } while (null == first && !others.isEmpty());

        return Others.builder()
            .firstOther(ofNullable(first).map(Element::getValue).orElse(null))
            .additionalOthers(others)
            .build();
```
it empties the list until it finds the first non null other or reaches the end of the list

lewisbirks (dismissed):

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/OthersControllerAboutToSubmitTest.java
```diff
@@ -48,6 +48,23 @@ void shouldOnlyRemoveConfidentialDetailsWhenOtherIsMarkedAsConfidential() {
         assertThat(unwrapElements(caseData.getConfidentialOthers())).containsOnly(confidentialOther());
     }
 
+    @Test
+    void shouldNotSaveFirstOtherWhenItIsEmpty() {
```
Rob-Doherty, line 52: maybe this test could be a little clearer, since it sounds like the save should fail rather than the first Other being overwritten by any following non-empty other Others

Rob-Doherty: is it worth including a 2nd `additionalOthers()` here to show that the values 'move up one', and it's not that the `getFirstOther()` and `getAdditionalOthers()` just flip values? so like
```
        assertThat(caseData.getOthers().getFirstOther()).isEqualTo(additionalOthers.get(0).getValue());
        assertThat(caseData.getOthers().getAdditionalOthers()).isEqualTo(additionalOthers.get(1).getValue());
```

aliveni: Updated the test to add another element and assert the size and elements of the collection

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OthersController.java
```diff
@@ -49,7 +50,12 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
 
         List<Element<Other>> others = confidentialService.removeConfidentialDetails(allOthers);
 
-        caseDetails.getData().put("others", Others.from(others));
+        Others othersList = Others.from(others);
+        if (isNull(othersList)) {
+            caseDetails.getData().remove("others");
```
aliveni, line 55: I had to remove "others" fields as some case data already have "others" value empty

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OthersController.java
```diff
@@ -49,7 +50,12 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
 
         List<Element<Other>> others = confidentialService.removeConfidentialDetails(allOthers);
 
-        caseDetails.getData().put("others", Others.from(others));
+        Others othersList = Others.from(others);
```
lewisbirks, line 53: not a list, var names are the wrong way round.

lewisbirks (dismissed):

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OthersController.java
aliveni: updated the variable names

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Others.java
```diff
@@ -22,16 +24,21 @@
...
         final LinkedList<Element<Other>> others = new LinkedList<>(Optional.ofNullable(allOthers).orElse(emptyList()));
-        return Others.builder()
-            .firstOther(ofNullable(others.pollFirst()).map(Element::getValue).orElse(null))
-            .additionalOthers(others)
-            .build();
+        others.removeIf(other -> isNull(other) || isNull(other.getValue()) || other.getValue().isEmpty());
+
+        if (isEmpty(others)) {
+            return null;
+        } else {
```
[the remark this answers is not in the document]
aliveni: removed else

lewisbirks (approved):

PR #3009, opened 2021-08-18
impr: See more messages in judicial message threads (FPLA-3001)

SooryaSu:
[FPLA-3001](https://tools.hmcts.net/jira/browse/FPLA-3001)

### Change description ###
Changes the judicial messages under the tab to be displayed within a table with expanding rows.
Each row can be expanded to show the full message history.

```
[ ] Yes
[x] No
```

### DIFF
--- a/ccd-definition/ComplexTypes/CareSupervision/JudicialMessage/JudicialMessage.json
+++ b/ccd-definition/ComplexTypes/CareSupervision/JudicialMessage/JudicialMessage.json
@@ -91,7 +91,8 @@
     "FieldType": "Collection",
     "FieldTypeParameter": "Document",
     "ElementLabel": "Related documents",
-    "SecurityClassification": "Public"
+    "SecurityClassification": "Public",
+    "FieldShowCondition": "relatedDocuments=\"*\""
   },
   {
     "LiveFrom": "01/01/2017",
@@ -133,8 +134,7 @@
     "ListElementCode": "messageHistory",
     "FieldType": "TextArea",
     "ElementLabel": "Message history",
-    "SecurityClassification": "Public",
-    "FieldShowCondition": "status=\"CLOSED\""
+    "SecurityClassification": "Public"
   },
   {
     "LiveFrom": "01/01/2017",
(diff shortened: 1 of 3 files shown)


ccd-definition/ComplexTypes/CareSupervision/JudicialMessage/JudicialMessage.json
```diff
@@ -133,15 +133,15 @@
...
   },
   {
     "LiveFrom": "01/01/2017",
     "ID": "JudicialMessage",
     "ListElementCode": "closeMessageLabel",
     "FieldType": "Label",
     "ElementLabel": "This message will now be marked as closed",
-    "SecurityClassification": "Public"
+    "SecurityClassification": "Public",
+    "FieldShowCondition": "closeMessageLabel=\"DO_NOT_SHOW\""
```
lewisbirks, line 145: Is this still displayed in the event? I am only guessing but I am assuming this is what it was originally for.

e2e/tests/messageJudge_test.js
```diff
@@ -4,6 +4,8 @@ const mandatoryWithAdditionalApplicationsBundle = require('../fixtures/caseData/
 let caseId;
 let message = 'Some note';
 let reply = 'This is a reply';
+const messageHistoryInitial = config.ctscEmail + ' - ' +  message;
+const messageHistoryReply = config.ctscEmail + ' - ' +  message + '\n \n' + config.judicaryUser.email + ' - ' + reply;
```
lewisbirks, line 8: I would define this as `messageHistoryInitial + ...` and also it would probs be nicer if you used template strings here (would have to make sure it is syntactically valid)
```suggestion
const messageHistoryInitial = `${config.ctscEmail} - ${message}`;
const messageHistoryReply = `${messageHistoryInitial}\n \n${config.judicaryUser.email} - ${reply}`;
```
Also is there a reason why there is a space between the 2 `\n`?

ccd-definition/ComplexTypes/CareSupervision/JudicialMessage/JudicialMessage.json
SooryaSu: I've checked and it still appears in the event as expected due it being in **judicialMessageReply.json**

e2e/tests/messageJudge_test.js
SooryaSu: The `buildMessageHistory` function in `MessageJudgeService.java` has been adding a space between the two `\n` so the test will fail without it

lewisbirks (approved):

PR #3018, opened 2021-08-19
More portable shebang

jasonpaige:
As ubuntu uses dash by default, the `function` keyword doesn't work and the terminal will complain about a "(" when running the generate-local-user-mappings.sh script. This shebang will work fine on linux and osx

```
[ ] Yes
[ ] No
```


lewisbirks (approved):
Might be worth checking all of the other scripts, as well as the ones in fpla-docker

PR #3054, opened 2021-08-31
fix: Migrate task list after migration of legacy applicant (FPLA-3238)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/FPLA-3238

### DIFF
--- a/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
+++ b/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.fpl.controllers.support;
 
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
@@ -7,10 +8,12 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 import org.springframework.boot.test.autoconfigure.OverrideAutoConfiguration;
 import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
 import uk.gov.hmcts.reform.ccd.model.Organisation;
 import uk.gov.hmcts.reform.ccd.model.OrganisationPolicy;
 import uk.gov.hmcts.reform.fpl.controllers.AbstractCallbackTest;
+import uk.gov.hmcts.reform.fpl.enums.Event;
 import uk.gov.hmcts.reform.fpl.enums.SolicitorRole;
 import uk.gov.hmcts.reform.fpl.enums.State;
 import uk.gov.hmcts.reform.fpl.model.Address;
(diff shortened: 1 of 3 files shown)


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
```diff
@@ -224,6 +245,55 @@ private OrganisationPolicy blankPolicyFor(SolicitorRole role) {
             .email("solicitor@legacy.com")
             .build();
 
+        final List<Task> tasks = List.of(Task.builder()
```
jvictorcf, line 248: What would you think of having these two fields be internal variables in the init() method (since they're only used there?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
```diff
@@ -224,6 +245,55 @@ private OrganisationPolicy blankPolicyFor(SolicitorRole role) {
...
+            final CaseData caseData = CaseData.builder()
+                .state(State.OPEN)
+                .applicants(wrapElements(Applicant.builder().party(legacyApplicant).build()))
+                .solicitor(legacySolicitor)
+                .localAuthorityPolicy(designatedOrg)
+                .build();
+
+            Map<String, Object> data = postAboutToSubmitEvent(buildCaseDetails(caseData, migrationId)).getData();
+
+            assertThat(data.get("taskList")).isEqualTo("<h1>Task list</h1>");
```
jvictorcf, line 278: isEqualTo(expectedTaskList)?

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
```diff
@@ -49,6 +61,15 @@
 
     private static final String INVALID_MIGRATION_ID = "invalid id";
 
+    @MockBean
+    private TaskListService taskListService;
```
jvictorcf, line 65: I don't know much about these elements, but it seems they only work with the case data itself (not external components).
Shouldn't we use the case data to assert that they this is working (as opposed to using a mock)?
Or is the case that these components are so well tested that just by knowing they're being called, we're already happy with this controller test?

jvictorcf (commented):
Just a few questions/comments.

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
tomaszpowroznik: done

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
tomaszpowroznik: done

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseControllerTest.java
tomaszpowroznik: yea, they are already tested, so no point to repeat it here. 
Code used in migration is mostly copy of 
https://github.com/hmcts/fpl-ccd-configuration/blob/master/service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseEventHandler.java

which is verified by unit and integration tests, e.g. here 

https://github.com/hmcts/fpl-ccd-configuration/blob/master/service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/TaskListControllerSubmittedTest.java

jvictorcf (dismissed):

PR #3090, opened 2021-09-20
ci: create shuttered definitions for each merge

lewisbirks:
### Change description ###

To shutter an individual CCD service you need to set the authorisation of a role to `D`.

To do this for FPLA we have introduced 2 versions of the AuthorisationCaseType file:
- `AuthorisationCaseType-nonshuttered.json` a renamed version of `AuthorisationCaseType.json`
- `AuthorisationCaseType-shuttered.json` contains a verison of the previous but with all permissions set to `D`

The generation scripts have been updated to ignore the shuttered version by default.
The pipeline steps have been updated to generate a shuttered version of the config with every merge to master.

**Does this PR introduce a breaking change?** (check one with \"x\")
```
[ ] Yes
[x] No
```


Jenkinsfile_CNP
```diff
@@ -69,6 +70,7 @@ def archiveCoreCaseDataDefinitions(env) {
       steps.archiveArtifacts artifacts: "ccd-release-config/ccd-fpl-preview-${CHANGE_ID}-toggle-off.xlsx"
     } else {
       steps.archiveArtifacts artifacts: "ccd-release-config/ccd-fpl-${env}.xlsx"
+      steps.archiveArtifacts artifacts: "ccd-release-config/ccd-fpl-${env}-shuttered.xlsx"
```
lewisbirks, line 73: If we wanted to dynamically be able to switch between shuttered an non shuttered then I think the only way to do that would be to `uploadCoreCaseDataDefinitions` method such that it can be controlled by some toggle from either a launch darkly (ideal) or key vault.

Once that is done we can update the else block to select the shuttered version of the defs if it the env is prod and the shutter toggle is enabled

lewisbirks: This would then just require that we set the toggle and then run the master pipeline for the defs to be updated

Jenkinsfile_CNP
```diff
@@ -80,16 +85,19 @@ def archiveCoreCaseDataDefinitions(env) {
...
     if("preview" == env){
-      ccdDefinitionFilePath="${WORKSPACE}/build/ccd-release-config/ccd-fpl-preview-${CHANGE_ID}-toggle-on.xlsx"
+      ccdDefinitionFilePath = "${WORKSPACE}/build/ccd-release-config/ccd-fpl-preview-${CHANGE_ID}-toggle-on.xlsx"
       sh """
         eval \$(./variables/load-preview-environment-variables.sh ${CHANGE_ID})
         ${WORKSPACE}/fpla-docker/bin/utils/ccd-import-definition.sh ${ccdDefinitionFilePath}
       """
     } else {
-      ccdDefinitionFilePath="${WORKSPACE}/build/ccd-release-config/ccd-fpl-${env}.xlsx"
+      ccdDefinitionFilePath = "prod" == env && useShutteredCaseDef == '1'
```
lewisbirks, line 97: do we need the prod condition here, if we remove it then we can easily test on other envs as well, as long as the secret pulled is per env

jasonpaige: you're right - gonna bin it off

SomeshRA (dismissed):
LGTM

jasonpaige (approved):

PR #3097, opened 2021-09-21
feat: Amend placement application (DFPL-115)

tomaszpowroznik:
https://tools.hmcts.net/jira/browse/DFPL-115


service/src/test/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsHelperTest.java
```diff
@@ -136,6 +141,100 @@ void shouldRemoveTempFieldsBasedOfGivenClass2() {
...
+
+            assertThat(caseDetailsMap).containsExactly(entry("key1", "some value 1"));
+        }
+
+        @Test
+        void shouldRemoveFieldsFromMultipleGroupsWhenNullInTargetObjects() {
+
+            final TestClass target = new TestClass(null, null, null, null);
+
+            final CaseDetailsMap actualCaseDetails = addFields(caseDetails, target, "Group1", "Group2");
```
jvictorcf, line 225: I wonder if "putFields" would be a better name for this method (since it replaces existing fields, like the "put" method in a map).

jvictorcf: Or even "copyFieldsMarkedWithGivenGroups".

service/src/main/java/uk/gov/hmcts/reform/fpl/utils/CaseDetailsHelper.java
```diff
@@ -64,6 +69,22 @@ public static CaseDetails removeTemporaryFields(CaseDetails caseDetails, Class c
         return caseDetails;
     }
 
+    public static CaseDetailsMap addFields(CaseDetailsMap caseDetails, Object target, String... groups) {
```
jvictorcf, line 72: I'd say the second parameter is not the target (as in it's not the object that will be modified). It's actually the "model" (the object to be copied - or partially copied)

jvictorcf (approved):
Just a few comments.

PR #3105, opened 2021-09-23
fix: ignore extension case on assigning render format (DFPL-264)

lewisbirks:

[DFPL-264](https://tools.hmcts.net/jira/browse/DFPL-264)

### Change description ###

When uploading orders, if the extension was all uppercase rather than the lower case (`DOCX` vs `docx`) then it would fail to assign a render format and as such the user wouldn't be able to complete the manage orders event.

To solve this we now ignore case when checking the extension in `RenderFormat#fromFileName`

Also update `DocumentHelper#hasExtension` method to perform the same check.

```
[ ] Yes
[x] No
```


SooryaSu (approved):
LGTM

PR #3125, opened 2021-09-29
DFPL-82 - Allow multiple court bundle documents and add ability to mark the bundle as confidential

KianKwa:

https://tools.hmcts.net/jira/browse/DFPL-82

### Change description ###

Allow multiple court bundle documents and add ability to mark the bundle as confidential

```
[ ] Yes
[x] No
```


ccd-definition/AuthorisationCaseField/CareSupervision/court-admin.json
```diff
@@ -2377,7 +2377,7 @@
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "courtBundleList",
     "UserRole": "caseworker-publiclaw-courtadmin",
-    "CRUD": "R"
+    "CRUD": "CRU"
```
lewisbirks, line 2380: why are we updating the admin permissions to be more than just read? the ticket doesn't mention them adding or updating the list

ccd-definition/AuthorisationCaseField/CareSupervision/cafcass.json
```diff
@@ -431,7 +431,7 @@
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "courtBundleList",
     "UserRole": "caseworker-publiclaw-cafcass",
-    "CRUD": "R"
+    "CRUD": "CRU"
```
lewisbirks, line 434: see previous

ccd-definition/AuthorisationCaseField/CareSupervision/cafcass-solicitor.json
```diff
@@ -468,6 +468,13 @@
     "UserRole": "[CAFCASSSOLICITOR]",
     "CRUD": "R"
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseFieldID": "courtBundleList",
+    "UserRole": "[CAFCASSSOLICITOR]",
+    "CRUD": "CRU"
```
lewisbirks, line 476: might just want to be R

ccd-definition/AuthorisationCaseField/CareSupervision/gatekeeper.json
```diff
@@ -774,7 +774,7 @@
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "courtBundleList",
     "UserRole": "caseworker-publiclaw-gatekeeper",
-    "CRUD": "R"
+    "CRUD": "CRU"
```
lewisbirks, line 777: see previous

ccd-definition/AuthorisationCaseField/CareSupervision/judiciary.json
```diff
@@ -781,7 +781,7 @@
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "CaseFieldID": "courtBundleList",
     "UserRole": "caseworker-publiclaw-judiciary",
-    "CRUD": "R"
+    "CRUD": "CRU"
```
lewisbirks, line 784: see previous

ccd-definition/AuthorisationCaseField/CareSupervision/court-admin.json
lewisbirks: if it it for the auth complex type then it might be worth just having `R` and `U` permissions and add a comment to it stating that the `U` is to enable functionality in the auth complex type. (same for the other comments related to this one)

e2e/pages/events/manageDocumentsLAEvent.page.js
```diff
@@ -135,6 +141,11 @@ module.exports = {
     I.click(this.fields.supportingDocuments(elementIndex).confidential);
   },
 
+  async selectConfidentialCourtBundle() {
+    const elementIndex = await this.getActiveElementIndex();
```
lewisbirks, line 145: I think that there is an `I.getActiveElementIndex()` method (see the actor.js file) in which case it could replace the current method in here.

e2e/tests/manageDocuments_test.js
```diff
@@ -329,9 +334,102 @@ Scenario('Solicitor with access uploads documents', async ({I, manageDocumentsEv
...
+  caseViewPage.selectTab(caseViewPage.tabs.courtBundle);
+  assertCourtBundleTabForSolicitor(I);
+
+  // Remove all documents from the court bundle
+  await I.navigateToCaseDetailsAs(config.swanseaLocalAuthorityUserOne, caseId);
+  await removeDocumentsFromCourtBundleEvent(I, caseViewPage, manageDocumentsLAEventPage);
 
+  // Assert no more documents on Court bundle tab
+  caseViewPage.selectTab(caseViewPage.tabs.courtBundle);
+  I.dontSeeTagInTab(['Court bundle 1', 'Documents 1']);
```
lewisbirks, line 367: I don't think this is required, although it would be very nice to have the e2e tests are the flakiest part of our pipeline so the less we are doing the better tbh

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -275,7 +275,7 @@ public DynamicList buildRespondentDynamicList() {
     @JsonProperty("documents_socialWorkEvidenceTemplate_document")
     @Valid
     public final Document socialWorkEvidenceTemplateDocument;
-    public final CourtBundle courtBundle;
+    public final List<Element<CourtBundle>> courtBundle;
```
lewisbirks, line 278: this looks rather concerning, have you thought about how this is going to interact with old cases? It seems like it would probably break

lewisbirks: unless it is a temporary field, in which case ignore me

lewisbirks (commented):
Looks good but I am concerned about backward compatibility with existing cases

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CourtBundle.java
```diff
@@ -1,34 +1,33 @@
...
 import lombok.EqualsAndHashCode;
+import lombok.extern.jackson.Jacksonized;
 import uk.gov.hmcts.reform.fpl.model.common.DocumentMetaData;
 import uk.gov.hmcts.reform.fpl.model.common.DocumentReference;
 
-import java.time.LocalDateTime;
+import java.util.List;
 
 @Data
+@Builder(toBuilder = true)
```
lewisbirks, line 15: Does this not need to be a super builder?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
lewisbirks: it doesn't look like this field is used, could it just be removed?

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -680,14 +680,18 @@ public String getComplianceDeadline() {
     private final Object courtBundleHearingList;
     private final Object respondentStatementList;
 
-    private final CourtBundle manageDocumentsCourtBundle;
-    private final List<Element<CourtBundle>> courtBundleList;
+    private final List<Element<CourtBundle>> manageDocumentsCourtBundle;
```
lewisbirks, line 683: ahh this is the temp field

ccd-definition/CaseField/CareSupervision/manageDocumentsLA.json
```diff
@@ -26,12 +26,22 @@
...
+    "SecurityClassification": "Public",
+    "Searchable": "N"
+  },
   {
     "LiveFrom": "01/01/2017",
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "ID": "manageDocumentsCourtBundle",
-    "FieldType": "CourtBundleV2",
-    "Label": " ",
+    "FieldType": "Collection",
```
lewisbirks, line 42: this looks like the temp field so should be okay

ccd-definition/CaseField/CareSupervision/caseDocuments.json
```diff
@@ -128,7 +128,7 @@
     "CaseTypeID": "CARE_SUPERVISION_EPO",
     "ID": "courtBundleList",
     "FieldType": "Collection",
-    "FieldTypeParameter": "CourtBundleV2",
+    "FieldTypeParameter": "HearingCourtBundle",
```
lewisbirks, line 131: so this will break existing cases, in a way that will be very annoying to fix, the best solution that I can advise is creating a new separate field for it. and then have everything reference this new field, the migration can then transform everything from the old field to the new field and then

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CourtBundle.java
lewisbirks: which is what the constructor was doing below by the looks of things

ccd-definition/CaseField/CareSupervision/caseDocuments.json
lewisbirks: at the end of the migration after the transformation you can then remove the old field

ccd-definition/AuthorisationCaseField/CareSupervision/court-admin.json
KianKwa: I added a new generated complex type field called 'courtBundleNC'. This field is supposed to be shown to the non-LA and non-HMCTS users only and doesn't contain the confidential documents. The existing 'courtBundle' complex type field will only be shown to LA and HMCTS users. In order to restrict these fields, I will need to mark some roles as 'U' only. If I leave 'courtBundleList' as 'R', it complains about the parent type has lower access level than the child type. 

In order to restrict the access to 'courtBundle' and 'courtBundleNC' properly, the permission on the higher level need to up and hence I change them to 'CRU'. Note that this is done the same way for other document sub-event like 'Respondent statement' etc. 

I can try changing it to 'R' and 'U' only and check it still work.

KianKwa: All the comments regarding the 'CRU' permission has been update to 'CU'

e2e/tests/manageDocuments_test.js
KianKwa: OK I removed that.

PR #3207, opened 2021-11-22
Nightly dev

jasonpaige:

```
[ ] Yes
[X] No
```


Jenkinsfile_nightly
```diff
@@ -84,7 +84,6 @@ withNightlyPipeline(type, product, component) {
   enableFullFunctionalTest(240)
   // 'microsoft' and 'safari' to be returned to crossbrowser tests once FPLA-3172 improvements are complete
   enableCrossBrowserTest(['chrome', 'firefox'])
-  enableFortifyScan()
```
DanCatchpole, line 87: do we want to get rid of the fortify scan on nightly? I don't think we run it anywhere else as far as I can see?

SomeshRA: Wasn't Fortify for static code analysing which is required.

jasonpaige: true, i think we need to keep this. I'll add it back in

DanCatchpole (approved):

PR #3282, opened 2022-01-12
DFPL-331: Remove serial number on placement order creation

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-331

### Change description ###
Hide the Serial Number field from the manage orders > create order > placement order (A70) page
_This field is not used in the generation of the order so should not have any side effects for its removal_
Also stops the E2E tests from filling in this field as it is no longer present in the ui

```
[ ] Yes
[X] No
```


jamesrferguson1 (approved):
Looks good, removes the serial number as requested in JIRA.

PR #3300, opened 2022-01-20
DFPL-524: Update test fixtures

DanCatchpole:
N/A (one issue found during DFPL-331 testing, one during DFPL-104)

### Change description ###
- Adds the Local Authority information to all of the test fixtures (this is added during CaseInitiation for 'real' created cases - around line 183 in CaseInitiationService) - **This was missing in all bar about 3**
- Also supersedes old PR which found that a SDO was missing in the prepareForHearing case fixture (this should never be missing - is mandatory to get to that stage in 'real cases')

```
[ ] Yes
[X] No
```


randersonMOJ (approved):
Approved

PR #3307, opened 2022-01-25
fix: Send notifications to LA shared inbox (DFPL-377)

alivenichoppa:
https://tools.hmcts.net/jira/browse/DFPL-377

### Change description ###
Notifications should be sent to the LA shared inbox in addition to the LA Group Email. 
LA group email can be different from the LA shared inbox email address as it is an editable text field.

```
[ ] Yes
[x] No
```


qasimnawaz-moj (dismissed):
Looks good to me!

PR #3402, opened 2022-03-09
DFPL-498 Notifications to be sent to LA/Respondents upon submission of interlocutory applications

so99y:
https://tools.hmcts.net/jira/browse/DFPL-498

### Change description ###

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/AdditionalApplicationsUploadedEventHandler.java
```diff
@@ -195,6 +182,24 @@ public void notifyApplicant(final AdditionalApplicationsUploadedEvent event) {
         }
     }
 
+    @EventListener
+    @Async
+    public void notifyAllLocalAuthorities(final AdditionalApplicationsUploadedEvent event) {
+        final CaseData caseData = event.getCaseData();
+        final OrderApplicant applicant = event.getApplicant();
```
jamesrferguson1, line 189: Don't think this is used?

service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/AdditionalApplicationsUploadedEventHandler.java
```diff
@@ -164,23 +165,9 @@ public void notifyApplicant(final AdditionalApplicationsUploadedEvent event) {
...
-        } else if (applicant.getType() == SECONDARY_LOCAL_AUTHORITY) {
-
-            final RecipientsRequest recipientsRequest = RecipientsRequest.builder()
-                .caseData(caseData)
-                .designatedLocalAuthorityExcluded(true)
-                .build();
-
-            recipients.addAll(localAuthorityRecipients.getRecipients(recipientsRequest));
+        if (applicant.getType() == LOCAL_AUTHORITY || applicant.getType() == SECONDARY_LOCAL_AUTHORITY) {
+            // do nothing
```
jamesrferguson1, line 169: Guess you could get rid of the if/else and do something like if(applicatn.getType() != LOCAL_AUTHORITY && applicant.getType() != SECONDARY_LOCAL_AUTHORITY)?

Or rather than two event listeners, you could perhaps convert notifyAllLocalAuthorities to a helper method and always call it in this method so LAs are always notified?

jamesrferguson1 (approved):

PR #3532, opened 2022-05-10
DFPL-644: Add the issuing date rather than the original order date

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-644

### Change description ###
Added the correct date to the recovery order so it doesn't show the care/EPO order creation date as the C29 issuing date.

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
LGTM

PR #3557, opened 2022-05-26
DFPL-338: Add new courts and remove existing ones from the dropdown

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-338

### Change description ###
- Add 4 new courts for the expected new venues 
  - _keep the old ones in the address config file for existing hearings to not get mixed up_
- Label the 2 existing courts in the CCD drop down to say (CLOSED) so they exist but are unlikely to be added to new hearings 

```
[ ] Yes
[X] No
```


jamesrferguson1 (dismissed):
Adding new courts and deprecating and marking as closed existing renamed courts

PR #3574, opened 2022-06-09
DFPL-666 Migrate the case to remove the empty hearing

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-666

### Change description ###
User gets an exception when trying to start the manageDocumentsLA event
The case have a hearing created where every sub-field is null. 
We will have to migrate the case to remove the empty hearing causing the issues

```
[ ] Yes
[X] No
```


qasimnawaz-moj (dismissed):
lgtm

qasimnawaz-moj (dismissed):
lgtm

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -43,13 +44,13 @@
 public class MigrateCaseController extends CallbackController {
     private static final String MIGRATION_ID_KEY = "migrationId";
     private final Map<String, Consumer<CaseDetails>> migrations = Map.of(
-        "DFPL-500", this::run500,
```
chak-shing-lo-justice, line 46: Map.of() only support 10 entries at most. So I removed the old and done script from the map

jamesrferguson1 (approved):

PR #3612, opened 2022-06-28
DFPL-715: Add CDAM to preview deployments

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-715

### Change description ###
spinning up a cdam container in preview

```
[ ] Yes
[X] No
```

### DIFF
--- a/charts/fpl-case-service/values.preview.template.yaml
+++ b/charts/fpl-case-service/values.preview.template.yaml
@@ -92,7 +92,7 @@ xui-webapp:
       HEALTH_TERMS_AND_CONDITIONS_API: http://xui-terms-and-conditions-${SERVICE_FQDN}.service.core-compute-preview.internal/health
       SERVICES_CCD_COMPONENT_API: http://gateway-${SERVICE_FQDN}
       SERVICES_CCD_DATA_STORE_API: http://${SERVICE_NAME}-ccd-data-store-api
-      SERVICES_DOCUMENTS_API_V2: http://ccd-case-document-am-api-aat.service.core-compute-aat.internal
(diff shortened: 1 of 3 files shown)


charts/fpl-case-service/values.preview.template.yaml
```diff
@@ -294,3 +294,19 @@ aac-manage-case-assignment:
...
+    releaseNameOverride: ${SERVICE_NAME}-cdam
+    imagePullPolicy: Always
+    image: hmctspublic.azurecr.io/ccd/case-document-am-api:latest
+    ingressHost: ccd-case-document-am-api-${SERVICE_FQDN}
+    environment:
+      IDAM_API_URL: https://idam-api.aat.platform.hmcts.net
+      IDAM_OIDC_URL: https://idam-web-public.aat.platform.hmcts.net
+      OIDC_ISSUER: https://forgerock-am.service.core-compute-idam-aat.internal:8443/openam/oauth2/hmcts
+      S2S_URL: http://rpe-service-auth-provider-aat.service.core-compute-aat.internal
+      DM_STORE_BASE_URL: http://dm-store-aat.service.core-compute-aat.internal
```
jamesrferguson1, line 310: One question (most probably a stupid one :) ) ... how do we know what the value for ${SERVICE_NAME} is? From above I guess that it's "aat.service.core-compute-aat.internal"?

If so I guess we could reuse it here?

DanCatchpole: ah think it's `fpl-case-service-pr-${PR_NUM}`, so it keeps the pod names separate per PR

PR #3693, opened 2022-08-08
DFPL-779 Reduce number of case data fields

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-779

### Change description ###
We are starting to see "to many fields on CaseData" errors when we are trying to add new fields to this class. We should look at a work around for this - probably a combination of:
1. Removing deprecated fields
2. Removing migrated fields
3. Moving fields out into another class

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -146,16 +148,18 @@
 import static uk.gov.hmcts.reform.fpl.utils.ElementUtils.unwrapElements;
 
 @Data
-@Builder(toBuilder = true)
-@AllArgsConstructor
+@SuperBuilder(toBuilder = true)
+@Jacksonized
+@EqualsAndHashCode(callSuper = true)
+//@AllArgsConstructor
```
qasimnawaz-moj, line 154: We should remove this commented out line

qasimnawaz-moj (approved):

PR #3707, opened 2022-08-11
DFPL-652: Add application form removal to super-user tool

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-652

### Change description ###
**User flow changes**
 - Adds a new option to the super-user removal tool: "Remove Application Form"
 - The user can then enter a reason why the document is being removed.
 - Upon completing the event the Application Form field will be removed from the case + the original application form will be moved to a 'Removed Application Form' field on the Documents tab which is **only visible to superusers**.

**Technical changes**
 - Refactors the `CaseData` fields related to the removal tool into a separate containing class - `RemovalToolData`

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RemovalToolController.java
```diff
@@ -94,36 +109,30 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
...
+        } else {
+            if (isEmpty(caseData.getC110A().getDocument())) {
+                return respond(caseDetailsMap, List.of(APPLICATION_FORM_ALREADY_REMOVED_ERROR_MESSAGE));
+            }
+            caseDetailsMap.put("submittedForm", null);
+            caseDetailsMap.put("supplementDocument", null);
+            caseDetailsMap.put("hiddenApplicationForm", RemovedApplicationForm.builder()
+                    .submittedForm(caseData.getC110A().getDocument())
+                    .submittedSupplement(!isEmpty(caseData.getC110A().getSupplementDocument())
+                        ? caseData.getC110A().getSupplementDocument() : null)
```
jamesrferguson1, line 131: don't know if we could maybe get away with having this just as 
```suggestion
                        .submittedSupplement(caseData.getC110A().getSupplementDocument())
```?

Comment (DanCatchpole) [reply service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RemovalToolController.java:131]: ah yeah true! if it's not there it'll be null anyway!

File: service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/RemovalToolController.java [hunk 2]
```diff
@@ -134,14 +143,19 @@ public void handleSubmittedEvent(@RequestBody CallbackRequest callbackRequest) {
...
+            caseData.getRemovalToolData().getHiddenCMOs(), caseDataBefore.getRemovalToolData().getHiddenCMOs()
         );
         Optional<AdditionalApplicationsBundle> removedApplication = applicationService.getRemovedApplications(caseData
-                .getHiddenApplicationsBundle(),
-            caseDataBefore.getHiddenApplicationsBundle());
+            .getRemovalToolData().getHiddenApplicationsBundle(),
+            caseDataBefore.getRemovalToolData().getHiddenApplicationsBundle());
+
+        Optional<DocumentReference> removedApplicationForm =
+            (caseData.getC110A().getSubmittedForm() != caseDataBefore.getC110A().getSubmittedForm())
```
jamesrferguson1, line 156: Just want to check, will we always have a C110a application at this stage on the case data?

DanCatchpole: .getC110A() will always be a thing (as it's a JsonUnwrapped container class), and as it's the submitted event the checks in about-to-submit would make sure that there was a submittedForm beforehand, so generally this condition will always be true, but is there just incase nothing has happened (either no removal or no document being there in the place, as in both of these they'll be equal!)

PR #3833, opened 2022-10-10
update gradle script to support local docmosis

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-819

### Change description ###
Update the gradle script to support using local docmosis

Run the command : 
_DOCMOSIS_KEY=<docmosis key> ./gradlew bootWithCCD -Pdocmosis_
Then the script will create the local docmosis docker container and point FPL boot to use the local docmosis

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -492,6 +492,22 @@ bootWithCCD {
     environment 'SEND_LETTER_URL', 'http://dm-store-aat.service.core-compute-aat.internal'
     environment 'PRD_HOST', 'http://rd-professional-api-aat.service.core-compute-aat.internal'
     environment 'MANAGE_CASE_S2S_AUTHORISED_SERVICES', 'ccd_gw,xui_webapp,ccd_data,fpl_case_service'
+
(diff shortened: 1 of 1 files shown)


service/build.gradle
```diff
@@ -492,6 +492,22 @@ bootWithCCD {
...
+      exec {
+        ignoreExitValue = true
+        commandLine 'docker', 'compose', '-f', '../fpla-docker/compose/docmosis.yml', 'create'
+      }
+      exec {
+        ignoreExitValue = false
+        commandLine 'docker', 'compose', '-f', '../fpla-docker/compose/docmosis.yml', 'start'
+      }
+      environment 'docmosis.tornado.url', 'http://localhost:5433'
+      environment 'docmosis.tornado.key', 'ACCESS_KEY'
```
DanCatchpole, line 507: should we just include this in .aat-env? (assuming it's not literally ACCESS_KEY 😅 ) then it would get autoloaded in when it downloads that file from the vault

PR #3837, opened 2022-10-11
DFPL-929: Do not clear old hearing docs when saving newly uploaded docs

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-929

### Change description ###
 - Stops clearing the old hearing documents from the 3 collections (case summary, respondent position statements, child position statements) during about-to-submit
 - Still allows the user to upload a new version (if same hearing (and same respondent/child) then it will just replace that hearing's version)

```
[ ] Yes
[X] No
```


so99y (approved):
LGTM

PR #3855, opened 2022-10-24
DFPL-949: Get high court email direct from config

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-949

### Change description ###
 - Adds the specific case to retrieve the high court email if that is the court on the case, as it needs to bypass the CTSC.
   - Needed as we don't encode the high court email into the case data when it's transferred which the existing code expects if it doesn't go to the CTSC.

```
[ ] Yes
[X] No
```


AlistairEdwinOsborne (approved):
LGTM

PR #3871, opened 2022-10-31
Dfpl 990 Secure Accommodation Order - show mandatory field in "why can't i submit"

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-990

### Change description ###
Need to add checks on the secureAccommodationOrderSection field (Under which section are you applying?) to the OrdersSoughtChecker if the user is applying for a Secure Accommodation Order, as it is used within the C20 supplement docmosis generation, so they shouldn't be allowed to complete the app without that question answered.

Error should be visible in the "Why can't I submit" section similar to the others.

```
[ ] Yes
[X] No
```


so99y:
missing test case?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/OrdersSoughtChecker.java
```diff
@@ -59,6 +67,10 @@ public boolean isCompleted(CaseData caseData) {
             }
         }
 
+        if (orders.isSecureAccommodationOrder() && isSecureAccomodationOrderNotCompleted(orders)) {
```
DanCatchpole, line 70: Bit of a weird one, as I can't add a comment to the right line 😆

was just checking the orders.isSecureAccomodationOrder() function, I think it needs a null/empty check in it (similar to the containsEPO/isDischargeOfCare) otherwise the validator might break before we fill in that section

chak-shing-lo-justice: Oh right! Added null checking

DanCatchpole (dismissed):
lgtm!

PR #3936, opened 2022-11-28
DFPL-809-V2: Check for empty confidential documents list

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-809

Add check to prevent empty confidential documents list being added to case data causing the confidential documents tab to show with no data in it.

Also add migration to remove empty list from live case

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
lgtm

PR #3978, opened 2022-12-09
DFPL-1062 'Free Textbox' missing from 'Special Guardianship' Order

so99y:
https://tools.hmcts.net/jira/browse/DFPL-1062

### Change description ###
Free text box missing on special guardianship order and so unable to create an order for certain cases.
Where more than one party is selected it should say 'and' instead of a ',' between the names.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/orders/generator/C43aSpecialGuardianshipOrderDocumentParameterGenerator.java
```diff
@@ -81,7 +82,20 @@ private String getSpecialGuardianAppointeeMessage(CaseData caseData, int numOfCh
...
 
-        return format("The Court orders that %s appointed as special guardian for the %s.", applicant, childOrChildren);
+        ManageOrdersEventData eventData = caseData.getManageOrdersEventData();
+
+        StringBuilder sb = new StringBuilder();
+        if (isNotEmpty(applicant)) {
+            sb.append(format("The Court orders that %s appointed as special guardian for the %s.", applicant,
+                    childOrChildren));
+        }
+        if (isNotEmpty(eventData.getAppointedGuardianDetails())) {
```
jamesrferguson1, line 92: why do we need the getAppointedGuardianDetails() here too (trying to see from the ticket)? just wondering what it contains?

jamesrferguson1, line 92: Just to double check, would we expect there to be a situation where the user uses both the free text field and selects the guardians from the drop box?

Is it needed to add additional guardians to the ones list above?

jamesrferguson1 (approved):

PR #4014, opened 2023-01-06
DFPL-1110 Add legal counsel cannot remove users if no longer in manag…

so99y:
https://tools.hmcts.net/jira/browse/DFPL-1110

### Change description ###
Not to throw an exception if user cannot be found. Put a log to info logs instead.. 
Therefore, allowing user to remove users which are no longer in manage org.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/LegalRepresentativeService.java
```diff
@@ -46,23 +46,20 @@ public void updateRepresentatives(Long caseId,
                     userId -> caseService.addUser(Long.toString(caseId),
                         userId,
                         userToBeAdded.getRole().getCaseRoles()),
-                    throwException(userToBeAdded)
+                    logUnableToFindUserFromOrganisationService(userToBeAdded)
```
DanCatchpole, line 49: wonder if we should keep the exception for the adding part, so we don't add people that don't exist, but then log when we remove?

so99y: oh yes, you're right.

DanCatchpole (dismissed):

DanCatchpole (approved):
adding the doc size logs to this one should be alright! 👍

PR #4188, opened 2023-03-02
DFPL-1270 Remove Case Summary Document From MB22C50275

so99y:
https://tools.hmcts.net/jira/browse/DFPL-1270

### Change description ###
Remove Case Summary Document From MB22C50275

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -135,4 +136,11 @@ private void run1210(CaseDetails caseDetails) {
             "EDIT_HEARING", migrationId);
         caseDetailsData.put("hearingOption", HearingOptions.EDIT_PAST_HEARING);
     }
+
+    private void run1270(CaseDetails caseDetails) {
+        String migrationId = "DFPL-1270";
+        migrateCaseService.doCaseIdCheck(caseDetails.getId(), 1656581935993992L, migrationId);
+        caseDetails.getData().putAll(migrateCaseService.removeCaseSummaryByHearingId(getCaseData(caseDetails),
+            migrationId, UUID.fromString("9e6bc008-3577-4c6b-8a29-521c9027b17f")));
```
DanCatchpole, line 144: I imagine it'll be fine but perhaps worth confirming that the UUID on the ticket is the right one - just had a worry with the function name being `removeCaseSummaryByHearingId`

so99y: Looks like the case summary using hearing id as an identifier. I will double check it.

so99y: I have checked that it seems the relation between hearing and case summary is 1 to 1. That means there is only a single case summary for each hearing.

so99y: and verified that the id in case summary is actually using the corresponding hearing id.

PR #4214, opened 2023-03-08
DFPL-1296 make allocation proposal mandatory for all case type

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1295

### Change description ###
make allocation proposal mandatory for all case type

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/CaseSubmissionChecker.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/CaseSubmissionChecker.java
@@ -59,7 +59,7 @@ public class CaseSubmissionChecker extends CompoundEventChecker {
 
         events.add(CASE_NAME);
         events.add(ORDERS_SOUGHT);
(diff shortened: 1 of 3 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/validators/CaseSubmissionChecker.java
```diff
@@ -59,7 +59,7 @@ private List<Event> getRequiredEvents(CaseData caseData) {
 
         events.add(CASE_NAME);
         events.add(ORDERS_SOUGHT);
-        if (!caseData.isC1Application() 
+        if (!caseData.isC1Application()
```
so99y, line 62: revert this change? Seems no difference.

chak-shing-lo-justice: done

so99y (approved):
Looks good.

PR #4288, opened 2023-03-29
DFPL-1352: Add migration to set sendToCtsc to true

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-1352

### Change description ###
 - IF the case is not in the high court AND is not sending to the CTSC by default,
 - We should migrate the case to make sure it is being sent to the CTSC, and so that notifications aren't being lost

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -73,4 +78,25 @@ private void run1226(CaseDetails caseDetails) {
         migrateCaseService.doDocumentViewNCCheck(caseDetails.getId(), "DFPL-1226", caseDetails);
         caseDetails.getData().putAll(migrateCaseService.refreshDocumentViews(getCaseData(caseDetails)));
     }
+
+    private void run1352(CaseDetails caseDetails) {
+        var migrationId = "DFPL-1352";
+
+        CaseData caseData = getCaseData(caseDetails);
+
+        if (caseData.getCourt().getCode().equals(RCJ_HIGH_COURT_CODE)) {
```
so99y, line 87: Assuming court and its code should not be null, right?

so99y (approved):
lgtm

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
DanCatchpole: yeah, should hopefully be alright as going to do the migration selectively (think anything missing a court + code can be taken off the list for migration, as likely very very old cases) just trying to figure out the set of cases now 😄

PR #4320, opened 2023-04-11
Fix sorting based on string to based on date

DanCatchpole:
N/A - Fix as one of the unit tests was failing

### Change description ###
 - Replaces the sorting from the string version of date to the LocalDateTime variable

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
lgtm

PR #4451, opened 2023-06-06
DFPL-1477 Cafcass receiving multiple notifications

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1477

### Change description ###
Cafcass receiving multiple notifications

Checked the case id 1683040396030080 (NE23C50251), looks like the submitted event is triggered 3 times.
The response time of each request is around 1 minute, while the CCD retry the callback every 30 seconds.

Change the spring listener to async, avoiding blocking the callback.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/FurtherEvidenceUploadedEventHandler.java
```diff
@@ -178,6 +183,7 @@ public void sendCourtBundlesUploadedNotification(final FurtherEvidenceUploadedEv
         }
     }
 
+    @Async
```
jamesrferguson1, line 186: Just wondering - do we want these to be @Retryable too?

chak-shing-lo-justice: The service will not throw exception if it fails to send notification. It will only log an error message.
So, if the handler catches any exception, it should be some data issue / program bugs (e.g. null pointer) which I think is a bit pointless to retry. 
But sure I can add @Retryable if we want to play safe.

jamesrferguson1: yeah makes sense - ok as it is i think

jamesrferguson1 (approved):

PR #4458, opened 2023-06-08
DFPL-1475: Duplicate notifications being sent to CTSC

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-1475

Setup handle submitted callback to use new concurrency standard for postSubmitCallback
AfterSubmissionCaseDataUpdatedEventHandler to be async
Shift handle post submitted publish event to submitted

```
[ ] Yes
[X] No
```


jamesrferguson1 (dismissed):

service/build.gradle
```diff
@@ -54,51 +54,58 @@ sourceSets {
...
 }
 
+/*tasks.withType(Test) {
+  useJUnitPlatform()
+
+  testLogging {
+    exceptionFormat = 'full'
+  }
+}*/
+
```
DanCatchpole, line 105: Any reason why we're keeping this commented out? Might be useful in future if we do get an exception, saves uncommenting + rerunning the build

service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/RespondentControllerSubmittedTest.java
```diff
@@ -364,34 +387,39 @@ void shouldUpdateRepresentativesAccess() {
...
+
         // setup change field, submit to aac (org1), setup change field, submit to aac (org2), update case-summary
-        verify(concurrencyHelper, times(5)).submitEvent(any(),
-            eq(caseData.getId()), caseCaptor.capture()
-        );
+        checkThat(() -> {
+            verify(concurrencyHelper, times(5)).submitEvent(startEventCaptor.capture(),
+                eq(caseData.getId()), caseCaptor.capture()
+            );
+        }, Duration.ofSeconds(10));
```
DanCatchpole, line 401: Was there any issues with the checkThat in combination with the times(5) thing, I think we chained them before right, `times(5).timeout(10000)` ?

AlistairEdwinOsborne: yeah for this instance the standard times(5) one failed so swapped it back to the checkThat()

service/build.gradle
AlistairEdwinOsborne: When this is in it fails the build so thought it would be good to leave it in so anyone who wants to can uncomment it

DanCatchpole (approved):

PR #4490, opened 2023-06-21
DFPL-1542: Update notifications when solicitors are added to the case

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-1542

### Change description ###
 - Uses the 1st applicant's name as entered during initial application by default, NOT the name of the relating LA on the case when creating the "Unrepresented Respondent" notification (only as a fallback).
 - Update the notification to not include "C110a" as it may be a C1 application with supplement.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -1298,4 +1298,13 @@ public boolean isEPOCombinedWithICO() {
     public String getCaseLaOrRelatingLa() {
         return isEmpty(caseLocalAuthority) ? relatingLA : caseLocalAuthority;
     }
+
+    @JsonIgnore
+    public Optional<String> getApplicantName() {
+        if (!isEmpty(getLocalAuthorities())) {
+            return Optional.of(getLocalAuthorities().get(0).getValue().getName());
```
jamesrferguson1, line 1305: probably ok, but do we want to maybe use getCaseLocalAuthority()?

jamesrferguson1 (approved):

PR #4543, opened 2023-07-18
DFPL-1627 Cafcass not receiving notice of hearing

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1627

### Change description ###
Cafcass not receiving notice of hearing when a gatekeeping hearing is listed.
Fix "List gatekeeping hearing" event to force FPL to build a notice of hearing for the hearing booking.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/ManageHearingsService.java
```diff
@@ -319,18 +319,22 @@ public HearingBooking getCurrentHearingBooking(CaseData caseData) {
 
     public void sendNoticeOfHearing(CaseData caseData, HearingBooking hearingBooking) {
         if (YES.getValue().equals(caseData.getSendNoticeOfHearing())) {
```
so99y, line 321: Imagine `caseData.getSendNoticeOfHearing` is null in `listGatekeepingHearing` event since the field `sendNoticeOfHearing` was not shown. If so, why don't we just add `|| caseData.getSendNoticeHearing() == null` instead?

chak-shing-lo-justice: In "List gatekeeping hearing" event, notice of hearing is always sent automatically. It is meaningless to check the flag for gatekeeping hearing.
Also, I don't wanna bring any potential change to ManageHearing event.

so99y: Maybe add a boolean parameter (`boolean force`) overloading the same method? 
Looks like people may not know `buildNoticeOfHearing` will send a notification.

chak-shing-lo-justice: Both sendNoticeOfHearing() and buildNoticeOfHearing() won't send a notification. They just build the "notice of hearing" document and attach it to the hearing.
The notification will be sent in submitted stage

(Don't know who name this method as "send", lol )

chak-shing-lo-justice: Instead ,maybe I should rename the sendNoticeOfHearing() to buildNoticeOfHearingIfYes() 
lol

DanCatchpole: I may be misunderstanding the method - but I thought the tick box was for sending the notice of hearing, I thought it was always generated just not sent out, would this change not mean that it now won't get generated if they select "no"?

chak-shing-lo-justice: ### The existing behaviour:
It won't generate the notice of hearing if user selected YES. The notification will be sent out in submitted stage if there is a notice of hearing generated. 
In "listing gatekeeping hearing", there is no such option for user to select, it is always set to "YES"

### New behaviour:
I didn't change the behaviour of "ManageHearing", I just simply extract that generating part to a separated method.
Fo "gatekeeping hearing", I skip the  YES/NO checking, and always generate it.

DanCatchpole (approved):

PR #4545, opened 2023-07-20
DFPL-1476: Default behaviour when auto-share fails

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-1476

### Change description ###
 - If the grant access to a set of users in the organisation fails, then attempt to share it just to the case creator - they should always be able to have access (bar the usual timeouts, etc.), so that if any issues are in the organisation, the user can create urgent cases.

```
[ ] Yes
[X] No
```


so99y (dismissed):
Looks good to me

so99y (approved):

PR #4566, opened 2023-08-02
Remove e2e tests from PR pipeline for FPL

amoriartyMOJ:
### Change description ###
The end to end tests are overkill. They only assert UI behaviour which is owned by XUI. Running it against every PR is too much. These should be running in the nightly pipeline only.

N/A

```
[ ] Yes
[X] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -9,8 +9,6 @@ def type = "java"
 def product = "fpl"
 def component = "case-service"
 
-// def yarnBuilder = new uk.gov.hmcts.contino.YarnBuilder(this)
-
 AppPipelineConfig pipelineConf;
 
 def serviceSecrets = [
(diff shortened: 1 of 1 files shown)


Jenkinsfile_CNP
```diff
@@ -268,20 +268,20 @@ withPipeline(type, product, component) {
...
-
-  afterSuccess('functionalTest:preview') {
-    env.PARALLEL_CHUNKS = 12
-    yarnBuilder.yarn('test:functional')
-    steps.archiveArtifacts allowEmptyArchive: true, artifacts: 'output/**/*'
-    steps.archiveArtifacts allowEmptyArchive: true, artifacts: 'build/reports/tests/**/*'
-    steps.archiveArtifacts allowEmptyArchive: true, artifacts: 'build/reports/serenity/**/*'
-    steps.archiveArtifacts allowEmptyArchive: true, artifacts: 'test-results/functional/**/*'
-  }
+//  before('functionalTest:preview') {
```
qasimnawaz-moj, line 271: Should maybe add a short comment explaining why we're commenting these tests out

amoriartyMOJ: Hey Qasim, my bad for raising PR early. Was commenting to test. As it's worked i've removed them entirely. Thanks for the comment though!

Will mention this in standup tomorrow and get everyones thoughts before continuing with it

DanCatchpole (approved):

PR #4644, opened 2023-08-22
DFPL-1681-3: Add NC documents to removal

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-1681

Document is also present in NC document collection

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
@@ -710,6 +710,24 @@ public class MigrateCaseService {
         return Map.of("correspondenceDocuments", newCorrespondenceDocuments);
     }
 
+    public Map<String, Object> removeCorrespondenceDocumentNC(CaseData caseData,
+                                                            String migrationId,
+                                                            UUID expectedDocumentId) {
+
+        List<Element<SupportingEvidenceBundle>> newCorrespondenceDocuments =
+            caseData.getCorrespondenceDocuments().stream()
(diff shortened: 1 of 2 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
```diff
@@ -701,13 +702,19 @@ public Map<String, Object> removeCorrespondenceDocument(CaseData caseData,
                 .filter(el -> !expectedDocumentId.equals(el.getId()))
                 .collect(toList());
 
+        List<Element<SupportingEvidenceBundle>> newCorrespondenceDocumentsNC =
+            newCorrespondenceDocuments.stream()
+                .filter(el -> isNull(el.getValue().isConfidentialDocument()) || !el.getValue().isConfidentialDocument())
```
FaisalMoJ, line 707: `isNull(el.getValue().isConfidentialDocument()` will always return false because it will either return true or false and will never be null

service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
```diff
@@ -701,13 +702,19 @@ public Map<String, Object> removeCorrespondenceDocument(CaseData caseData,
...
+                .collect(toList());
```
FaisalMoJ, line 708: Based on personal experience with a ticket few days ago, Sonar will raise this as code smell and fix for that is to use `.toList()` instead of `.collect(toList())` 🙂

FaisalMoJ (approved):

PR #4673, opened 2023-09-07
DFPL-1600: Remove placement documents

FaisalMoJ:

https://tools.hmcts.net/jira/browse/DFPL-1600

### Change description ###

Currently placement non confidential documents are available for respondents and respondent soliticors to view before notice of placement hearing is produced. Requirement is that these documents should only be visible to respondents and solicitors once notice of placement hearing is produced.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/event/PlacementEventData.java
```diff
@@ -73,9 +72,12 @@ public List<Element<Placement>> getPlacementsNonConfidential(boolean withNotices
...
             return emptyList();
         }
+
         return placements.stream()
-            .map(element -> element(element.getId(), element.getValue().nonConfidential(withNoticesResponses)))
-            .collect(toList());
+                .filter(element -> element.getValue().getPlacementNotice() != null)
+                .filter(element -> !element.getValue().getPlacementNotice().isEmpty())
+                .map(element -> element(element.getId(), element.getValue().nonConfidential(withNoticesResponses)))
+                .toList();
```
chak-shing-lo-justice, line 80: Just did a quick code search, looks like this method is invoked by ManageDoc event, migration service and the remove tools as well.
It should be fine for ManageDoc event, as user can only upload notice response when a notice exist.

But i am a bit concern about the removal tools. Will these changes disable admin from removing the placement application which don't have a notice yet? Or maybe potentially cause some leftover in the case data?

chak-shing-lo-justice: If it won't break any other stuff, could we rename this method a bit to avoid any confusion in the future. Because, instead of getting non-confidential placement, this method is now getting non-confidential placement with notice.

chak-shing-lo-justice (commented):
some questions about the changes

chak-shing-lo-justice (dismissed):
lgtm. But you may want to fix the code style check before moving the ticket?

PR #4737, opened 2023-10-05
FPL - Remove references to fpla-docker ready for removal of sub module

amoriartyMOJ:
https://tools.hmcts.net/jira/browse/DFPL-1745

### Change description ###
We want to remove FPLA docker as we don't use it. Instead we rely on CFTLib to run our service locally.

```
[ ] Yes
[X] No
```


bin/import-ccd-definition.sh
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
+# TODO: This script appears to be unused. Marked for future deletion.
+
 definition_input_dir=$(realpath 'ccd-definition')
 definition_output_file="$(realpath ".")/build/ccd-development-config/ccd-fpl-dev.xlsx"
 params="$@"
 params=${params:='-e *-prod.json,*-shuttered.json'}
-./fpla-docker/bin/import-ccd-definition.sh "${definition_input_dir}" "${definition_output_file}" "${params}"
+./bin/import-ccd-definition.sh "${definition_input_dir}" "${definition_output_file}" "${params}"
```
DanCatchpole, line 9: think this would cause a recursion - guess with the comment it appears to be unused anyways (was the fpla-docker version with the same name different somehow?)

amoriartyMOJ: Ah good spot. Yes the fpla-docker version was different
https://github.com/hmcts/fpla-docker/blob/master/bin/import-ccd-definition.sh

amoriartyMOJ: I'll delete it in this PR. It's not used

DanCatchpole (approved):

PR #4754, opened 2023-10-11
DFPL-1796: Filter out internal (WA) related case roles

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-1796

### Change description ###
 - When looking up case roles to determine if the user is a LA solicitor/respondent/child solicitor, it's picking up internal case roles for judges too now (as part of pre-req work for WA), need to filter them out as they are causing errors.

```
[ ] Yes
[X] No
```


FaisalMoJ (approved):

service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseAccessService.java
```diff
@@ -33,6 +38,9 @@ public class CaseAccessService {
     private final SystemUserService systemUserService;
     private final OrganisationService organisationService;
 
+    private final List<String> excludedCaseRoles = List.of(ALLOCATED_JUDGE.getRoleName(), HEARING_JUDGE.getRoleName(),
```
amoriartyMOJ, line 41: Could we name this something more descriptive for the long term?

E.g. excludedInternalWACaseRoles

That way we could remove the comment on the filter as that might get out of date in future and be miss leading

amoriartyMOJ (approved):

PR #4823, opened 2023-10-26
DFPL-1839 add DirtiesContext annotation to fix flaky unit tets

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1839

### Change description ###
Mark the test as DirtyContext, so Spring clear cache after each test

```
[ ] Yes
[X] No
```


amoriartyMOJ (approved):
Approving - This is more of an attempt than a clear solution as we're unsure of the root cause. We believe it could relate to caching of some values so trying this annotation will make sure spring clears before calling each test in this class. if it doesn't work, we can undo this change.

PR #4826, opened 2023-10-26
DFPL-1341 Enforce uploading draft order on C2 application

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1341

### Change description ###
Split draft order field into mandatory field and optional field.
If LA or solicitor, the mandatory is shown.
If not, the optional field is shown.

**Acceptance criteria:** 
1. The Local Authority or Solicitor uploads additional application 
2. C2 confirmation of what they are applying for is confirmed 
3. Application by consent or without notice is selected 
4. C2 application uploaded
5. Draft Order should be mandatory- Not Optional 
6. User cannot proceed without this confirmation
7. Validation to be provided-Error message confirming cannot proceed 
8. Existing behaviour applies 

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/UploadAdditionalApplicationsController.java
```diff
@@ -206,7 +220,8 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmit(@RequestBody Cal
 
         removeTemporaryFields(caseDetails, TEMPORARY_C2_DOCUMENT, "c2Type",
             "additionalApplicationType", AMOUNT_TO_PAY, "temporaryPbaPayment",
-            TEMPORARY_OTHER_APPLICATIONS_BUNDLE, "applicantsList", "otherApplicant", SKIP_PAYMENT_PAGE);
+            TEMPORARY_OTHER_APPLICATIONS_BUNDLE, "applicantsList", "otherApplicant", SKIP_PAYMENT_PAGE,
+            "isDraftOrderMandatory");
```
jamesrferguson1, line 224: maybe we should use IS_DRAFT_ORDER_MANDATORY instead here as we define it and use it in the above function?

PR #4859, opened 2023-11-01
Suppress CVE-2023-31582

chak-shing-lo-justice:
### Change description ###
Suppress CVE-2023-31582

```
[ ] Yes
[X] No
```

### DIFF
--- a/config/owasp/suppressions.xml
+++ b/config/owasp/suppressions.xml
@@ -42,4 +42,11 @@
     <packageUrl regex="true">^pkg:maven/org.springframework.plugin/spring-plugin-core@.*$</packageUrl>
     <cve>CVE-2023-34047</cve>
   </suppress>
+  <suppress>
+    <notes><![CDATA[
+   file name: jose4j-0.7.7.jar
+   ]]></notes>
+    <packageUrl regex="true">^pkg:maven/org.bitbucket.b_c/jose4j@0.*$</packageUrl>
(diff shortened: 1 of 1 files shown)


config/owasp/suppressions.xml
```diff
@@ -42,4 +42,11 @@
     <packageUrl regex="true">^pkg:maven/org.springframework.plugin/spring-plugin-core@.*$</packageUrl>
     <cve>CVE-2023-34047</cve>
   </suppress>
+  <suppress>
```
GauravChawlaMOJ, line 45: This can be fixed by incrementing the version to 0.9.3 according to this. May be we should fix this and not suppress?
https://nvd.nist.gov/vuln/detail/CVE-2023-31582

chak-shing-lo-justice: Just discussed in dev standup. If renovate bot don't update it, we can suppress it and avoid it from blocking PR  merging to master pipeline.

GauravChawlaMOJ: yes but if it is not explicitly defined in project, renovate won't update it as far as I know. For example Jose dependency which is causing this vulnerability is not defined in build.gradle, so it will never be updated by renovate.

chak-shing-lo-justice: I saw your PR is fixing it, but i guess it need some regression test before merging it to master? 
As this CVE is blocking the release.

PR #4922, opened 2023-11-22
Dfpl 1916

GauravChawlaMOJ:
### Change description ###

test password as secret in keyVault

```
[ ] Yes
[X] No
```

### DIFF
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -33,6 +33,13 @@ def integrationTestSecrets = [
   ]
 ]
 
+def e2eTestSecrets = [
+  'fpl-${env}': [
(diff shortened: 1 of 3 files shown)


Jenkinsfile_nightly
```diff
@@ -33,6 +33,13 @@ def integrationTestSecrets = [
   ]
 ]
 
+def e2eTestSecrets = [
+  'fpl-${env}': [
```
DanCatchpole, line 37: wonder if this needs to be fpl-aat, looking at the integrationTestSecrets above it looks like the env is hardcoded?
```suggestion
  'fpl-aat': [
```

DanCatchpole (approved):
lgtm!

PR #4961, opened 2023-11-30
DFPL-1937 remove line in base.js that causes sign in twice

Iphiclus:
https://tools.hmcts.net/jira/browse/DFPL-1937

### Change description ###
removing the following line of code in base.js

line 227
await I.goToPage(config.baseUrl, config.hmctsUser);
It is causes the 'navigateToCaseDetailsAs' method to sign in twice which is not required.

```
[ ] Yes
[x] No
```

### DIFF
--- a/e2e/actors/base.js
+++ b/e2e/actors/base.js
@@ -224,8 +224,6 @@ module.exports = {
 
   async navigateToCaseDetailsAs(user, caseId) {
     await I.goToPage(config.baseUrl);
-    await I.goToPage(config.baseUrl, config.hmctsUser);
-
     await this.signIn(user);
     I.wait(10);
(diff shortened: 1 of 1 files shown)


e2e/actors/base.js
```diff
@@ -224,8 +224,6 @@ module.exports = {
 
   async navigateToCaseDetailsAs(user, caseId) {
     await I.goToPage(config.baseUrl);
-    await I.goToPage(config.baseUrl, config.hmctsUser);
```
GauravChawlaMOJ, line 227: May be replace it with await I.goToPage(config.baseUrl, user);
and comment lines above and below this.

GauravChawlaMOJ (approved):

PR #5003, opened 2023-12-18
DFPL-1989 Documents added during case creation are not shown in CFV after submission

so99y:
https://tools.hmcts.net/jira/browse/DFPL-1989

### Change description ###
Documents added during case creation are not shown in CFV after submission

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicationDocumentsService.java
```diff
@@ -32,6 +84,7 @@ public Map<String, Object> updateApplicationDocuments(List<Element<ApplicationDo
         Map<String, Object> data = new HashMap<>();
 
         data.put("applicationDocuments", updatedDocuments);
```
jamesrferguson1, line 86: Maybe fine for now but will this not cause duplication? I.e. they'll be added under applicationDocuments and the new fields right?

qasimnawaz-moj (dismissed):

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ApplicationDocumentsService.java
so99y: yes, in the case submission state we still use `applicationDocuments` and for a quick fix, we run the migration logic to calculate the new fields.

jamesrferguson1 (approved):

qasimnawaz-moj (approved):

PR #5027, opened 2024-01-05
DFPL-1624 Send judicial message or reply to message when case closed

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1624

### Change description ###
DFPL-1624 Send judicial message or reply to message when case closed

```
[ ] Yes
[X] No
```


so99y (approved):
LGTM

PR #5059, opened 2024-01-15
Nightly dev

GauravChawlaMOJ:
### Change description ###

nightly-dev merge into master

```
[ ] Yes
[X] No
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,9 @@ infrastructure/.terraform/
 
 # vscode
 .vscode/*
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
+tests-examples/demo-todo-app.spec.ts
+playwright-e2e/local-config/config.js
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -106,25 +106,24 @@ withNightlyPipeline(type, product, component) {
     env.CASE_SERVICE_URL = "http://fpl-case-service-aat.service.core-compute-aat.internal"
   }
 
-  before('crossBrowserTest') {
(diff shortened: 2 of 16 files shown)


ptrelease:
Looks great.  Some minor points:

* Without a change desc above, I'm assuming the cross browser code is commented out intentionally?
* I couldn't figure out why the CreateCase method in [this file](playwright-e2e/pages/create-case.ts) mixed an approach of using names like 'caseTypeFilterDropdown' then other times getting by label or role.
* Interim care order is checked, then immediately unchecked [here](playwright-e2e/pages/orders-and-directions.ts) sure there is a good reason, but not what.

ptrelease (approved):
Looks good.  Some minor comments in conversation section.

Iphiclus:
> Looks great. Some minor points:
> 
> * Without a change desc above, I'm assuming the cross browser code is commented out intentionally?
> * I couldn't figure out why the CreateCase method in this file playwright-e2e/pages/create-case.ts mixed an approach of using names like 'caseTypeFilterDropdown' then other times getting by label or role.
> * Interim care order is checked, then immediately unchecked in this file, playwright-e2e/pages/orders-and-directions.ts. Sure there is a good reason, but not what it is.

Hi, cross browser isn't being used atm, and hasn't been used for a while (possibly over a year). Playwright is likely to replace Sauce Labs as cross browser is built in to Playwright (for free, where as Sauce Labs is paid for).
Create Case is WIP
Orders and directions is also WIP
Only test being run is the smoke-test.spec.ts currently.
Hopefully that helps! Thanks for review Paul! :) 
Let me know if you have any questions.

PR #5087, opened 2024-01-18
DFPL-2096: LA's unable to edit existing placements

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-2096

Add read permissions back to placementApplication and supportingDocuments so they can be populated with the existing placement's values, caused overwriting/removal of some placement documents in older closed cases.

*also renamed childrenWithoutPlacements -> children as getAllChildren() doesn't exclude based on placement

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
lgtm

PR #5154, opened 2024-02-19
DFPL-1918: Bring back old setup judicial roles migration for rerun

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-1918

### Change description ###
 - Bringing back the runAM + runAMRollback migrations for one final run before WA NRO 

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
@@ -59,7 +59,6 @@ public class MigrateCaseController extends CallbackController {
         "DFPL-CFV-Failure", this::runCfvFailure,
         "DFPL-CFV-dry", this::dryRunCFV,
         "DFPL-1940", this::run1940,
-        "DFPL-2177", this::run2177,
         "DFPL-AM", this::runAM,
         "DFPL-AM-Rollback", this::runAmRollback,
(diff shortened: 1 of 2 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
```diff
@@ -198,4 +216,140 @@ private void run2177(CaseDetails caseDetails) {
...
+                    }
+                }
+                return el;
+            }).toList();
+
+        if (!hearings.isEmpty()) {
+            // don't add an empty array if there weren't any hearings beforehand
+            caseDetails.getData().put("hearingDetails", modified);
+        }
+        caseDetails.getData().put("hasBeenAMMigrated", "Yes");
```
so99y, line 347: I guess the existence of  `hasBeenAMMigrated` will be checked in migration tool. Shall we check it in `runAM` as well?

DanCatchpole: it must be present or the elasticsearch query for the rollback won't select the case, doesn't matter if it's present or not for the migration, as it'll delete + recreate roles anyways (and some cases will still have it from before, some wont and they all need migrating this time around!)

so99y (dismissed):

so99y (approved):

PR #5192, opened 2024-03-15
DFPL-2167-Respondent-Details (added respondent details to smoke test)

abigailsmith97:
https://tools.hmcts.net/jira/browse/DFPL-2167

### Change description ###
added 'respondent details' to smoke test 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/playwright-e2e/tests/smoke-test.spec.ts
+++ b/playwright-e2e/tests/smoke-test.spec.ts
@@ -11,8 +11,10 @@ test("Smoke Test @smoke-test", async ({
   groundsForTheApplication,
   riskAndHarmToChildren,
   factorsAffectingParenting,
+  respondentDetails,
   allocationProposal,
   addApplicationDocuments,
+  
   page
 }) => {
   const basePage = new BasePage(page);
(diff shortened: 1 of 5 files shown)


playwright-e2e/tests/smoke-test.spec.ts
```diff
@@ -69,6 +71,12 @@ test("Smoke Test @smoke-test", async ({
   await addApplicationDocuments.uploadDocumentSmokeTest();
   await startApplication.addApplicationDocumentsInProgress();
 
+  await page.pause();
```
Iphiclus, line 74: await page.pause(); perhaps not needed?

Iphiclus (changes requested):

playwright-e2e/tests/smoke-test.spec.ts
```diff
@@ -69,6 +71,12 @@ test("Smoke Test @smoke-test", async ({
...
+  // Add respondents' details
+  await startApplication.respondentsDetailsLink.isVisible();
+  await startApplication.respondentDetails();
+  await respondentDetails.respondentDetailsNeeded();
```
GauravChawlaMOJ, line 78: May be 'await startApplication.respondentsDetailsLink.isVisible();' can be added into startApplication.respondentDetails() method?

playwright-e2e/tests/smoke-test.spec.ts
abigailsmith97: ah yes! good spot, probably need to start using the breakpoints instead..

playwright-e2e/tests/smoke-test.spec.ts
abigailsmith97: i think this was actually duplicated perhaps so have removed.

Iphiclus (changes requested):
Hi, great stuff, there is a missing radio button towards the end of the test.
"Do they have legal representation?" which looks to missing, which leaves the status of this section as 'in progress'. Which would stop the case being submitted as far as I am aware. Have a quick look if this needs adding?

GauravChawlaMOJ (approved):

Iphiclus (approved):

PR #5193, opened 2024-03-18
DFPL-2229: Fix flaky async tests verifying things in the wrong order

DanCatchpole:
DFPL-2229

### Change description ###
 - Instead of checking for individual invocations, check that exactly 2 events happen (WITH TIMEOUTS) and we capture the correct 2 event IDs (in any order)

```
[ ] Yes
[X] No
```


AlistairEdwinOsborne (approved):
looks good to me, good luck!

PR #5194, opened 2024-03-18
DFPL-2208 no longer need to attach the shared inbox from the config

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-2208

### Change description ###
From ticket: we no longer need to attach the shared inbox from the config in any circumstances (as it is autofilled during case creation, and then possibly overridden on cases, so we're doing the reverse of what the applicant would have asked for)

```
[ ] Yes
[X] No
```


DanCatchpole (approved):
lgtm! didn't expect so many tests to not have the localAuthorities field before 😄

PR #5205, opened 2024-03-22
Vs code api helper changes

abigailsmith97:
N/A

### Change description ###
Changes to the api helper and config due to clashes in how intelliJ and vscode read the files.

```
[ ] Yes
[ x] No
```


GauravChawlaMOJ (approved):

playwright-e2e/settings/user-credentials.ts
```diff
@@ -1,8 +1,8 @@
 import dotenv from "dotenv";
 dotenv.config();
 
-const e2ePw = process.env.E2E_TEST_PASSWORD || '';
-const defaultPwd = process.env.SYSTEM_UPDATE_USER_PASSWORD || '';
+const e2ePw = process.env.E2E_TEST_PASSWORD || 'Password1234';
```
Iphiclus, line 4: Hi, do we need the passwords hardcoded?

GauravChawlaMOJ: May be not, if it fails to read environment variable in IDE, then add it into your local copy of project only and not commit? Also, fortify scan would catch this as a critical issue I suppose.

playwright-e2e/utils/api-helper.ts
```diff
@@ -15,7 +15,7 @@ export class Apihelp {
           'Content-Type': 'application/x-www-form-urlencoded',
         },
       };
-      let url = `${urlConfig.idamUrl}loginUser?username=${user.email}&password=${user.password}`;
+      let url = `https://idam-api.aat.platform.hmcts.net/loginUser?username=${user.email}&password=${user.password}`;
```
DanCatchpole, line 18: is there any reason this was changed from the config value, as would break tests against demo?

Iphiclus: fixed

DanCatchpole (approved):
lgtm!

Iphiclus (approved):

PR #5221, opened 2024-04-02
Preview tests running in all envs

GauravChawlaMOJ:
### Change description ###

Preview tests will only run in preview environment now

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/build.gradle
+++ b/service/build.gradle
@@ -365,6 +365,7 @@ dependencies {
   implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-joda', version: '2.15.3'
 
   implementation group: 'net.minidev', name: 'json-smart', version: '2.5.0'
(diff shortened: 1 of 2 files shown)


service/build.gradle
```diff
@@ -365,6 +365,7 @@ dependencies {
   implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-joda', version: '2.15.3'
 
   implementation group: 'net.minidev', name: 'json-smart', version: '2.5.0'
+  implementation group: 'com.nimbusds', name: 'nimbus-jose-jwt', version: '9.37.3'
```
jamesrferguson1, line 368: Guess this is for the new CVE just now. Looks like the Spring 2.7 upgrade fixes this without having to add this extra dependency explicitly. Wonder if we should maybe suppress for now and remove when the Spring upgrade is released?

GauravChawlaMOJ: sounds good! Removed it

DanCatchpole (approved):

PR #5288, opened 2024-04-22
DFPL-2009: Allow optional hearing to show in notification template

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-2009

Add branch for notification for application submitted to fill in subject line with Application received - hearing other when no hearing is provided to allow for CTSC to properly sort/triage

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/email/content/base/SharedNotifyContentProvider.java
```diff
@@ -43,6 +43,13 @@ protected <T extends SharedNotifyTemplate> T buildNotifyTemplate(T template, Cas
         template.setNonUrgentHearing(timeFrame.isPresent() && !timeFrame.get().equals("Same day")
                                      ? YES.getValue() : NO.getValue()
         );
+
+        //When hearing is not filled in put make the subject line Application Received - hearing other
+        if (template.getUrgentHearing().equals(NO.getValue()) && template.getNonUrgentHearing().equals(NO.getValue())) {
```
so99y, line 48: I'm not sure if we need to check null on template.getUrgencyHearing(). It's better to change it to `NO.getValue()(template.getUrgentHearing()`

AlistairEdwinOsborne: there shouldn't be a null value in either of those values due to the branching in the two statements above as when its not filled in they will be both be set to NO by default

so99y (approved):
LGTM

PR #5400, opened 2024-05-22
DFPL-2340: Removal tool throwing exception when getting c2doc label

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-2340

Return label rather than throw exception when c2bundle is empty

Replace with proper looking for C2confidentialbundle

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/AdditionalApplicationsBundle.java
```diff
@@ -66,7 +66,7 @@ public String toLabel() {
             return otherApplicationsBundle.toLabel();
         }
 
-        throw new MissingApplicationException(uploadedDateTime);
+        return otherApplicationsBundle.getUploadedDateTime();
```
DanCatchpole, line 69: if there's not an "other" bundle inside the bundle this will cause an NPE I think

DanCatchpole (approved):

PR #5405, opened 2024-05-23
RemoveAccessibility

prabhamuthu15:
[DFPL-2320](https://tools.hmcts.net/jira/browse/DFPL-2320)

### Change description ###

Accessibility test

```
[ ] Yes
[X ] No
```


gavinkeys (changes requested):
Hey, I'd suggest just removing this code rather than commenting it out as it's not needed any more. Also have you tested this on the nightly-dev pipeline?

PR #5520, opened 2024-07-30
DFPL-2316: c110a Design Changes to Court Services Needed

qasimnawaz-moj:

https://tools.hmcts.net/jira/browse/DFPL-2316

### Change description ###

c110a Design Changes to Court Services Needed

```
[ ] Yes
[X] No
```


ccd-definition/ComplexTypes/CareSupervision/HearingPreferences.json
```diff
@@ -2,150 +2,55 @@
...
-    "FieldType": "Label",
-    "ElementLabel": "Is any of the following needed to help someone take part in a hearing?",
-    "SecurityClassification": "Public"
-  },
-  {
-    "LiveFrom": "01/01/2017",
-    "ID": "HearingPreferences",
-    "ListElementCode": "interpreter",
-    "FieldType": "YesOrNo",
-    "ElementLabel": "Interpreter",
```
DanCatchpole, line 15: I wonder if instead of moving these to the deprecated file (which may cause confusion later as data will be hanging around with this stuff, and then models don't line up 1:1 with the ComplexType files, if it may be better leaving them in and just showConditioning them to something = "DEPRECATED" instead of "DO_NOT_SHOW" (and maybe moving to the bottom of the file?)

In two minds about it, as the DEPRECATED file already exists...

service/src/main/java/uk/gov/hmcts/reform/fpl/model/HearingPreferences.java
```diff
@@ -12,12 +15,14 @@ public class HearingPreferences {
     private final String interpreter;
     private final String intermediary;
     private final String welshDetails;
```
DanCatchpole, line 17: I guess adding to the other comment about these fields being deprecated, could annotate them as such in the model so we don't use them accidentally in the future

DanCatchpole (approved):

PR #5540, opened 2024-08-15
DFPL-1762

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1762

### Change description ###
Build POST endpoint for CAFCASS to upload guardian information

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/cafcass/api/CafcassApiGuardianService.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public boolean validateGuardians(List<Guardian> guardianUpdateList) {
+        return guardianUpdateList.stream()
+            .noneMatch(guardian -> isEmpty(guardian.getGuardianName())
+                                   || isEmpty(guardian.getChildren())
+                                   || guardian.getChildren().stream().anyMatch(String::isEmpty));
+    }
+
+    public boolean checkIfAnyGuardianUpdated(CaseData caseData, List<Guardian> guardianUpdateList) {
+        List<Element<Guardian>> existingGuardians = caseData.getGuardians();
```
jamesrferguson1, line 33: Probably me just not understanding this  - would this not always return true if there are existing guardians and their details haven't changed?

chak-shing-lo-justice: This method will always return false (means no changes) if the existing and update guardians lists are identical regardless of their order.

If there are existing guardians and their details haven't changed but the update list add / delete any guardian, the method will still return true (means has changes).

If there is no changes, we won't fire the CCD event.

jamesrferguson1 (approved):

PR #5547, opened 2024-08-21
DFPL-1767: Add POST endpoint for uploading documents from Cafcass API

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-1767

- Add endpoint and method for uploading documents 
- Allow for the implementation of other document types than GUARDIAN REPORTS

```
[ ] Yes
[X] No
```


service/src/integrationTest/java/uk/gov/hmcts/reform/fpl/controllers/cafcass/CafcassCasesControllerTest.java
```diff
@@ -26,34 +56,80 @@
 public class CafcassCasesControllerTest extends AbstractTest {
```
chak-shing-lo-justice, line 56: 1. Remove the `Deprecated` annotation if you wanna reuse this file.
1. Also, maybe rename this test to something like CafcassApiDocumentUploadControllerTest, as it only covers upload document.

FYI, there is a CafcassApiControllerBaseTest which provide some basic stuff for Cafcass API controller unit test. You can extend it if you like.

chak-shing-lo-justice (commented):
Main concern is the type of document supported. Did Cafcass confirmed that they will only upload guardian reports?

Others comment are just minor thought trying to align the coding style

PR #5552, opened 2024-08-28
Add populate case - returned event

DanCatchpole:
### Change description ###
 - Adds populateCase-RETURNED variant so we can create test cases that start in the Returned state

```
[ ] Yes
[X] No
```


Braimah101 (approved):
Good to go

PR #5559, opened 2024-09-02
DFPL-2516: Add missing nextHearingDetails field to populate hearing info on "My Cases"

DanCatchpole:
[DFPL-2516](https://tools.hmcts.net/jira/browse/DFPL-2516)

### Change description ###
 - Add new field `nextHearingDetails`, which has property `nextHearingDate` (`DateTime` type), which is pulled from by ExUI on the "My Cases" / "All Cases" tab.
 - Can migrate to this or just let it populate before switching our caseSummaryNextHearingDateTime which does the same thing (TBC, can perform later if needed).

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
lgtm

PR #5578, opened 2024-09-06
DFPL 2499

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-2499

### Change description ###
We are going to set up a system user for Cafcass with the right set of roles to access our new Cafcass APIs.

This PR: update the validation of the access token. add Cafcass system user configurartion

```
[ ] Yes
[X] No
```


charts/fpl-case-service/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: fpl-case-service
 apiVersion: v2
 home: https://github.com/hmcts/fpl-ccd-configuration
-version: 1.12.79
+version: 1.12.80
```
jamesrferguson1, line 4: Is this needed? Doesn't look like anything's changed here.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/cafcass/api/CafcassApiDocumentService.java
```diff
@@ -10,8 +10,9 @@
 @RequiredArgsConstructor(onConstructor = @__(@Autowired))
 public class CafcassApiDocumentService {
     private final SecureDocStoreService secureDocStoreService;
+    private final CafcassSystemUserService cafcassSystemUserService;
 
     public byte[] downloadDocumentByDocumentId(String documentId) throws IllegalArgumentException, EmptyFileException {
-        return secureDocStoreService.downloadDocument(documentId);
+        return secureDocStoreService.downloadDocument(documentId, cafcassSystemUserService.getUserToken());
```
DanCatchpole, line 16: Can we not use the calling user's token here like we normally do when calling the docstore (on requestData)? That should be the cafcass system user, saves logging in again/if ever an accidental bug allows the non-docstore user to login.

chak-shing-lo-justice: The IDAM /userinfo endpoint will only return user name but not the other details (e.g. user roles) if the token is generated via password grant. 
CCD / CDAM will throw exception if they can't grab the user role with the tokens.

That's why we need to get the token via login, such that IDAM will return a full userinfo

PR #5589, opened 2024-09-18
DFPL 2214 Fix c2 draft order and duplicated doc

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-2214

### Change description ###
1. Prevent C2 Draft orders from being sent as C2 applications. In existing Sendgrid code, C2 draft order was sent as draft orders instead of C2 applications
2. Update to return distinct case documents list 
3. Fix duplicated or confidential C1/C2 document when converting ManagedDocument

```
[ ] Yes
[X] No
```


AlistairEdwinOsborne (approved):
all good by me!

PR #5590, opened 2024-09-19
DFPL-2441: Hide history tab for external parties

DanCatchpole:
https://tools.hmcts.net/jira/browse/DFPL-2441

### Change description ###
 - Remove historyTab permissions for all external users
 - Move caseName + familyManCaseNumber off of the history tab and onto other tabs that are always shown in all states (so they work in the title at the top of the case!)

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
lgtm

PR #5641, opened 2024-10-17
Cafcass api test

prabhamuthu15:
[DFPL-1726](https://tools.hmcts.net/jira/browse/DFPL-1726)

### Change description ###

API test for the new cafcass end points

```
[ ] Yes
[ x] No
```


Jenkinsfile_nightly
```diff
@@ -67,94 +67,114 @@ def teardownSecretsForIntegrationTests() {
...
-            allowMissing: true,
-            alwaysLinkToLastBuild: true,
-            keepAll: true,
-            reportDir: 'playwright-report',
-            reportFiles: 'index.html',
-            reportName: 'UITestWebkitReport'
-          ])
-          steps.archiveArtifacts allowEmptyArchive: true, artifacts: 'build/reports/tests/functionalTest/**/*'
-        }
+    stage('cafCassAPITest') {
```
GauravChawlaMOJ, line 156: do you want to add this stage in both main and nightly pipeline?

prabhamuthu15: The new endpoints does not trigger any state changes in FPL service. These end points  are used only by cafcass user not by FPL service so having them nightly will be sufficient. If added to master build it going to add extra constrain on each release and doesn't add any values to service.

GauravChawlaMOJ: ok, the stage is in Jenkins_cnp file as well, may be it can be removed from there then?

prabhamuthu15: Yes, I have removed Jenkin_cnp , so the tests run only in nightly builds.

PR #5646, opened 2024-10-21
DFPL-1607 synchronized locking on postSubmit Call back by Case id

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-1607

### Change description ###
By using ConcurrentHashMap, achieving synchronisation locking on the performPostSubmitCallback by case id.

AtomicInteger is used to keep track on the number of active performPostSubmitCallback on a case. The corresponding entry will be removed if there is no more active performPostSubmitCallback on that case.

```
[ ] Yes
[X] No
```


qasimnawaz-moj (approved):
lgtm!

PR #5705, opened 2024-11-13
DFPL-1749: Reduce Code Smells

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-1749

- Add constants for strings repeated multiple times
- Add asserts to tests missing them
- Refactor multiple if statements to switches where appropriate

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
LGTM.

Though there are same "case field" string constant scattering around, think the best practice would be updating the caseDetails map within its corresponding service. e.g. updating "allocatedJudge" field within AllocatedJudgeService only. then we won't have to create the same "allocatedJudge" constant in two different controllers.

But don’t think it worth the effort and risk, as this will be a lot of changes across the service and put lots of pressure on regression test.

PR #5716, opened 2024-11-15
DFPL-2544 Create alert for async call back thread queue

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-2544

### Change description ###
Add monitoring to "executor_pool_size". Alert if reach zero

```
[ ] Yes
[X] No
```


DanCatchpole (changes requested):
not sure the query will work as expected, looks like somewhat regularly in prod that metric hits 0, based on the previous 3 days

DanCatchpole (approved):

PR #5696, opened 2024-11-08
DFPL-2546 Show Refuge Address- address option for Respondent/Other persons Journeys

chak-shing-lo-justice:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/DFPL-2546

### Change description ###
DFPL-2546 Show Refuge Address- address option for Respondent/Other persons Journeys

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[X] No
```

PR #5755, opened 2024-12-02
DFPL-2546 Refuge Address option for Respondent/Other persons Journeys

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-2546

### Change description ###
Refuge Address option for Respondent/Other persons Journeys.

#5696 Release failed as the pipeline failed to import CCD definition. The ticket was rolled back.
Error message was:
`ElasticSearch initialisation exception" mapper [data.hearingDetails.value.others.value.addressKnow] cannot be changed from type [keyword] to [text]`

This PR fix the issue by introducing a new field "addressKnowV2" insetad of updating the type of the existing "addressKnow" field
The old field is deprecated, and the application will convert the old field to new field during case data deserialisation.

```
[ ] Yes
[X] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
```diff
@@ -48,7 +52,12 @@ public class Other implements Representable, ConfidentialParty<Other> {
     private final String detailsHiddenReason;
     private List<Element<UUID>> representedBy;
     private final String addressNotKnowReason;
-    private final IsAddressKnowType addressKnow;
+
+    // Flag for preventing from purging the converted old field value during deserialization if addressKnowV2 is null
+    @Builder.Default
+    @Getter(AccessLevel.NONE)
+    private final boolean isConvertedAddressKnow = false;
```
jamesrferguson1, line 59: Does this get used? Can see it's also defined on 173

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
```diff
@@ -48,7 +52,12 @@ public class Other implements Representable, ConfidentialParty<Other> {
     private final String detailsHiddenReason;
     private List<Element<UUID>> representedBy;
     private final String addressNotKnowReason;
-    private final IsAddressKnowType addressKnow;
```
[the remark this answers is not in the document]
chak-shing-lo-justice: Think no,  I only remove the Java side but not the CCD definition.

The 'old' addressKnow in historical data will always be converted to new addressKnowV2 during deserialisation. And builder is the only way to construct this object.

Also there is no getter for the 'old' addressKnow field, so when updating the case data, the old field won't be serialised in the Json anymore.

Have tested this in both local and preview env with some historical data, and can't see any issue.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/Other.java
chak-shing-lo-justice: no, it won't get serialised to Json, as @Getter(AccessLevel.NONE) mean no getter is generated for this field

chak-shing-lo-justice: oh! right, that one in builder is redundant

chak-shing-lo-justice: My bad, not that one in the builder, this one is redundant. Just did some simple test, that one in builder is thread safe. so should be fine

jamesrferguson1 (dismissed):

PR #5788, opened 2024-12-13
DFPL-1976: Add freemarker template rendering + welsh configuration

DanCatchpole:
[DFPL-1976](https://tools.hmcts.net/jira/browse/DFPL-1976)

### Change description ###
 - Replaces code template generation with a template (Apache FreeMarker)
 - Uses markdown tables to ensure formatting works
 - Converts task states to be HTML rendered, not image rendered
 - Adds welsh translation of event names + states
 - Adds welsh error messages

**_TODO:_** 
- Add permissions for other applicant roles
- Remove option for combined task list after demo if confirmed not using that approach
- Remove existing code for generating the task list (currently deprecated)

```
[ ] Yes
[x] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseEventHandler.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/handlers/CaseEventHandler.java
@@ -5,7 +5,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.event.EventListener;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.ccd.client.model.CaseDetails;
-import uk.gov.hmcts.reform.fpl.enums.Event;
 import uk.gov.hmcts.reform.fpl.events.CaseDataChanged;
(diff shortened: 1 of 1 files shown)


ccd-definition/CaseField/CareSupervision/caseField.json
```diff
@@ -89,6 +89,15 @@
     "SecurityClassification": "Public",
     "Searchable": "N"
   },
+  {
+    "LiveFrom": "01/01/2018",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "taskListWelsh",
+    "Label": "Case name",
```
jamesrferguson1, line 96: Does this label need to be in Welsh

jamesrferguson1 (dismissed):

ccd-definition/CaseField/CareSupervision/caseField.json
DanCatchpole: Shouldn't be necessary, this is the hidden field which is then embedded in a different field for markdown rendering, so this won't be displayed, not even sure why it was labeled as "case name" in english either though 😆

PR #5793, opened 2024-12-18
DFPL-2611 Update user role for third party applicants to be EPSMANAGING

jamesrferguson1:

https://tools.hmcts.net/jira/browse/DFPL-2611

### Change description ###

On case initiation for third party applicants, the case creator's org should be assigned EPSMANAGING instead of SOLICITORA or CHILDSOLICITORA (depending on option selected in drop down they see when creating a case).

```
[ ] Yes
[ x] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseInitiationService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/CaseInitiationService.java
@@ -30,11 +30,10 @@ import static java.util.Objects.nonNull;
 import static org.apache.commons.lang3.ObjectUtils.isEmpty;
 import static org.apache.commons.lang3.ObjectUtils.isNotEmpty;
 import static uk.gov.hmcts.reform.ccd.model.OrganisationPolicy.organisationPolicy;
-import static uk.gov.hmcts.reform.fpl.enums.CaseRole.CHILDSOLICITORA;
 import static uk.gov.hmcts.reform.fpl.enums.CaseRole.CREATOR;
+import static uk.gov.hmcts.reform.fpl.enums.CaseRole.EPSMANAGING;
 import static uk.gov.hmcts.reform.fpl.enums.CaseRole.LAMANAGING;
 import static uk.gov.hmcts.reform.fpl.enums.CaseRole.LASOLICITOR;
-import static uk.gov.hmcts.reform.fpl.enums.CaseRole.SOLICITORA;
 import static uk.gov.hmcts.reform.fpl.enums.OutsourcingType.EPS;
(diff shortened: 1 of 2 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/noc/NoticeOfChangeFieldPopulator.java
```diff
@@ -64,6 +65,30 @@ public Map<String, Object> generate(CaseData caseData, Representing representing
...
 
+    public Map<String, Object> generateApplicantAnswer(CaseData caseData) {
+        Map<String, Object> data = new HashMap<>();
+        NoticeOfChangeAnswers nocAnswers = NoticeOfChangeAnswers.builder().build();
+
+        if (caseData.isOutsourced()) {
+            LocalAuthority localAuthority = caseData.getLocalAuthorities().stream()
+                .map(Element::getValue)
+                .filter(la -> la.getId().equals(caseData.getOutsourcingPolicy().getOrganisation().getOrganisationID()))
+                .findFirst().orElse(null);
```
DanCatchpole, line 76: findFirst() returns an optional - so could potentially use that instead of the explicit null check after

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/NoticeOfChangeController.java
```diff
@@ -52,7 +61,17 @@ public CallbackResponse handleAboutToStart(@RequestBody CallbackRequest request)
 
         caseData = getCaseData(caseDetails);
 
-        caseDetails.getData().putAll(legalCounselUpdater.updateLegalCounselFromNoC(caseData, originalCaseData));
+        ChangeOrganisationRequest nocRequest = caseData.getChangeOrganisationRequestField();
+
+        Map<String, Object> changeOrgRequestField = (Map<String, Object>) caseDetails.getData().get(
+            "changeOrganisationRequestField");
```
DanCatchpole, line 67: think this is the same data in two different ways - probably can just use nocRequest? (and then re-use it in lines 69-74)

DanCatchpole (approved):

ccd-definition/AuthorisationCaseEvent/CareSupervision/AuthorisationCaseEvent.json
```diff
@@ -519,8 +519,7 @@
           "[EPSMANAGING]",
           "[LAMANAGING]",
           "[SOLICITORA]",
-          "[CHILDSOLICITORA]",
-          "[APPSOLICITOR]"
```
DanCatchpole: need to keep this one, old 3rd party cases could still use this route

DanCatchpole (dismissed):
lgtm!

PR #5849, opened 2025-02-21
DFPL-2689 Flaky test stability

AustenStevensMOJ:
### Change description ###
One line change - continue button showed a flaky pattern and sometimes requires a single continue press, to accommodate for environment flakiness a second continue press is required after a few milliseconds

```
[ ] Yes
[ x] No
```


playwright-e2e/pages/base-page.ts
```diff
@@ -54,7 +54,7 @@ export class BasePage {
   }
 
   async clickContinue() {
-    await this.continueButton.click();
+    await this.continueButton.click({clickCount:2, delay:300});
```
chak-shing-lo-justice, line 57: Does that mean Playwright auto-waiting is not enough in this situation?
https://playwright.dev/docs/actionability

Wondering would it be better if we check everything in network is completed.
`await page.waitForLoadState('networkidle');`

chak-shing-lo-justice (approved):

PR #5880, opened 2025-03-11
Fix for bug where validation enforced on field in grounds for application preventing submission of case

qasimnawaz-moj:
### Change description ###

Fix for bug where validation enforced on field that is not populated in grounds for application preventing submission of case

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
lgtm

PR #6023, opened 2025-05-21
DFPL-2615 Orders Coversheet for Judges (C2)

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-2615

### Change description ###
Add Cover sheet when Judges approved a C2 draft order

```
[ ] Yes
[X] No
```


AlistairEdwinOsborne (approved):
look all good! btw for any references to the user service/judicial user service should we always do the @retryable tag? I got stuff like that in my pba api ticket

service/src/main/java/uk/gov/hmcts/reform/fpl/service/JudicialService.java
```diff
@@ -549,4 +550,65 @@ private List<RoleAssignment> getAllocatedJudgeAndLegalAdvisorRoleAssignments(Lon
     private List<RoleAssignment> getHearingJudgeAndLegalAdviserRoleAssignments(Long caseId, ZonedDateTime endTime) {
         return roleAssignmentService.getCaseRolesAtTime(caseId, HEARING_ROLES, endTime);
     }
+
+    public List<JudicialUserProfile> getJudicialUserProfiles(JudicialUserRequest request) {
```
jamesrferguson1, line 554: I wonder if there's a way we could combine some of these so that we are maybe able to limit the number of requests to the judicial API? 

E.g. could we combine getJudgeTitleAndNameOfCurrentUser and isCurrentUserFeePaidJudge so we only have one call to get all this information?

chak-shing-lo-justice: I have added Cacheable to JudicialApi.findUsers(). So it shouldn't call the judicial API multiple times.

PR #6029, opened 2025-06-09
DFPL-2784: Show Represeting as empty if on LAApplicant flow

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-2784

Only set representingDetails if first and last name are present

```
[ ] Yes
[X] No
```


chak-shing-lo-justice (approved):
lgtm

PR #6273, opened 2026-04-21
DFPL-3145: Generate generic task instead of error when no valid am role

AlistairEdwinOsborne:
https://tools.hmcts.net/jira/browse/DFPL-3145

Remove error message and replace with generic ctsc task when additional application is uploaded when no valid AM role for allocated judge/legal adviser on case

Required to allow closed cases and cases with invalid roles to upload additional apps

```
[ ] Yes
[X] No
```

### DIFF
--- a/service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
+++ b/service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
@@ -9,7 +9,6 @@ import uk.gov.hmcts.reform.fpl.enums.ApplicationType;
 import uk.gov.hmcts.reform.fpl.enums.CaseRole;
 import uk.gov.hmcts.reform.fpl.enums.JudicialMessageRoleType;
 import uk.gov.hmcts.reform.fpl.enums.YesNo;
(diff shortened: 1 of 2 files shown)


service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/UploadAdditionalApplicationsService.java
```diff
@@ -367,9 +366,8 @@ public JudicialMessageRoleType getAllocatedJudgeOrLegalAdviserType(CaseData case
             } else if (roleTypes.contains(ALLOCATED_LEGAL_ADVISER.getRoleName())) {
                 return JudicialMessageRoleType.OTHER;
             } else {
-                throw new UserLookupException(
-                    String.format("Allocated judge or legal adviser has invalid am role for case id: %s",
-                        caseData.getId()));
+                // If no valid AM roles on case return a generic task
+                return JudicialMessageRoleType.CTSC;
```
hfilmore, line 372: is it worth logging something?

AlistairEdwinOsborne: Seeing as they no longer want to consider this an error I don't think a logging message is required

hfilmore (approved):

PR #6290, opened 2026-05-27
DFPL-3254 Remove case creation logic that shares with all LA users

chak-shing-lo-justice:
https://tools.hmcts.net/jira/browse/DFPL-3254

### Change description ###
DFPL-3254 Remove case creation logic that shares with all LA users

```
[ ] Yes
[X] No
```


jamesrferguson1 (dismissed):

AlistairEdwinOsborne (dismissed):
all good!

PR #6295, opened 2026-06-03
DFPL-3151: Replace Child/Respondent Selector/w DynamicMultiSelectList

JustGC-01:
https://tools.hmcts.net/jira/browse/DFPL-3151

Replacing Respondent/Child Selectors with DynamicMultiSelectList 

```
[] Yes
[x] No
```


ccd-definition/CaseField/CareSupervision/createOrderFields.json
```diff
@@ -105,9 +105,19 @@
...
+    "Label": " ",
     "FieldType": "ChildSelector",
     "SecurityClassification": "Public",
+    "Searchable": "N",
+    "Comment": "Field still used in closeCase & recordFinalDecisions events"
+  },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "ID": "childSelectorForManageOrders",
```
chak-shing-lo-justice: Just thinking if this childSelector could be reused by other event in the future.
If yes, sounds like we should name it as sth like `childSelectorV2` or `childMultiSelector`

service/src/main/java/uk/gov/hmcts/reform/fpl/model/common/dynamic/DynamicMultiSelectList.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @JsonIgnore
+    public UUID getValueCodeAsUuid() {
+        return Optional.ofNullable(getValueCode()).map(UUID::fromString).orElse(null);
+    }
+
+    @JsonIgnore
+    public String getValueCode() {
+        return value == null ? null : value.toString();
+    }
```
chak-shing-lo-justice, line 47: looks like `getValueLabel`, `getValueCodeAsUuid`, `getValueCode` are just copy and paste, and not really usable. 
Probably we don't need them for now, in case of having any confusion in the future

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
```diff
@@ -499,8 +500,9 @@ public List<AdditionalApplicationType> getAdditionalApplicationType() {
     private final Integer orderMonths;
     private final InterimEndDate interimEndDate;
     private final Selector childSelector;
+    private final DynamicMultiSelectList childSelectorForManageOrders;
```
chak-shing-lo-justice: If `childSelectorForManageOrders` is created specifically for `Manage order` only, could we put it in `ManageOrdersEventData` instead of `CaseData`?

We are having an issue that too many class property & value in `CaseData`, and reaching the maximum amount of constructor / builder parameters allowed.

If fields are only being used in a particular event(s), we will group it in a class named as `XxxxEventData`

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ChildrenService.java
```diff
@@ -198,4 +202,44 @@ public boolean hasAddressChange(List<Element<Child>> after, List<Element<Child>>
             Collections.unmodifiableList(before));
     }
 
+    public DynamicMultiSelectList getChildrenMultiSelectList(CaseData caseData) {
+        List<DynamicMultiSelectListElement> listItems = new ArrayList<>();
+        if (caseData.getChildren1() != null) {
+            IncrementalInteger i = new IncrementalInteger(1);
+            caseData.getChildren1().forEach(child -> listItems.add(DynamicMultiSelectListElement
+                .builder().code(child.getId().toString()).label(child.getValue().getParty().getFirstName() + " "
+                + child.getValue().getParty().getLastName() + " (Child " + i.getAndIncrement() + ")").build()));
```
chak-shing-lo-justice: May re-use `child.asLabel()` or `child.getParty().getFullName()` instead of building the the full name by yourself here.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/orders/prepopulator/question/WhichChildrenBlockPrePopulator.java
```diff
@@ -26,10 +25,11 @@ public OrderQuestionBlock accept() {
...
     @Override
     public Map<String, Object> prePopulate(CaseData caseData) {
-        final Selector childSelector = newSelector(caseData.getAllChildren().size());
+        final DynamicMultiSelectList childSelectorForManageOrders = childrenService
+            .getChildrenMultiSelectList(caseData);
         return Map.of(
-            "childSelector", childSelector,
-            "children_label", childrenService.getChildrenLabel(caseData.getAllChildren(), false)
+            "childSelectorForManageOrders", childSelectorForManageOrders,
+            "children_label", childrenService.getChildrenLabelFromMultiSelectList(childSelectorForManageOrders)
```
chak-shing-lo-justice: why we need to change it from `childrenService.getChildrenLabel` to `childrenService.getChildrenLabelFromMultiSelectList`?

The logic looks a bit different, will this change the behaviour when a final order is issued to the child?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/ChildrenService.java
```diff
@@ -198,4 +202,44 @@ public boolean hasAddressChange(List<Element<Child>> after, List<Element<Child>>
...
+        if (useAllChildren(appliesToAllChildren)) {
+            return children;
+        } else {
+            if (isNull(dynamicMultiSelectList) || isEmpty(dynamicMultiSelectList.getValue())) {
+                return Collections.emptyList();
+            }
+            return children.stream()
+                .filter(child -> dynamicMultiSelectList.getValue().stream()
+                    .anyMatch(listValue -> listValue.hasCode(child.getId())))
+                .toList();
```
chak-shing-lo-justice: Could we change this bits to more generic and move it to the `DynamicMultiSelectList`?

e.g. 
`public List<Element<T>> getSelectedElementFromMultiSelectList(List<Element<T>> elements)`
then this bits could be reuse by the `respondentsSelectorV2` as well

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
```diff
@@ -249,22 +250,25 @@ private Organisation getOrganisation(WithSolicitor respondent) {
...
                 return Collections.emptyList();
             }
-            return selector.getSelected().stream()
-                .map(respondents::get)
-                .collect(toList());
+
+            return respondents.stream()
+                .filter(respondent -> dynamicMultiSelectList.getValue().stream()
+                    .anyMatch(listValue -> listValue.hasCode(respondent.getId())))
+                .toList();
```
chak-shing-lo-justice: May reuse the new generic method created for DynamicMultiSelectList.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/RespondentService.java
```diff
@@ -328,4 +332,24 @@ public void transformRespondentLocalAuthority(CaseDetails caseDetails, CaseData
...
+            IncrementalInteger i = new IncrementalInteger(1);
+            caseData.getRespondents1().forEach(respondent -> {
+                listItems.add(DynamicMultiSelectListElement.builder().code(respondent.getId().toString())
+                    .label(respondent.getValue().getParty().getFirstName() + " "
+                    + respondent.getValue().getParty().getLastName()
+                    + " (Respondent " + i.getAndIncrement() + ")").build());
+            });
+        }
+        return DynamicMultiSelectList.builder().listItems(listItems).build();
+    }
```
chak-shing-lo-justice, line 342: Just a thought, maybe not possible. 
It looks quite similar to ChildSelctor, could we possibly extract them as a common method? so to reduce some code redundency?

e.g., a generic method accepting an element list of `? extends Party`

service/src/main/java/uk/gov/hmcts/reform/fpl/selectors/ChildrenSmartSelector.java
```diff
@@ -22,20 +22,20 @@ public class ChildrenSmartSelector {
...
+        boolean onlyOneChildCanBeSelected = childSelectionUtils.canOnlyOneChildBeSelected(caseData);
+
         if (isChildSelectedByPlacementApplication) {
             UUID placementId =
                 caseData.getManageOrdersEventData().getManageOrdersChildPlacementApplication().getValueCodeAsUUID();
             childrenToReturn = List.of(placementService.getChildByPlacementId(caseData, placementId));
+        } else if (onlyOneChildCanBeSelected) {
+            childrenToReturn = childSelectionUtils.getSelectedChildFromSingleSelectionComponent(caseData);
+        } else if (!childrenService.getSelectedChildrenFromMultiSelectList(caseData).isEmpty()) {
+            childrenToReturn = childrenService.getSelectedChildrenFromMultiSelectList(caseData);
```
chak-shing-lo-justice: Maybe we could extract the result of `childrenService.getSelectedChildrenFromMultiSelectList()` instead of calling it twice?

It is not a simple getter and probably just 15 items at max, but still better to avoid calling it twice if we could

chak-shing-lo-justice (changes requested):
Feel like the selectors could be reuse by other event in the future. Maybe we could refactor the code a bit, so to reduce some code redundancy in the future.
Ping me if you got any issue.

service/src/main/java/uk/gov/hmcts/reform/fpl/model/CaseData.java
JustGC-01: It's only being used for manageOrders for now, but will also be used for closeCase/recordFinalDecisions later on

chak-shing-lo-justice (dismissed):
lgtm

chak-shing-lo-justice (approved):

PR #6306, opened 2026-06-17
DFPL-2881: Add flow for application approve but applicant needs to make changes.

jamesrferguson1:
https://tools.hmcts.net/jira/browse/DFPL-2881

### Change description ###
Adds logic for option to approve application and reject uploaded draft order.

```
[ ] Yes
[ x] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/service/additionalapplications/ReviewAdditionalApplicationService.java
```diff
@@ -155,4 +166,61 @@ public List<Element<AdditionalApplicationsBundle>> markSelectedBundleAsReviewed(
...
+                && findElement(draftOrderElement.getId(), selectedDraftOrders).isPresent()) {
+                List<Element<T>> confidentialOrders =
+                    defaultIfNull(targetBundle.getConfidentialOrdersBySuffix(suffix), new ArrayList<>());
+                confidentialOrders.add(orderToBeAdded);
+                updates.put(targetBundle.getFieldBaseName() + suffix, confidentialOrders);
+                targetBundle.setConfidentialOrdersBySuffix(suffix, confidentialOrders);
+            }
+        });
+
+        return updates;
```
chak-shing-lo-justice: looks like this bit is copy from ApproveDraftOrderService. Is it possible to refactor / extract it, so we can reuse this bit?

A bit worry if there is any change to the confidential stuff in the future

jamesrferguson1: Yes, thought about this at the time ... thought there was already a lot of changes here so thought it might be safer just to duplicate for now. Giving it a bit more thought, a safe way of doing it might be to extract it to a shared util class - ConfidentialOrderBundleUtils perhaps?

chak-shing-lo-justice (approved):
lgtm

PR #6310, opened 2026-06-23
DFPL-3165: Migrate application insights to UK South

JustGC-01:
https://tools.hmcts.net/jira/browse/DFPL-3165

Created a new application insights module with the location variable set to "UK South". 

```
[] Yes
[x] No
```


infrastructure/alerts.tf
```diff
@@ -35,7 +35,7 @@ module "fpl-performance-alert" {
 
 module "fpl-exceptions-alert" {
   source                     = "git@github.com:hmcts/cnp-module-metric-alert"
-  location                   = var.appinsights_location
+  location                   = var.location
   app_insights_name          = "${var.product}-${var.component}-appinsights-${var.env}"
```
chak-shing-lo-justice: just to double check. 
Does this mean we change the location to UK, but still want the alert to monitor the existing appinsights instead of the new one? 

But from the changes (moved section), guess we are moving all new log generated to the new UK appinsights?


(Same for the other alert module)

JustGC-01: No, the alerts are being moved too - the appinsights_location variable is "West Europe" and location is "UK South", so both would be moved to the UK

chak-shing-lo-justice: Yes, I understand we are moving the location of all the module (alert and appinsght) to UK
But how about the app_insights_name? looks like the alert module is pointing to the old appinsghts?
"${var.product}-${var.component}-appinsights-${var.env}"

The new appinsights is ${var.product}-${var.component}-appinsights-uksouth-${var.env}

JustGC-01: Sorry, I misunderstood - I hadn't noticed that, i'll change it now

AlistairEdwinOsborne (dismissed):

AlistairEdwinOsborne (dismissed):

Braimah101 (approved):
Good to go

PR #6324, opened 2026-07-10
DFPL-3213: Migration Fleetwood cases to Blackpool

mfshabbir:
[DFPL-3213](https://tools.hmcts.net/jira/browse/DFPL-3213)

### Change description ###

Contains the migration script for Fleetwood cases to Blackpool

```
[ ] Yes
[ ] No
```


service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededController.java
```diff
@@ -180,6 +199,24 @@ public AboutToStartOrSubmitCallbackResponse handleAboutToSubmitEvent(
         return respond(caseDetails);
     }
 
+
+    private boolean isFleetwoodCase(CaseData caseData) {
+        if (caseData == null) {
```
jamesrferguson1: Not sure if we really need this null check - if the caseData is null we have much bigger problems!

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededController.java
```diff
@@ -53,6 +53,15 @@ public class OrdersNeededController extends CallbackController {
     @SuppressWarnings("unchecked")
     public AboutToStartOrSubmitCallbackResponse handleMidEvent(@RequestBody CallbackRequest callbackrequest) {
         final CaseData caseData = getCaseData(callbackrequest);
+
+        CaseDetails caseDetailsIfFleetWood = callbackrequest.getCaseDetails();
```
jamesrferguson1, line 62: Might be worth adding a TODO to remove this check ... is this a long term solution (or just a workaround to avoid a migration)?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
```diff
@@ -1537,6 +1540,66 @@ public Map<String, Object> removeFirstOther(String migrationId, CaseData caseDat
...
+        final Orders orders = caseData.getOrders();
+
+        //  validation checks
+        if (caseManagementLocation == null) {
+            throw new AssertionError(String.format(
+                "Migration {id = %s, case reference = %s}, caseManagementLocation structure is missing",
+                migrationId, caseData.getId()));
+        }
+
+        if (!"401452".equalsIgnoreCase(caseManagementLocation.getBaseLocation())) {
```
jamesrferguson1: Would it perhaps be better to move check list this to the migration in the controller and keep this logic more generic rather than referencing Fleetwood directly?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
```diff
@@ -1537,6 +1540,66 @@ public Map<String, Object> removeFirstOther(String migrationId, CaseData caseDat
...
+            throw new AssertionError(String.format(
+                "Migration {id = %s, case reference = %s}, expected Fleetwood (401452) but found baseLocation: %s",
+                migrationId, caseData.getId(), caseManagementLocation.getBaseLocation()));
+        }
+
+        // Update core location block to Blackpool's epimmsId
+        caseManagementLocation.setBaseLocation("214320");
```
jamesrferguson1: Similar to above comment, maybe make the new base location a method parameter and pass it in in our controller?

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededController.java
mfshabbir: Thanks James! I have added a comment above the check explaining that this is temporary validation for the Fleetwood court migration (linked to ticket DFPL-3213) and noted that it should be removed once migration is complete. (Note: Omitted the literal TODO keyword as Sonar flags TODO tags as code smells in our pipeline), hope that’s okay?

service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
mfshabbir: Done, All tests and Checkstyle checks are now passing and green. Ready for another look when you're free!. Thanks

service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
mfshabbir: Thanks James, makes total sense.
I have refactored updateCaseManagementLocation in MigrateCaseService to be generic and take the location codes, court details, and DFJ key as parameters. I also updated the controller and test suites accordingly.

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/OrdersNeededController.java
mfshabbir: Thanks for the review and Good point, Removed the null check as caseData will always be present at this point.

service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
```diff
@@ -1537,6 +1539,76 @@ public Map<String, Object> removeFirstOther(String migrationId, CaseData caseDat
...
+        updates.put(CASE_MANAGEMENT_LOCATION, caseManagementLocation);
+
+        if (updatedCourt != null) {
+            updates.put(COURT, updatedCourt);
+        }
+        if (updatedOrders != null) {
+            updates.put(ORDERS, updatedOrders);
+        }
+
+        // Dynamic DFJ court property update
```
jamesrferguson1, line 1601: Do we need to set this field here - think it's a temp field that should be getting cleared at the end of the event?

(see removeTemporaryFields(caseDetails, CaseProgressionReportEventData.class);

in the handleAboutToSubmit in CaseProgressionReportController.)

jamesrferguson1 (dismissed):

chak-shing-lo-justice (approved):

PR #6347, opened 2026-08-13
Dfpl 2883: C2 List at next hearing

jamesrferguson1:

https://tools.hmcts.net/jira/browse/DFPL-2883

### Change description ###

Adding List at next hearing option when approving C2 Applications.

```
[ ] Yes
[ x] No
```


chak-shing-lo-justice (changes requested):
do we still need to add the doc template to this repo?

ccd-definition/CaseEventToFields/CareSupervision/reviewAdditionalApplication/reviewAdditionalApplication.json
```diff
@@ -234,6 +234,21 @@
     "PageColumnNumber": 1,
     "FieldShowCondition": "approveAdditionalAppRouter=\"REFUSE\""
   },
+  {
+    "LiveFrom": "01/01/2017",
+    "CaseTypeID": "CARE_SUPERVISION_EPO",
+    "CaseEventID": "reviewAdditionalApplication",
+    "CaseFieldID": "previewApprovedOrder2",
```
chak-shing-lo-justice, line 241: just wondering if we can reuse the `previewApprovedOrder1`.
But  it is a just temp field any way, so should be fine

jamesrferguson1: I tried this and it was a bit of a headache. I think the issue was that previewApprovedOrder1 is tied to the page show condition of the review refusal order route so settled on it being cleaner to just use previewApprovedOrder2

PR #6356, opened 2026-08-26
DFPL-3213: fix-transferred-case

mfshabbir:

[DFPL-3213](https://tools.hmcts.net/jira/browse/DFPL-3213)

### Change description ###
Follow-up fix for DFPL-3213 to enable migration for remaining case

```
[ ] Yes
[ ] No
```


AlistairEdwinOsborne (approved):

AlistairEdwinOsborne (dismissed):

service/src/main/java/uk/gov/hmcts/reform/fpl/service/MigrateCaseService.java
chak-shing-lo-justice: looks like this method is not updating case management location anymore
probably worth either creating a new method and name it as something like `updateSelectedCourtInOrderNeeded`.

Then refactor this existing method to reuse the new method just created

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
AlistairEdwinOsborne, line 1: try not to use statics outside of your migration method, for the migrations Map always put the string in so its more readable especially when someone is looking across this and the tool repo. Looks all good but just put the location/court code and ID inside your migration methods

AlistairEdwinOsborne (changes requested):

service/src/main/java/uk/gov/hmcts/reform/fpl/controllers/support/MigrateCaseController.java
mfshabbir: Thanks @AlistairEdwinOsborne , Done. Removed the top-level statics and moved everything inside the migration method as suggested

AlistairEdwinOsborne (approved):

